PR #6, opened 2023-12-06
Refine schema to more closely reflect metadata collection applications

caufieldjh:


caufieldjh:
Standard practice on fixing the linkml import-related issue causing tests to fail is to copy n' paste the necessary imported bits into this schema.

caufieldjh:
Schema updates are not complete by any means but the schema is fully usable.

PR #154, opened 2026-06-01
Rubric review

Bankso:


realmarcin:
## Review decomposed into tracking issues

Review of this branch against the new D4D-core + semantic exchange layer. The edits here are a net improvement and merge-worthy after the pre-merge items below; the larger core-driven gaps are filed as follow-ups.

**Pre-merge (blocking):**
- #155 — "report but don't score" state has no output encoding or denominator rule (most important)
- #156 — introduced typos + trailing whitespace

**Core/exchange follow-ups:**
- #157 — rubric10-semantic lacks the conditional applicability added to rubric20 (penalizes by-design core omissions)
- #158 — both -semantic agents lag the v2.1 source rubric fields (RAI, Data Protection & Compliance, composition)
- #159 — no collection-level / distribution-level traversal guidance for D4D-core
- #160 — (decision) no check for semantic exchange-layer conformance / mapped identifiers

realmarcin:
## Adversarial review (Codex) — results

Ran `/codex:adversarial-review` over the branch diff. Verdict: **needs-attention / no-ship as-is.** All findings were independently re-verified against the files before acting.

**New substantiated issues opened:**
- #161 `bug` — **circular applicability gates**: several rubric20 questions are gated on evidence that only exists in the question's own fields (Q8/Q9/Q11), so real gaps can be silently unscored — worst on sparse core files.
- #162 `bug` — **"program of origin" is ungrounded**: no `program`/`project` field exists in the D4D datasheet, so the new check forces context-dependent inference and breaks the temp=0.0 determinism guarantee.

**Existing issues audited (verdicts + new evidence commented inline):**
- #155 **KEEP/blocking** — strengthened: max_points is hard-fixed at 84 (`d4d-rubric20-semantic.md:455,720`), committed baselines + summary schema assume it; conditional non-scoring breaks comparability. Fix must add raw+normalized percentages (or keep full denominator) **and** migrate baselines. (Absorbs Codex Finding 2.)
- #156 **DOWNGRADE** — real but not blocking.
- #157 **KEEP/blocking** — core intentionally excludes fields (`D4D_Core.yaml:124-126`); rubric10 still penalizes by-design omissions.
- #158 **KEEP/blocking** — scope expanded to absorb Codex Finding 4: the PR's new section-name anchors ("Licensing & Governance" etc.) don't match real field names, and Q8 omits `data_protection_impacts`/`reidentification_risk`. Fix = anchor to exact fields.
- #159 **KEEP/blocking** — collection/distribution hierarchy (`D4D_Core.yaml:88-107,140-149`) isn't traversed.
- #160 **KEEP** — real but **not** a blocker; decide rubric-check vs. dedicated validator.

**Minimal blocking subset before merge:** #155, #158, #159, #161, #162 (plus the #156 cleanup folded in). #157 should be designed alongside #155/#161. #160 is non-blocking.

Bankso:
Changes added since LLM review address issues #155, #156, #157, #158, #159, #161, #162 
I am still reviewing issue #160 to determine if this is something we should address/implement.

realmarcin:
> Changes added since LLM review address issues #155, #156, #157, #158, #159, #161, #162 I am still reviewing issue #160 to determine if this is something we should address/implement.

@Bankso Sounds good, let me know if any updates -- Sat. 6/27 I am taking off for a few weeks to Europe so we may have to submit the paper later in July.

#160 is worth thinking about though the ro-crates have their own independent eval. I would vote for something simple here, maybe the rubrics can have an identified subset that's is more directly about the exchange layer? We would have to report two sets of scores. However, D4D-core should be more or less representative of exchange layer ...

PR #164, opened 2026-07-29
Prompt-condition study: generic vs tuned, with corpus provenance and an API generation path

realmarcin:
## What this is

The arms study measured what upstream structured metadata (RO-Crate packages, AI-READI's Healthsheet) adds to a D4D record. Partway through it emerged that the results were tracking **prompt content** rather than crate content, so this branch reworks the experimental design, adds the infrastructure the study needed, and records the corpus properly.

## The finding that forced the rework

The 2026-07-27 series gave a clean ordering — CHORUS ≫ VOICE ≫ CM4AI — and that ordering matched what each project's prompt had told the agent to expect. Re-running without expectation statements moved the verdicts materially:

| project | arm | primed | de-primed |
|---|---|---|---|
| CHORUS | crate | `+22,+16,+20` ±13 → **real** | `+28,+12,+14` ±18 → **marginal** |
| VOICE | crate | `+6,+10,+8` ±7 → **marginal** | `0,+1,0` ±6 → **not resolvable** |

**VOICE's crate contribution was substantially a priming artifact.** Told the crate was "expected to be genuinely additive", the arm produced 81/85/83 slots against a 75/75/75 baseline; told nothing, 76/77/77 against 76/76/77. Its stable crate-only set falls from 4 slots to 1.

Of four Grand Challenges, only CHORUS retains a reproducible crate contribution, and only at *marginal*. Nine of its twelve stable crate-only slots survive de-priming — that 9-slot core is the defensible claim, not the delta.

## The design rework

Three conditions turned out not to be one axis: each project's prompt carried a *different kind* of content, so "tuned vs generic" was four experiments sharing a label. Five arms gave five different winners:

| arm | tuned prompt carried | best condition |
|---|---|---|
| CHORUS baseline | a decision rule, GC-scoped by accident | **generic** 84.7% |
| AI-READI baseline | a GC-specific fact | **tuned** 95.2% |
| AI-READI healthsheet | a density expectation | **deprimed** 87.0% |
| VOICE baseline | a GC-specific fact | generic collapses to **76.7%** |

GC-specific *facts* are load-bearing; GC-specific *expectations* cost accuracy; rules that should be uniform should be uniform. "Less prompt content is better" is not supported.

So conditions are now **generic** (one identical prompt, `src/download/prompts/d4d_generic_arm_prompt.md`) and **tuned** (= generic + a typed component block). The tuned prompt references the generic body rather than copying it, so the base cannot drift. Components are typed `fact` / `decision-rule` / `referent-pin`; `expectation` is excluded as the one category with measured harm. Six tests enforce this rather than trusting the prose.

## Infrastructure

- **`runs.py`** — replicate/revision distinction and a completeness gate. Reading a run that has only written phase 1 silently measures an unfinished run; that produced a spurious +12 and a spurious −14 before the gate existed.
- **`provenance.py`** — live vs reconstructed records, with `unrecoverable` for fields that cannot be determined and `unverified` for fields that are populated but asserted.
- **`rocrate_normalize.py`** — crate normalisation, plus a document-corpus guard so with-crate and without-crate arms stay separable.
- **`fetch.py` + `d4d download supplements`** — 8 of 35 manifest sources are not obtainable from the input sheet, so a fresh clone silently got a smaller corpus than every run consumed.
- **`schema_digest.py`** — the merged schemas are ~254 KB / ~216 KB, larger than any input bundle. The digest is 18 KB / 9 KB, a 98–99% reduction, carrying structure only.
- **`api_runner.py` + `d4d api`** — four-phase generation over the Anthropic API, sharing the GitHub assistant's deterministic config.

## Two provenance problems this surfaced

1. **`Temperature: 0.0` was never measured.** Claude Code exposes no temperature setting — the agents wrote that line because the prompt template said to. It is in 33+ records as though observed. Now carried under `unverified`.
2. **Model mismatch.** The assistant config pinned `claude-opus-4-8` while every run used `claude-opus-5[1m]` — the assistant and batch pipeline had been running different models. Both now pinned to `claude-opus-5`.

## Corpus

`source_manifest.yaml` (35 sources) was untracked, so the only record of what the corpus contained lived in one working directory. Four superseded entries are reclassified with `superseded_by` / `retained_because` rather than dropped — each preserves content that exists nowhere else, and keeping them leaves the bundles byte-identical, which is what preserves comparability across every run in the study.

Six upstream sheet discrepancies were actioned (five cells corrected, one row inserted); `notes/UPSTREAM_SOURCE_DISCREPANCIES.md` records them.

## Verification

- 487 tests pass. One pre-existing failure (`test_render.py`) is unrelated and predates this branch — confirmed by reverting the only schema change on it.
- All 33 Claude Code runs: LinkML-clean, live provenance, no unrecoverable fields, per-arm input hashes recorded.
- `data/poster_assets/` is gitignored and excluded.

## Known limitations

- Agreement measures self-consistency, not correctness. Without CM4AI's referent pin, `file_collections` falls 10 → 0 in all three replicates and agreement does not penalise it — the runs omit it *consistently*. Scoring against the curated records is the obvious next step.
- n=3 with three-way noise at ±13–18 cannot resolve effects of that size; CHORUS's crate verdict flipped on the noise floor moving, not the effect.
- The healthsheet arm is AI-READI-only and crates cover 3 of 4 GCs, so pooling across arms is not valid.
- The API path reimplements the four phases without tools, so a generic-vs-generic comparison across harnesses mixes "no tools" with "different orchestration".


realmarcin:
## Review

Reviewed the diff (861 files) for defects rather than style. Checks run:

- **No secrets or credentials.** The service-account path used for the Google Sheet corrections appears nowhere in the diff; the API runner reads `ANTHROPIC_API_KEY` from the environment and never persists it.
- **`poster_assets/` excluded** and now gitignored, so a future `git add -A` cannot sweep it in.
- **Ordering invariants hold** in `api_runner.execute()` — outputs are written before the provenance record, and the record file last, so a failed generation cannot leave an orphan record claiming success. `d4d api run` refuses to reuse a run label.
- **`fetch.py` guard ordering** — `is_manual` is checked before `force`, so `--force` cannot make the runner pretend a manual capture is re-fetchable.
- **Unused imports** found and removed (`Iterable`, `json`).

### Fixed in this PR

- #167 — the render test fixture. Pre-existing failure; first fix attempt was wrong (put `id` at the top level, but `validate_d4d_yaml` unwraps `DatasetCollection` and validates the extracted resources). Now correct. **Suite is 488 passing, 0 failing** — green for the first time across this work.

### Filed, not fixed here

- #165 — agreement measures self-consistency, not correctness. The strongest evidence is in this PR: CM4AI's `file_collections` falls 10 → 0 in all three generic replicates and agreement does not penalise it. This bounds every prompt-quality claim and is the highest-value follow-up.
- #166 — generic condition incomplete; de_novo and crate_only arms not run (18 runs).
- #168 — provenance mixes md5 (inputs) and sha256 (prompts).
- #169 — n=3 cannot resolve effects near the ±13–18 noise floor; design unbalanced across arms; thresholds were applied post-hoc.
- #170 — the API path has never been exercised against the live API.

### Assessment

Merging. The infrastructure and corpus work is verified and self-contained. The experimental conclusions are explicitly bounded in the PR body and in `notes/D4D_GENERATION_ARMS.md`, which now marks the earlier headline figures as a primed condition rather than the result — the point of the branch is that the previous conclusions were not safe to keep, not that these ones are final.

Issue #171, opened 2026-07-29
Test: generate CHORUS datasheet from concatenated input docs


realmarcin:
@d4dassistant please create a D4D datasheet for **CHORUS** using File-Based Mode.

**Input:** `data/sheets_d4dassistant/inputs/CHORUS/CHORUS_preprocessed.txt`

This is the same 36 KB concatenated corpus the batch pipeline uses for CHORUS (4 sources, selected by `data/preprocessed/source_manifest.yaml`), so the output should be directly comparable to the runs already in `data/d4d_concatenated/claudecode_agent/`.

### Why this run

Exercising the assistant end to end after several changes landed in #164:

- the deterministic config is now pinned to `claude-opus-5` (it previously pinned `claude-opus-4-8` while every batch run used `claude-opus-5[1m]`, so the two paths had silently diverged)
- the CHORUS reference example path was stale — it moved into `2026-04-10_sonnet-4.6/` when outputs were reorganised into dated run directories, and has just been corrected

### What to check in the result

- LinkML validation against `Dataset` passes
- `{dataset}_d4d_metadata.yaml` is produced with SHA-256 hashes of input, schema and prompts, plus the git commit
- the metadata records `claude-opus-5`, matching the config
- slot count against the validation gates in `d4d_assistant_deterministic.config` — note CHORUS is the thinnest corpus in the study by roughly an order of magnitude (36 KB vs 322–421 KB), so a low count is expected and is a property of the sheet, not a failure

Absent evidence here means absent from the input sheet, not absent upstream — CHORUS has no publication, licence, IRB or DUA row. Please report what the bundle supports and leave the rest unpopulated.

realmarcin:
The assistant could not run. `respond-to-mention` failed at job setup, before any repository code executed:

```
Getting action download info
##[error]Repository access blocked
```

**Cause:** the workflow depends on `dragon-ai-agent/run-claude-obo@v1.0.2`, and both that repository and the `dragon-ai-agent` organisation now return 404. Filed as #172.

This is not a problem with the input or the instructions — the job dies before reading either. The CHORUS input staged at `data/sheets_d4dassistant/inputs/CHORUS/CHORUS_preprocessed.txt` is in place and ready for a re-run once the workflow is repaired.

[...]

Issue #170, opened 2026-07-29
API generation path has never been exercised against the live API


realmarcin:
PR #164 adds `src/data_sheets_schema/api_runner.py` and `d4d api plan` / `d4d api run`. Everything is tested through `plan()`, which renders each phase's request without a key or a charge — 20 tests cover prompt resolution, cache layout, phase assembly and cost estimation.

**No call has been made against the real API.** Untested in particular:

- response parsing — `_extract()` strips markdown fences, but real responses may wrap output differently per phase
- the phase-4 contract, which expects a JSON object with `full_yaml`, `core_yaml`, `report_markdown`; a model that returns YAML-in-YAML or omits a key will raise
- cache hit behaviour — the bundle+digest prefix should be written once and read on phases 2–4, and `api_usage` records `cache_read` / `cache_write` to verify, but the numbers have never been observed
- whether `max_tokens: 16000` is sufficient for a full record on the largest bundles (VOICE records run 139 KB)

### Suggested first run

CHORUS baseline, generic, one replicate — the cheapest cell at ~109k input tokens — and inspect `api_usage` before committing to a batch.

realmarcin:
Partially exercised. Three live runs from #171 took the API path from never-executed to authenticating against the real endpoint.

**Confirmed working:** request construction for all four phases, the shared deterministic config resolving `claude-opus-5` at temperature 0.0, and the cost estimate — the workflow printed `~108,820 input tokens` against the local `d4d api plan` figure of `~108,815`, so the estimator is accurate in the runner environment.

**Found and fixed:** `execute()` imported `record_path` from `provenance`, where it does not exist (#174). Exactly the failure mode this issue predicted — `execute()` imports inside the function body, so every `plan()`-based test missed it and it could only surface on a live call. Offline `execute()` coverage with a fake client now exists and would have caught it.

**Still unverified**, because the run stops at authentication:

- response parsing on real output — `_extract()` fence-stripping is only exercised against synthetic responses
- the phase-4 contract (`full_yaml` / `core_yaml` / `report_markdown` in one JSON object)
- cache hit behaviour — `api_usage` records `cache_read` / `cache_write`, but no real numbers observed
- whether `max_tokens: 16000` suffices for a full record

All of those need a valid `ANTHROPIC_API_KEY`; the repo secret dates from 2025-11-11 and returns 401.

realmarcin:
Done. The path was exercised against the live API on 2026-07-29 and has run
[...]

PR #175, opened 2026-07-29
Make the API generation path fit real records

realmarcin:
Readiness audit of the API path against the 112 records already generated. It would have failed on almost every project.

## 1. Phase 4 was structurally impossible

It returned `full_yaml` + `core_yaml` + `report_markdown` in one JSON object — all three artifacts in a single response:

| project | phase-4 output | over the old 16k ceiling |
|---|---|---|
| CHORUS | ~30,974 tok | 1.9× |
| CM4AI | ~55,050 tok | 3.4× |
| VOICE | ~67,088 tok | 4.2× |
| AI-READI | ~67,736 tok | 4.2× |

Not a limit to raise — a shape to abandon. Split into `reconcile_full`, `reconcile_core`, `report`: one artifact per call, six phases total.

## 2. `max_tokens: 16000` truncated five of six projects

Real records run 14k–43k output tokens. Now derived per phase from the largest artifact of each kind already generated (112 full, 104 core, 97 report) plus ~40% headroom — 64000 / 56000 / 12000 — rather than guessed.

A `stop_reason` of `max_tokens` now **raises instead of writing**. A truncated datasheet parses as valid YAML while silently missing most of its content; that is worse than no output, because it validates.

## 3. A failure discarded every paid call

Nothing was written until phase 4 parsed, so a phase-5 failure forfeited five calls. Each phase now writes as it completes, with progress in `{project}_api_progress.json`.

**Progress is recorded, not inferred.** A `full` record on disk may be pre- or post-reconciliation and nothing in the file distinguishes them — inferring would either skip reconciliation silently or redo it. The file is deleted on success; a corrupt one means redo rather than crash.

## 4. `d4d api batch`, retry, cost reporting

Sweeps projects × replicates with cumulative token reporting; each run resumes independently, so an interrupted sweep costs only unfinished phases.

Retries 429 / connection / 5xx with exponential backoff. **Does not retry auth or other 4xx** — an invalid key fails identically every attempt, and retrying only delays the operator seeing the real problem. That is not hypothetical: the live run in #171 failed on exactly that.

## Cost, six phases, uncached

```
full baseline sweep (4 projects x 3 replicates): ~11,052,426 input tokens
```

Higher than the four-phase estimate because reconciliation is now three calls carrying the record forward. Prefix caching applies to the bundle and digest on every phase.

## A test bug this exposed

The fake client identified phases by substring-matching `"Phase 3"` — but the prompt body enumerates all four playbook phases, so **every** request matched and the fake reported every call as the audit phase. Tests passed while exercising one phase six times. It now matches the actual instruction text.

## Verification

499 tests pass. New coverage: resume after mid-run failure, `resume=False` forcing a redo, corrupt progress file, truncation raising rather than writing, transient retry, auth not retried.

**Still unverified:** no live API call has succeeded — the repo key returns 401 (#170, #171). Response parsing and cache-hit behaviour remain untested against real output.

### DIFF
--- a/src/data_sheets_schema/api_runner.py
+++ b/src/data_sheets_schema/api_runner.py
@@ -30,6 +30,7 @@ from __future__ import annotations
 import json
 import os
 import re
+import time
 from dataclasses import dataclass, field
 from datetime import datetime, timezone
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


realmarcin:
Superseded — closing unmerged.

Everything this PR proposed is already on `main`, carried in by the later API
work rather than by this branch:

- the six-phase split (`full`, `core`, `audit`, `reconcile_full`,
  `reconcile_core`, `report`) — present
- `PHASE_MAX_TOKENS` sized per phase from real records instead of a flat 16k —
  present, at 64k/56k/24k/12k

Checked mechanically: every function, class and module-level constant this
branch defines in `api_runner.py` and `cli/api.py` already exists on `main`, and
several of its comments appear there verbatim. `main` is 33 commits ahead; the
branch has one, and its content is duplicated.

Merging would contribute nothing and risk conflicting with work that has since
built on the same functions — including the resume, extraction, retry and
rate-limit fixes in #213 and #220.

The audit that motivated it stands; it simply landed by another route.

PR #200, opened 2026-07-31
Adopt the RO-Crate agent doc from rocrate-skill; document all four transformation paths

realmarcin:
Salvages the one file worth keeping from `rocrate-skill`, and documents what it
is among.

## What was salvaged, and why the rest was not

`.claude/agents/d4d-rocrate.md` is the only file on `rocrate-skill` that main
lacks and still wants. The nine transformation scripts it drives are already on
main, and every path it references resolves.

The rest of that branch is stale. It is one commit from **2026-03-09**, 61 behind
main, and the only two files that differ are ones where main moved *forward*:

| file | branch | main |
|---|---|---|
| `auto_process_rocrates.py` | RO-Crate `1.1` context | **`1.2`** |
| `field_prioritizer.py` | `vulnerable_populations` | **`at_risk_populations`** |

The second is not cosmetic — `vulnerable_populations` no longer exists in the
schema. Merging would reintroduce a field name that resolves to nothing and
silently break policy-field prioritisation.

Its remaining files are a README duplicating the agent doc, two notes describing a
March implementation the `d4d rocrate` CLI has since replaced, and stale generated
*outputs* (`*_validation_errors.txt`, `transformation_report.txt`) rather than
fixtures. No test on main references `.claude/agents/scripts` at all.

`rocrate-skill` should be closed unmerged.

## What the doc was missing

It is accurate about what it covers and silent about what it is *among*. The
repository holds **four** RO-Crate transformations, not one bidirectional pair:

    1  RO-Crate -> D4D   this agent (TSV mapping, multi-crate merge)
    2  RO-Crate -> D4D   d4d rocrate map (repo's static mapping, quality-reported)
    3  RO-Crate -> D4D   fairscape_to_d4d.py (FAIRSCAPE, SSSOM-guided)
    4  D4D -> RO-Crate   d4d_to_fairscape.py (emits ROCrateV1_2)

The doc mentioned none of the other three. An agent reading it would not have
known an outbound direction exists at all, nor that paths 1 and 2 both go inbound
and **will not agree** — one applies an external mapping table, the other this
repository's own, so a record from one is not interchangeable with a record from
the other and comparing them measures the mappings rather than the crates.

Path 2 is what the generation study uses via `emit-map-arm`, so the doc now
directs anything destined for comparison against a generated arm there, and
reserves this agent for crates outside that pipeline.

## The table as data

`src/docs/rocrate_transformation_paths.tsv`, five columns, carrying what the
prose table implied but did not state — which path the study uses, and that path
4 is a different subsystem rather than the inverse of 1-3.

It is **not** in `docs/` directly, despite that being where it was asked for.
`.gitignore` ignores `docs/*` and `make clean` runs `rm -fr docs/*`, so a
hand-authored file there is dropped by git and destroyed by the next clean. A
first attempt committed only the *reference*, pointing at a file the repository
did not contain. The tracked source is `src/docs/`, which `gendoc` copies into
`docs/` — but only `*md`, so the copy rule now carries `.tsv` too. It reaches
`docs/` as asked and survives.

## Verification

- `720 passed, 3 skipped`
- All four entry points in the TSV resolve — `rocrate_to_d4d.py --help`,
  `d4d rocrate map --help`, and both fairscape modules import
- The TSV parses with a uniform five-column shape
- `cp src/docs/*.tsv docs/` places the file, so `gendoc` will

Closes the question raised on the `rocrate-skill` branch.

### DIFF
--- a/.claude/agents/d4d-rocrate.md
+++ b/.claude/agents/d4d-rocrate.md
@@ -1,0 +1,561 @@
+---
+name: d4d-rocrate
+description: |
+  Transform RO-Crate JSON-LD metadata (from fairscape-cli) into D4D YAML datasheets.
+  Examples:
+    - "Convert CM4AI RO-Crate to D4D"
+    - "Generate D4D from fairscape-cli output"
+    - "Transform ro-crate-metadata.json to D4D YAML"
+    - "Map RO-Crate to datasheet format"
+model: inherit
+color: purple
+---
+
+## Which path is this, and when to use it
+
+The repository has **four** RO-Crate transformations, not one bidirectional
+pair. They were built at different times for different purposes and none
+supersedes the others. Pick deliberately.
+
+| # | direction | entry point | use when |
+|---|---|---|---|
+| 1 | **RO-Crate → D4D** | this agent — `.claude/agents/scripts/rocrate_to_d4d.py` | transforming arbitrary crates with a TSV mapping, merging several crates, or ranking sources by informativeness |
+| 2 | **RO-Crate → D4D** | `d4d rocrate map` | the study's own static mapping, which reports the declared mapping quality of every field it fills |
+| 3 | **RO-Crate → D4D** | `src/fairscape_integration/fairscape_to_d4d.py` | a FAIRSCAPE crate specifically, mapped under SSSOM guidance |
+| 4 | **D4D → RO-Crate** | `src/fairscape_integration/d4d_to_fairscape.py` | emitting a FAIRSCAPE RO-Crate (`ROCrateV1_2`) *from* a datasheet |
+
+**This agent is path 1: inbound only.** It reads a crate and writes a datasheet.
+It cannot produce a crate — for that, use path 4, which is a different subsystem
+with its own Pydantic models rather than the TSV mapping used here.
+
+Paths 1 and 2 both go RO-Crate → D4D and will not agree. Path 1 applies an
+external mapping table and can merge multiple crates; path 2 applies this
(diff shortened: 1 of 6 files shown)


realmarcin:
## `rocrate-skill` closed unmerged — superseded by this PR

The branch is one commit (`4bb47854`, 2026-03-09), 61 behind main. Its nine
transformation scripts and mapping table are **already on main**; only the agent
doc was missing, and that is what this PR adopted.

**Merging it would have regressed main.** The only two files that differ are ones
where main moved forward:

| file | branch | main |
|---|---|---|
| `auto_process_rocrates.py` | RO-Crate `1.1` context | **`1.2`** |
| `field_prioritizer.py` | `vulnerable_populations` | **`at_risk_populations`** |

The second is not cosmetic — `vulnerable_populations` no longer exists in the
schema. Merging would have reintroduced a field name that resolves to nothing and
silently broken policy-field prioritisation.

That same rename turned out to be live in the mapping table too, which this PR
fixed (#202) along with the stale coverage figure (#201).

**The remaining files were not worth carrying:** a README duplicating the agent
doc, two notes describing a March implementation the `d4d rocrate` CLI has since
replaced, and stale generated *outputs* (`*_validation_errors.txt`,
`transformation_report.txt`) rather than fixtures. No test on main references
`.claude/agents/scripts` at all.

The branch tip is recorded here as `4bb47854412e2622a70f48c59c83d845072bd0e0` —
recoverable with `git fetch origin 4bb47854` if anything is wanted from it later.

Issue #226, opened 2026-08-02
FileCollection.resources declares a File|FileCollection union that has never been satisfiable


realmarcin:
Found while fixing `make test-modules`.

`FileCollection.resources` declares:

```yaml
resources:
  any_of:
    - range: File
    - range: FileCollection
  range: File
```

The schema already carries a NOTE saying "Generated artifacts do not fully
reflect this union type constraint". The situation is worse than that: **the
union has never been satisfiable at all.**

`range` and `any_of` are ANDed by the JSON-Schema generator, so the effective
constraint is `File AND (File OR FileCollection)` — which admits `File` and
rejects nested `FileCollection`, the precise case the union exists to allow.
Measured on a minimal schema:

| formulation | errors on a `File` value |
|---|---|
| `range: File` + `any_of(File, FileCollection)` | 1 |
| `any_of(File, FileCollection)`, no range | 2 (falls back to `string`) |
| `range: File` alone | 0 |

Before the module-validation fix the outer range was `Dataset` (inherited from
the base slot), giving `Dataset AND (File OR FileCollection)` — unsatisfiable
for *any* value. That is consistent with the corpus: **zero records populate
this slot**, across every run of every arm.

## Options

1. **Give `File` and `FileCollection` a common parent** and range the slot on
   it. Neither declares `is_a` today. This is the modelling fix and it makes the
   union unnecessary.
2. **Drop the union and range on `File`**, accepting that nested collections are
   expressed through `file_collections` on the parent rather than recursively.
   Cheapest, and matches how the corpus is actually written.
3. Keep the union and wait for generator support.

Option 1 or 2; the current state documents an intent the toolchain cannot
express, and a slot nothing can populate is worse than an honest narrower one.

## Note

Not fixed alongside the module-validation work on purpose: that branch is about
making validation run, and changing this slot's semantics is a modelling
decision that deserves its own review.

realmarcin:
Fixed in #232 (merged). Verified on `main`: `FileCollection.resources` is
`range: File` with no `any_of`.

Option 2 from this issue — drop the union, range on `File`. Option 1 (a common
parent) was rejected on measurement: `File` and `FileCollection` already share
`is_a: Information`, but that also parents `Dataset` and `DatasetCollection`, so
ranging on it would have been looser than the union it replaced. A new
intermediate class would have been a fresh idiom for this schema, which has no
abstract or mixin class among its 78 — introduced for a capability nothing uses.
[...]

PR #227, opened 2026-08-02
Make module-level schema validation actually run

realmarcin:
`make test-modules` has failed on `main` for as long as the target has existed,
so module-level validation has never run for anyone. It fails on the first
module and exits, which is why only the first cause was ever visible.

Four separate causes, each uncovered by fixing the one in front of it.

## 1. The base module pointed upward

`D4D_Base_import.yaml` gave slot `resources` `range: Dataset`. `Dataset` is
defined in `data_sheets_schema.yaml`, which **imports** the base — so the
reference pointed outside the base's own import closure and could not resolve.

The slot's own description already claimed "the specific range is defined via
`slot_usage` in each class". It now genuinely is: `DatasetCollection` gains the
explicit range it had been silently inheriting, and `Dataset`/`DataSubset`
already had one.

## 2. The target ran generators the project does not build

Bare `gen-project` runs *every* generator, including `shaclgen`, which fails on
these modules with `KeyError: data-sheets-schema-base` — a generator limitation
around imported schemas, not a fault in the modules.

`gen-project` in this Makefile builds python, jsonschema, jsonld and owl.
`test-modules` now validates that same set, so a failure here is something
someone can act on rather than a wall.

## 3 & 4. D4D_FileCollection could not stand alone

- imported `linkml:types` without declaring the `linkml` prefix
- used `data_sheets_schema` as `default_prefix` without declaring it
- referenced `ExternalResource` without importing the module that defines it

The first two resolved only when some *other* schema in the closure happened to
declare them — precisely what building a module alone is meant to catch.
`D4D_Composition` imports only the base, so importing it introduces no cycle.

## Result

All 17 modules build.

The merged schema is unchanged except `FileCollection.resources`, which moves
from `Dataset` to `File`. `Dataset` was never intended there — the description
says "individual files or nested file collections" — and was only inherited from
the base slot's unresolvable range. **Validation behaviour is identical**: one
error before, one after, on the same input, verified by validating the same
record against both branches.

## Filed, not fixed here

#226 — `FileCollection.resources` declares a `File|FileCollection` union that
has never been satisfiable, because `range` and `any_of` are ANDed by the
generator. Zero records populate that slot across the entire corpus, which is
consistent. Changing its semantics is a modelling decision that deserves its own
review rather than riding along here.

## Testing

889 passing, 3 skipped. `make test-modules` passes — for the first time.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -221,11 +221,18 @@ test-schema: $(SOURCE_SCHEMA_ALL)
 	$(RUN) gen-project ${GEN_PARGS} -d tmp $(SOURCE_SCHEMA_ALL)
 
 # Test individual D4D module schemas
+# GEN_PARGS-free on purpose, but the generator set must match `gen-project`
+# above. Bare `gen-project` runs *every* generator, including shaclgen, which
+# fails on these modules with `KeyError: data-sheets-schema-base` — a generator
+# limitation around imported schemas, not a fault in the modules. The project
+# never builds SHACL, so validating it here only produced a failure nobody could
+# act on, and the target had failed for so long that module-level validation was
+# not running at all.
 test-modules:
 	@echo "Validating all D4D module schemas..."
 	@for module in $(SOURCE_SCHEMA_DIR)D4D_*.yaml; do \
 		echo "Validating $$module..."; \
-		$(RUN) gen-project -d tmp $$module || exit 1; \
+		$(RUN) gen-project -I python -I jsonschema -I jsonld -I owl -d tmp $$module || exit 1; \
 	done
 	@echo "All D4D module schemas validated successfully!"
 
--- a/project/jsonld/data_sheets_schema.jsonld
+++ b/project/jsonld/data_sheets_schema.jsonld
@@ -1778,7 +1778,7 @@
     },
     {
       "name": "FileTypeEnum",
-      "definition_uri": "https://w3id.org/bridge2ai/data-sheets-schema/file-collection/FileTypeEnum",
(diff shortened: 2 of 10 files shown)


realmarcin:
**Review pass — no defects found.** What was checked:

- **Merged-schema diff is exactly 7 changes, all intended**: the base slot's
  range removed, `DatasetCollection` gaining the range/description/multivalued
  it had been inheriting, and `FileCollection.resources` moving `Dataset` ->
  `File`. Nothing else in the merged schema moves.
- **No record changes validity.** Validated all 92 sweep records against both
  branches: 69/92 on `main`, 69/92 here, and **zero** records differ in outcome.
- **The narrowed generator set still catches real breakage.** Injected an
  unresolvable `range: NoSuchClass` into `D4D_Motivation` and `make
  test-modules` failed on it — so dropping shaclgen does not make the target
  vacuous. It still catches precisely the class of defect that broke it
  originally.
- **No class is left with an untyped `resources`.** All four resolve to a class
  range, none falls back to `string`.

One thing worth stating rather than filing: this narrows module validation to
python/jsonschema/jsonld/owl, so a module that breaks *only* shaclgen, gen-excel
or gen-markdown will now pass. That is deliberate — the project builds exactly
those four, `gen-project` above uses the same set, and validating a generator
nobody builds produced a failure nobody could act on, which is how the target
came to be ignored. If SHACL is ever built for real, the generator's
`KeyError: data-sheets-schema-base` on imported schemas needs solving first.

#226 was filed during development and stands on its own.

Issue #215, opened 2026-08-01
17 runs fail validation on date-time formatting, mostly /issued


realmarcin:
Surveying every provenance record on disk: **17 runs record
`validation.passed: false`**, against 105+ that pass. The failures are dominated
by one avoidable shape.

## Breakdown

| problem | count |
|---|---|
| value "is not a 'date-time'" | 11 |
| no matching schema (union) | 2 |
| Python `datetime` object leaked into YAML | 2 |
| other | 7 |

| slot | count |
|---|---|
| `/issued` | 10 |
| `/related_datasets/0/relationship_type` | 6 |
| `/dialect`, `/license_and_use_terms`, `/at_risk_populations`, `/distributions/*` | 1 each |

## Root cause for `/issued`

`issued` declares `range: datetime`. The overwhelming majority of records emit a
correctly zoned value and validate:

    14  '2026-04-03T00:00:00Z'
    13  '2026-06-17T00:00:00Z'
    13  '2026-05-01T00:00:00Z'

The failures are the minority that drop the zone or the time:

     3  '2026-04-03T00:00:00'      <- naive, no timezone
     2  '2026-06-30'               <- date only

Plus two records where an unquoted YAML scalar was parsed into a Python
`datetime` object before validation ever saw a string.

So this is not a schema mismatch — the schema is being satisfied most of the
time by the same generator on the same slot. It is output variance on a
mechanically checkable format.

## Options

1. **Coerce on write.** Normalise anything date-shaped to RFC 3339 with an
   explicit zone before the artifact is written. Deterministic, fixes the
   `datetime`-object case for free, and needs no prompt change — so no prompt
   hash churn (cf. #214).
2. **Say it in the prompt.** Cheap, but probabilistic, and changes the prompt
   hash — which splits any arm mid-flight.
3. Widen the schema to accept a bare date. Rejected unless a date-only `issued`
   is genuinely meaningful, which would be a modelling decision rather than a
   fix.

Option 1 preferred.

## Not urgent

These runs produced usable records; validation flags them rather than losing
them. Worth fixing before the corpus is published or scored on completeness.

realmarcin:
Fixed on `fix-prompt-date-and-issued`, but the diagnosis in this issue was
**half wrong** and the correction matters.

I claimed the schema was fine and this was "output variance on a mechanically
checkable format". Verified against `linkml-validate` directly:

| value | result |
|---|---|
[...]

PR #228, opened 2026-08-02
Profile the sweep's validation failures instead of repairing them (#215)

realmarcin:
#215 asks what to do about 23 records that fail validation. The answer is
nothing — and that is worth writing down rather than leaving the issue open.

## Why not fix them

**Editing them would falsify the experiment.** These records are what the
generator produced under a stated prompt, model and temperature. That is the
observation. Rewriting `IsNewVersionOf` to `is_new_version_of` makes the corpus
a record of what we wish had been generated, and every quality figure computed
from it becomes uninterpretable.

**It would also break provenance.** Each run pins its artifacts' hashes, so
editing the bytes makes `d4d runs check` report the run as *drifted* — 23
validation errors traded for 23 broken provenance records, which is a worse
claim to make about a study corpus.

The failures are evidence about generation quality. The response is to fix the
generator and rerun, which the forward fixes already do.

## What the profile shows

69/92 validate (75%). Two things not visible from the raw failure list:

| arm | invalid / total |
|---|---|
| baseline | 10 / 50 (20%) |
| **de_novo** | **8 / 18 (44%)** |
| crate_only | 5 / 18 (28%) |
| healthsheet | 0 / 6 (0%) |

**de_novo fails about twice as often as baseline**, healthsheet not at all. n is
small — 3 records per project per arm, 3 healthsheet records total — so this is
a signal to follow, not a result.

**Core records fail less than their full counterparts in every arm**, consistent
with `CoreDataset` being the smaller target: 79 slots against 94.

By cause, 20 of 23 failures are temporal format (11) or enum values (9), both
fixed forward. The nine enum failures split into 6 DataCite spellings — the
model reaching for a vocabulary it knew because the digest never showed it the
declared one — and 2 outright inventions, which no alignment repairs.

## Falsifiable

If the forward fixes work, a rerun under the current schema and digest should
land near 3 remaining failures rather than 23. If it does not, they did not
land, and this note is the baseline that says so.

`scripts/validation_profile.py` regenerates the whole thing, no API calls.

### DIFF
--- a/notes/validation_profile_2026-07-31.md
+++ b/notes/validation_profile_2026-07-31.md
@@ -50,6 +50,15 @@ and fewer object-ranged slots to get wrong.
 
 ## By cause
 
+**These count records, and a record may have several causes, so they sum to 28
+rather than 23.** Five records carry more than one:
+
+| causes | records |
+|---|---|
+| 1 | 18 |
+| 2 | 4 |
+| 3 | 1 |
+
 | records | cause | fixed forward? |
 |---|---|---|
 | 11 | temporal format | yes — `normalise_temporal()` on write |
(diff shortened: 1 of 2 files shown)


realmarcin:
**Review pass — one defect found and fixed, in this PR's own note.**

#231: the by-cause table sums to **28** against **23** failing records, with
nothing saying why. Five records carry more than one cause.

That was not presentational. The note predicted a rerun would leave "near 3"
failures — arrived at by adding the *cause* counts (11 temporal + 9 enum = 20)
and treating the sum as records. Counting records whose **every** cause is fixed
forward gives **16**, so a rerun should leave **7**.

The number is the falsifiable claim the note exists to make. Predicting 3 and
observing 7 would have read as the fixes underperforming, when 7 is what the
evidence actually predicts.

`scripts/validation_profile.py` now reports the causes-per-record distribution
and computes the clearable count directly, so the two quantities cannot be
confused again by arithmetic in prose.

Also verified in review:

- **The profile still holds after #227 and #229 merged**: 69/92, and every
  per-arm figure unchanged. Re-ran after rebasing onto current `main`.
- Per-arm percentages check out: baseline 10/50 (20%), de_novo 8/18 (44%),
  crate_only 5/18 (28%), healthsheet 0/6.
- The script makes no API calls and does not write into `data/`.

897 passing.

Issue #176, opened 2026-07-29
Design: produce a canonical D4D from N stochastic replicates (coverage merge, not selection)


realmarcin:
## The gap

Generation is stochastic. We run three replicates per condition and measure how much they agree — but nothing ever picks one. After a sweep there is **no answer to "which record *is* the CHORUS datasheet"**. Replicates are a measurement device; the pipeline has no production output.

`reconcile` does not fill this gap. It operates **within** a single run, making that run's full and core consistent with each other and with the audit findings. Replicates never meet. Nothing in `runs.py` selects or merges — `compare`, `arm_delta`, `check_replicate` all measure.

> **This issue was rewritten on 2026-07-29 after the selection score was built and measured.** The original proposal — score each replicate, select the winner — rests on a premise that did not survive measurement. What replaced it is below; the original plan and the measurements that overturned it are in the comments.

## What was measured

`src/data_sheets_schema/evidence_score.py` now exists, with two scoring axes and the partition machinery. Run against the CM4AI and AI-READI generic replicates:

**1. Grounding ("is this value supported by the source documents?") cannot rank.** 78 judgements on CM4AI spanned 0.951–0.958 of slot count for *all three* replicates and reproduced the free presence ranking exactly.

The reason is structural, not calibration: the generator was handed those documents, so asking whether its output traces back to them is close to tautological. Across 78 judgements nothing scored below 0.60.

**2. Fitness ("does this value satisfy the field as the schema specifies it?") discriminates well — but only per field.**

| axis | range | sd | below 1.0 | diagnoses? |
|---|---|---|---|---|
| grounding | 0.85–1.00 | ~0.03 | ~0 | no |
| fitness | 0.40–1.00 | **0.121** | 73/78 | form 8, target 7, substance 2 |

Confirmed on AI-READI (sd 0.139, min 0.25), so it is not a CM4AI artifact.

**3. But fitness ranks records identically to counting slots.** Per-record mean fitness is flat:

```
rep3  fit_slots=58.78  mean=0.8519  n=69
rep2  fit_slots=54.13  mean=0.8592  n=63
rep1  fit_slots=51.13  mean=0.8522  n=60
```

A spread of 0.007 in mean quality. rep2 has the *highest* mean and still ranks second, because the sum is driven by count. Grounding gave the same ordering, as does slot count, for free.

[...]

PR #229, opened 2026-08-02
Expose the coverage merge as `d4d runs merge` (#176)

realmarcin:
#176 opens with: after a sweep there is no answer to "which record *is* the
CHORUS datasheet". `merge.py` has had `union_merge` and `write_merge` since the
selection work, but **nothing called them**, so the gap was still open.

`d4d runs merge --project CHORUS --config <config>` takes every replicate,
unions their slots, and writes a `record_mode: derived` record whose provenance
names each contributor by hash. Dry run by default.

## Measured, and it argues against overselling this

| project | best single | merged | gain |
|---|---|---|---|
| AI_READI | 70 | 74 | +4 |
| CHORUS | 50 | 53 | +3 |
| CM4AI | 72 | 77 | +5 |
| VOICE | 79 | 80 | +1 |

The gain is small, and the reason matters. #176's rewrite rests on "replicates
differ in coverage, not quality". That is only weakly true. Of the slots two or
more replicates both populate:

| project | shared slots | values agree | values differ |
|---|---|---|---|
| AI_READI | 70 | 12 | 58 |
| CHORUS | 48 | **1** | 47 |
| CM4AI | 70 | 16 | 54 |
| VOICE | 78 | 9 | 69 |

**77–98% of shared slots carry different values.** CHORUS agrees on one slot out
of 48. So replicates differ enormously in *content*, and only marginally in
coverage — the union captures the marginal part.

What the merge produces is therefore the base record plus the handful of slots
only the others had. It does not adjudicate between differing values, and
without a scorer it cannot. The earlier measurement in this issue is why there
is no scorer: neither axis could discriminate replicates.

## The output says so

An earlier draft reported a "contested" count, which reads as unresolved
disagreement. It is not unresolved — the base's value is used — and an operator
shipping the record needs to know that, so the command states it plainly.

## Refusals are refusals, not tracebacks

`check_sources` declines to combine runs whose conditions cannot be established:
a derived record inherits its contributors' standing, so merging an unattested
run would launder it. That surfaced as a stack trace; it is now a clean error.

## Testing

895 passing. New `tests/test_cli/test_runs_merge.py` covers the dry run writing
nothing, the union covering slots no single replicate had, refusing a
single-replicate "merge", naming a missing record rather than skipping it, and
that the written record is `derived` and never `live`.

## Still open in #176

Nothing downstream consumes a canonical record yet — `d4d render generate-all`
is label-unaware. And whether a base-plus-margin union deserves to be called
canonical is a judgement this PR deliberately does not make; it provides the
operation and reports honestly what it does.

### DIFF
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -262,6 +262,14 @@ def compare(method, project, labels, require_live, require_attested):
         click.echo("   Only the hardware is unrecorded, which cannot affect a "
                    "generation.\n")
 
+    if result.get('excluded_derived'):
+        # Say it, rather than letting the record vanish. A silent exclusion
+        # leaves the operator to wonder whether the merge ran at all.
+        click.echo(f"ℹ️  excluded as derived (built from the runs being "
+                   f"compared, so including it would bias the agreement it is "
+                   f"measured against): "
+                   f"{', '.join(result['excluded_derived'])}\n")
+
     if result.get('excluded_incomplete'):
         click.echo(f"⚠️  excluded as still running / incomplete: "
                    f"{', '.join(result['excluded_incomplete'])}\n")
@@ -465,5 +473,6 @@ def merge_cmd(method, project, labels, config, out_label, unguarded, execute):
     write_merge(result, target, sources=sources, project=project,
                 method=method, label=out)
     click.echo(f"\nWrote {target}")
-    click.echo("record_mode: derived — not a generated record, and excluded "
-               "from replicate agreement statistics by construction.")
+    click.echo("record_mode: derived — not a generated record. `d4d runs "
+               "compare` excludes it from agreement figures and says so, "
+               "because it is built from the runs those figures measure.")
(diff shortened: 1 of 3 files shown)


realmarcin:
**Review pass — one defect found and fixed, and it falsified a claim this PR
was printing.**

#230: `compare()` already excluded derived records with the right rationale, but
the check ran **after** `is_complete()`. A merged record has no core and no
reconciliation report by construction — it is a union of full records, not a
generation run — so it never reached the derived branch and was reported as
"still running / incomplete", sending the reader after a run that would never
arrive.

Worse, the CLI never printed `excluded_derived` at all. With the ordering fixed
the record would have disappeared from `compare` **silently**, which is worse
than a wrong label: nothing would tell the operator the merge had happened.

That also falsified this PR's own output. It printed "excluded from replicate
agreement statistics **by construction**". Nothing excluded it structurally — it
fell out of looking incomplete and having a differing `config` string, either of
which could have changed without anyone noticing a derived record had started
counting as a replicate. The claim is now true, and points at the command that
enforces it.

Also verified during review:

- `--execute` writes both the record and a `record_mode: derived` provenance
  record naming each contributor.
- `check_sources` refusals surface as clean errors, not tracebacks.
- The review artifact written into `data/` while testing `--execute` has been
  removed; the corpus is untouched by this PR.

897 passing.

PR #232, opened 2026-08-02
Make FileCollection.resources populatable (#226)

realmarcin:
`FileCollection.resources` declared `any_of: [File, FileCollection]` alongside a
concrete `range`. LinkML **ANDs** the two when generating JSON Schema, so the
effective constraint was `range AND (File OR FileCollection)`:

- with the inherited `range: Dataset` — unsatisfiable for *every* value;
- with `range: File` — admits a File, rejects the nested FileCollection the
  union existed to allow.

Either way the recursive shape the description promised was unreachable.

## The change, concretely

This record **fails on `main`** and **validates here**:

```yaml
file_collections:
  - id: https://example.org/fc
    path: /d/fc
    resources:
      - id: https://example.org/f1
        path: /d/fc/a.csv
```

That is the whole of it: a slot that could not be populated now can be.

## Why narrowed rather than fixed as a union

Nothing was lost. Across **658 `file_collections` entries** in the corpus,
**zero** populate `resources` and none nests — the recursive shape was never
exercised by anything.

Expressing the union properly needs an abstract parent over `File` and
`FileCollection`. They already share `is_a: Information`, but `Information` also
parents `Dataset` and `DatasetCollection`, so ranging on it would be looser than
the union it replaces. A new intermediate class would be the modelling fix — and
this schema has **no abstract or mixin class among its 78**, so that is a new
idiom introduced for a capability nothing uses.

A test pins that observation. If someone later adds an abstract class, the
choice becomes visible rather than incidental, and this slot is the first place
to revisit.

## The description was also wrong

It promised "nested FileCollection objects" while the schema rejected them.
Documenting a shape that cannot validate is worse than a narrower one, because
it sends a generator after it. It now says plainly that a collection does not
nest, and points at `file_collections` on the dataset for grouping.

## Scope

Merged-schema diff is **4 changes, all in this slot** — `any_of` removed and the
description reworded, in both the attribute and the slot_usage. Corpus validity
unchanged at 69/92; these records do not use the slot.

## Testing

905 passing, `make test-modules` passes. New
`tests/test_schema/test_filecollection_resources.py` covers the record that
previously failed, the ordinary no-resources case (658 of 658 corpus entries),
that no `any_of` returns beside a concrete range, and that the description no
longer promises nesting.


realmarcin:
**Review pass — no defects found, and one consequence worth recording.**

The most important check was whether *code*, not just data, depends on the
nesting being dropped. It does not — and better than that, every path already
assumed `File`:

- `src/validation/unified_validator.py:231` sets
  `file_collection['resources'] = [file_obj]` — a **File**, as part of the
  Dataset-level deprecation migration.
- Both fairscape TODOs
  (`d4d_to_fairscape.py:292`, `fairscape_to_d4d.py:289-291`) describe converting
  `FileCollection.resources` as **"File objects"**.

So the narrowing matches what the codebase already intended; the union was
documentation nothing implemented.

**The fix repairs a live path.** The exact shape `unified_validator` emits fails
validation on `main` and validates here:

| ref | errors on the migration shape |
|---|---|
| `main` | 1 |
| this branch | 0 |

That path exists to migrate deprecated Dataset-level file properties into a
FileCollection — and it was producing records that could never validate. Nobody
would have noticed until a migrated record was checked.

Also verified:

- Merged-schema diff is 4 changes, all confined to this slot.
- Corpus validity unchanged at 69/92 — these records do not use it.
- `make test-modules` passes; 905 tests pass.

**A process note against myself.** While comparing branches I ran `git stash`
with nothing uncommitted to save, so `git stash pop` popped a *pre-existing*
stash that was not mine and left conflicts. Nothing was lost — the stash was
kept, the conflicts were only `.DS_Store`, and I resolved them narrowly rather
than with a hard reset so the unrelated `aurelian` change stayed intact. The
comparison was then redone with `git checkout`, which is what I should have used
in the first place. The "on main" figure in the table above is from the corrected
run; my first attempt had silently compared the branch against itself.

realmarcin:
**Second review pass — one defect found and fixed, one filed.**

Went deeper than the first pass, into generated artifacts and the
semantic-exchange layer.

**Fixed here:** dropping the union left two files describing a shape the schema
now rejects —

- `d4d_rocrate_sssom_mapping.tsv:111` — "Multivalued union(File, FileCollection)"
- `d4d_rocrate_structural_mapping.sssom.tsv:153` — range "File or FileCollection"

Same failure this PR fixes in the schema, one layer out: documentation promising
a shape nothing can produce. Both now say `File`.

**Filed as #234, not fixed here:** regenerating the structural mapping looked
like the clean way to correct it, and instead **dropped 10 rows** — `CoreDataset`,
`CoreDatasetCollection`, `CoreDistribution`, `DataSubset`, `File/file_type` and
five more. `make gen-sssom-structural` emits 150 rows against the committed 160,
so that file cannot be rebuilt from its declared inputs. Conspicuously, the
`Core*` losses are classes that live in the *core* merged schema, which the
target does not list as an input. Correcting a two-cell staleness is not the
moment to lose ten unrelated mappings, so the cells were edited directly and the
file restored byte-for-byte otherwise (160 rows before and after).

**Also verified clean this pass:**

- Generated Python datamodel types the slot `list[File]` — no `Dataset`, no
  union.
- Generated JSON Schema has **no `anyOf`**, `items: {$ref: File}`.
- No example or test data populates the slot, so nothing to migrate.
- Nothing in `src/` expects a `FileCollection` inside `resources`; every mention
  says "File objects".

905 passing.

PR #233, opened 2026-08-02
Make `d4d render generate-all` render, and know about run labels (#176)

realmarcin:
#176 stayed open because nothing downstream consumes a canonical record. Looking
at the downstream, there were **two** problems rather than one.

## It generated nothing

`d4d render generate-all` printed:

    ℹ️  Bulk HTML generation:
       Use: make gen-d4d-html

There is no `gen-d4d-html` target in the Makefile — only two help lines
mentioning it. The instruction pointed nowhere.

The script it otherwise pointed at, `src/html/render_concatenated.py`, read
`data/sheets_concatenated` — a directory that no longer exists, since the corpus
moved to the run-labelled layout. It exited **0** with "No alldocs YAML files
found". Nothing referenced it, so nobody noticed.

## Neither knew about run labels

Output was `data/d4d_html/concatenated/{method}/{PROJECT}.html`. No room for a
label, so a project rendered from two replicates overwrote itself.

Now `.../{method}/{label}/{PROJECT}.html`. Verified on the real corpus — CHORUS
from generic-v2 rep1 and rep2 coexist as distinct renders (59,362 and 56,829
bytes) where before the second replaced the first.

## Behaviour

    d4d render generate-all                    # list what would be rendered
    d4d render generate-all --execute          # render every record
    d4d render generate-all --method claudecode_agent --label <label> --execute

Dry run by default: the corpus holds **141 records across 118 method/label
directories**, and rendering all of them is not something to do by typo. An
empty selection is an error, not a silent success.

## `--publish`, and where this meets the canonical question

The docs build copies from the flat unlabelled path, which has no room for a
label. So `--publish` writes the flat copy too — and warns when more than one
label is selected, because otherwise sort order decides what ships:

    ⚠️  --publish with 2 labels: the flat copy has no room for a label, so later
       labels overwrite earlier ones and which one survives depends on sort
       order. Publish one label at a time.

That is the seam where a canonical record gets published, once #176's remaining
question — *which* label is canonical — is settled. This PR does not answer it,
and deliberately does not guess.

## The superseded script

Kept as a signpost pointing at the replacement rather than deleted. Its quiet
success is the trap: anyone finding it would reasonably conclude there was
nothing to render. It now exits non-zero and says what to run instead.

## Testing

905 passing. New `tests/test_cli/test_render_generate_all.py` covers the dry run
writing nothing, that it actually generates, that two labels do not collide,
filters, the empty-selection error, and both publish paths.

### DIFF
--- a/src/data_sheets_schema/cli/render.py
+++ b/src/data_sheets_schema/cli/render.py
@@ -2,6 +2,7 @@
 
 import click
 import json
+import collections
 import sys
 from pathlib import Path
 from data_sheets_schema.constants import METHODS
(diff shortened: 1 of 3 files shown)


realmarcin:
**Review pass — one defect found and fixed, one observation filed.**

**#235, fixed here.** `--publish` wrote `{PROJECT}.html`, but the published
directories contain `{PROJECT}_d4d.html` and the docs build copies every
`*.html` from them. So publishing would not have replaced the stale render — it
would have shipped **both**, under two names, with nothing saying which was
current. That defeats the flag, whose whole purpose is to make the docs show a
chosen label.

Both paths now name the output from the record (`{PROJECT}_d4d.yaml` ->
`{PROJECT}_d4d.html`), so publishing overwrites the served file and the labelled
and flat layouts agree. Verified against the real corpus: publishing CHORUS
replaces `CHORUS_d4d.html` and leaves no stray `CHORUS.html`.

**Also verified:**

- Label subdirectories do **not** break the docs build. `cp -r .../*.html` does
  not match directories, so the existing published files keep shipping
  untouched — checked directly rather than assumed.
- Rebased onto `main` after #232 merged; 914 passing.
- The HTML written into `data/` while testing has been removed; `git status`
  shows the corpus untouched by this PR.

**Noticed, filed in #235's body rather than fixed:** the flat directories have
accumulated at least four naming conventions —
`AI_READI_d4d.html`, `AI_READI.html`, `AI_READI_d4d_human_readable.html`,
`AI_READI_curated_human_readable.html` — some from renderers that no longer run.
Since the docs build globs `*.html`, all of them ship. Worth a tidy-up; out of
scope here, which only needed to stop adding to the pile.

Issue #219, opened 2026-08-01
related_datasets.relationship_type receives DataCite vocabulary the schema does not permit


realmarcin:
Split out of #215, which turned out to contain two unrelated classes of failure.
The date-formatting half is fixed; this is the rest.

After normalising temporal values, 12 of the 23 previously-failing artifacts
still fail. The largest group is enum values the schema does not accept, all in
`related_datasets.relationship_type`:

| value emitted | source |
|---|---|
| `IsNewVersionOf` | DataCite `relationType` |
| `Continues` | DataCite |
| `References` | DataCite |
| `related_to` | plausible-looking invention |
| `is a later release in the same series as` | free text |

The schema permits `derives_from`, `supplement_to`, and others in that style.

## Reading

The first three are real DataCite `relationType` values. The generator is
reaching for a vocabulary it knows rather than the one the slot declares — the
input bundles contain DataCite metadata, so the values are *plausible* and
*wrong*, which is the hardest kind to catch by eye. `related_to` and the
free-text one are not from any vocabulary.

This is the `target` failure class from the fitness work: answering with the
right idea in the wrong vocabulary. Note the v2 prompt rules reduced `target`
failures by ~60% but do not address enum vocabularies specifically.

## Options

1. Put the permitted values in the schema digest sent with the prompt, if they
   are not already there — the model cannot pick from an enum it has not been
   shown.
2. Map the common DataCite relations onto the schema's enum at write time, the
   way `normalise_temporal()` handles dates. Only safe where the mapping is
   unambiguous; `IsNewVersionOf` -> `derives_from` is arguable, and a wrong
   mapping silently records a false relationship, so this needs care.
3. Leave it to validation and fix by hand.

Option 1 first — it addresses the cause rather than the symptom, and does not
risk inventing a relationship the documents never claimed.

## Remaining, not enum-related

2 union "no matching shape", 1 wrong type (expected string), 3 other. Worth
looking at once the enum group is dealt with.

realmarcin:
Addressed by #223 (merged), taking the first of the three options proposed here.

**The vocabulary is now in the schema.** All 36 DataCite RelationType values are
represented, up from 14. `Continues` — one of the values a run actually emitted
— previously had nowhere to go at all. Each value carries its DataCite term in
`aliases`, so the alignment is machine-readable rather than asserted in prose,
and tests check it is total in both directions and one-to-one.

**And the digest now shows it.** This was the actual cause. `DatasetRelationship`
[...]

PR #236, opened 2026-08-02
Select one replicate as canonical, keeping all of them (#176)

realmarcin:
Selection, not merging — reversing the direction #176 took on 2026-07-29.

## Why the reversal

That rewrite was right that **scored** selection is not worth building: fitness
and grounding both rank replicates identically to counting slots, so paying 78
judgements to reproduce `len(record)` buys nothing. It was explicit about this:

> Selecting by score reduces to selecting by coverage, which needs no API calls.

But it made a coverage-maximising **merge** the primary line, and the merge has
since been measured (#229):

| project | merge gain | shared slots | values agree |
|---|---|---|---|
| AI_READI | +4 | 70 | 12 |
| CHORUS | +3 | 48 | **1** |
| CM4AI | +5 | 70 | 16 |
| VOICE | +1 | 78 | 9 |

**77–98% of shared slots carry different values.** So a merge mostly reproduces
its base record while splicing values from records that disagree — and a spliced
record can assert a participant count from one referent and a DOI from another,
which is the risk #176 flagged from the start. One replicate is internally
coherent because one generation produced it.

## Validity first, coverage second

Coverage alone is nearly arbitrary here. Margins across generic-v2 are **+0, +1,
+2, +1**, and AI-READI is an outright tie. Validity is what discriminates:

    AI_READI  rep1 valid, rep2/rep3 invalid   -> resolves the tie
    CHORUS    all three valid                 -> coverage, +1, flagged as thin
    CM4AI     rep3 invalid                    -> drops a 70-slot candidate
    VOICE     nothing validates               -> refuses to select

VOICE is why the ordering matters. A coverage rule would have confidently made
an invalid record canonical. This declines — picking the least-bad ships
something known to be broken.

## Validates live, not from the record

`compare()` reads validation status from provenance because it runs over 100+
runs in a hot path. Selection touches three, once, so it can afford the truth.

It matters: **CM4AI rep1 is recorded `invalid` but validates against the current
schema** — the DataCite alignment (#219) admits a value rejected when the run
was made. Selecting on the stale claim would have excluded a good record. Where
live and recorded disagree the command says so, because that disagreement means
the schema moved under the corpus.

## Nothing is moved

The winner's provenance gains a `canonical` block naming every candidate, its
slot count and both verdicts. Auditable, reversible, and replicates stay a
measurement device rather than being collapsed into one.

## Also fixed

`is_complete`, `validation_status` and `record_path_for` bind `concat_dir` as an
import-time default, so they answered about the default corpus regardless of
where the replicates were. Latent until something used a different one.

## Testing

922 passing. New `tests/test_cli/test_runs_select.py` covers validity
outranking coverage, the refusal when nothing validates, thin-margin reporting,
dry run writing nothing, the canonical block naming the losers, that nothing is
moved, and that one replicate is not a selection.

## Note on `d4d runs merge`

Left in place. It is honest about what it does and useful for asking "what would
full coverage look like", but it is no longer the answer to "which record *is*
the CHORUS datasheet".

### DIFF
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -476,3 +476,143 @@ def merge_cmd(method, project, labels, config, out_label, unguarded, execute):
     click.echo("record_mode: derived — not a generated record. `d4d runs "
                "compare` excludes it from agreement figures and says so, "
                "because it is built from the runs those figures measure.")
+
+
+def _validates(record: Path) -> bool:
+    """Run the validator, rather than trust a status recorded earlier."""
+    import subprocess
+    schema = "src/data_sheets_schema/schema/data_sheets_schema_all.yaml"
+    try:
+        return subprocess.run(
+            ["poetry", "run", "linkml-validate", "-s", schema, "-C", "Dataset",
+             str(record)], capture_output=True, text=True,
+            timeout=300).returncode == 0
+    except Exception:                                   # noqa: BLE001
+        return False
+
+
+@runs.command("select")
+@click.option("--method", default="claudecode_agent",
+              help="Method directory the replicates live under.")
+@click.option("--project", required=True)
+@click.option("--config", required=True,
+              help="Config prefix whose replicates are the candidates, e.g. "
+                   "2026-07-31_claude-opus-5-generic-v2.")
+@click.option("--allow-unverified", is_flag=True,
+              help="Consider replicates whose validation status is unknown. "
+                   "Off by default: absence of evidence is not validity.")
+@click.option("--execute", is_flag=True,
+              help="Record the choice in the winner's provenance. Without "
+                   "this, report and write nothing.")
+def select_cmd(method, project, config, allow_unverified, execute):
+    """Mark one replicate canonical, keeping all of them.
+
+    Selection, not merging. Merging splices values across replicates that
+    disagree on 77-98% of the slots they share, for a coverage gain of 1-5 slots
+    (#229) — and a spliced record can assert a participant count from one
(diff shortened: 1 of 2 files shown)


realmarcin:
**Review pass — one defect found and fixed.**

**#237: selection judged a run on its full record alone.** `select` validated
`{project}_d4d.yaml` against `Dataset` and nothing else, but a run ships **two**
records and marking it canonical marks both. A run whose full record validates
and whose core does not could be selected, and the command would report it
clean.

Not hypothetical — across the generic-v2 config there is exactly one such
replicate:

    CM4AI rep1   full: valid   core: INVALID

It is excluded today only because it loses on coverage (67 slots against 72).
Reverse those counts and it becomes canonical while shipping a broken core.

Both records are now validated, both must pass, and the report names which one
failed:

    ✗ rep1  67 slots  invalid (core)
      rep2  72 slots  valid
    ✗ rep3  70 slots  invalid (full)

"invalid" alone sends the reader to the wrong file.

**Also verified this pass:**

- Writing the `canonical` block does not disturb the attestation gate:
  `d4d runs check` still reports 128 runs, 46 subject, 0 failing, and
  `record_mode` stays `live`.
- The provenance header comment survives the rewrite — `ProvenanceRecord.write`
  preserves it, which an earlier sweep did not and stripped from 97 records.
- The winners are unchanged by the fix (AI_READI rep1, CHORUS rep2, CM4AI rep2),
  so this tightens the criterion without altering today's answers.
- The corpus is untouched: the provenance edited while testing `--execute` was
  restored, and `git status` on `data/` is clean.

923 passing.

Issue #169, opened 2026-07-29
n=3 cannot resolve effects near the noise floor; design is unbalanced across arms


realmarcin:
Two limitations that bound what the arms study can claim, recorded so they are not forgotten when results are written up.

### Replicates

Three-way noise runs **±13–18 slots**. CHORUS's crate verdict moved `real` → `marginal` between conditions because the *noise floor rose*, not because the effect changed (deltas +22/+16/+20 → +28/+12/+14). Effects of that magnitude are not resolvable at n=3.

Either raise replicates for arms whose delta sits near the floor, or state plainly in the writeup that those arms are unresolved rather than negative.

### Balance

- `healthsheet_only` exists for AI-READI alone
- crate arms cover 3 of 4 GCs (AI-READI has no RO-Crate yet)

So the only balanced cell is the baseline arm, 4 GCs × 3 conditions. Pooling across arms is not valid; per-arm reporting is.

### Pre-registration

The `real` / `marginal` / `not resolvable` thresholds in `arm_delta()` were applied after seeing counts. Fixing the metric and thresholds before the next batch would remove that degree of freedom.

realmarcin:
The n=3 question may be the wrong diagnosis. Measured today:

## No run in this corpus has ever applied temperature

All 46 runs of the 2026-07-31 sweep recorded:

    model=google/claude-opus-5-high  temperature=None
    basis="not applicable to this model"

The 81 runs elsewhere that *record* `temperature: 0.0` are the Claude Code agent
path, where that line is an assertion in a header rather than an applied
parameter — `api_runner`'s own docstring flags this ("Claude Code exposes no
temperature setting, so every existing header's `Temperature: 0.0` is an
assertion").

So every replicate in the study was sampled at the provider's default. The
77-98% value disagreement between replicates measured in #229 is, at least in
part, uncontrolled sampling rather than irreducible model variance.

## And the omission was ours, not the provider's

`NO_TEMPERATURE_MODELS` contains `claude-opus-5`, so `accepts_temperature()`
returns False and the runner omits the parameter. But the provider accepts it:

    google/claude-opus-5-high   temperature=0.0  -> 200 OK

(For contrast, `google/claude-sonnet-4.5` and `anthropic/claude-opus-4` return
400 **Invalid model** on CBORG — the names do not exist there. So "downgrade to
a model that allows temperature" is not the move; the current model already
allows it.)

## It is honoured, but does not give determinism

[...]

PR #238, opened 2026-08-03
Effort-tier experiment: the agreement metric is saturated (#169)

realmarcin:
Ran the experiment proposed for #169: does the effort tier change replicate
agreement? Three CHORUS replicates on the bare `google/claude-opus-5` route
against the three existing `-high` ones, everything else held constant.

**Only two opus-5 routes exist on CBORG.** `-low`, `-medium`, `-xhigh` and
`-max` return 400 *Invalid model* for this family — they exist for `sonnet-5`,
`opus-4-7` and `opus-4-8`, but not opus-5. So bare vs `-high` is the entire
available gradient, not a sample of it.

## Result

| arm | slot counts | shared | identical | rate |
|---|---|---|---|---|
| `-high` | 47, 49, 50 | 48 | 1 | 2.1% |
| bare | 45, 49, 50 | 49 | 2 | 4.1% |

One slot apart. But that cannot be read as "no effect":

| arm | scalar slots | list/dict slots |
|---|---|---|
| `-high` | 1/9 (11%) | **0/39 (0%)** |
| bare | 2/9 (22%) | **0/40 (0%)** |

**Zero of ~40 object-valued slots agree in either arm.** Exact equality over
nested objects carrying free-text descriptions essentially never holds, so the
measure cannot register a change in either direction. The only resolution is 9
scalar slots, where 1 against 2 at n=3 is noise.

So the experiment is **uninformative, not negative** — the instrument cannot
answer the question as posed.

## What it corrects

"Replicates disagree on 77–98% of the slots they share" — quoted in #229, #176
and #236 — is exact string equality on prose-bearing objects. Two records
describing the same collection method in different words count as disagreeing.

That was sound for its original purpose: showing a coverage merge cannot
adjudicate values, which still holds, because a merge must pick bytes and the
bytes do differ. It is **not** sound as evidence that the generator is unstable,
and #169's premise rests on that second reading.

## What would answer it

1. Scalar-only agreement — free, and the only part of the current metric with
   resolution.
2. Field-level judged equivalence — does value A state the same fact as value B,
   the way `evidence_score.py` already judges fitness per field. Costs calls,
   but on the axis that matters.
3. Presence-and-shape agreement — same slots, same cardinality, ignoring
   wording. Free, and closer to what a datasheet reader cares about.

Until one exists, "how much do replicates really differ" is unanswered and n=3
cannot be called adequate or inadequate.

## Cost

Three new runs, ~35 minutes. Worth it: a cheap experiment caught a measurement
problem that three open issues were building on.

`d4d runs check`: 131 runs, 49 subject, 0 failing.


realmarcin:
**Review pass — one defect found and fixed, and it makes the conclusion
stronger rather than weaker.**

**#239: the two arms did not receive the same prompt.** The note claimed
"everything held constant except the model route". They were not. The `-high`
replicates ran 2026-07-31, *before #220 merged*, when v2's `{DATE}` was still an
unsubstituted literal. The bare replicates ran after and got a real date:

    -high  # Generated: 2026-07-31   <- the model inferred it from {LABEL}
    bare   # Generated: 2026-08-03   <- substituted by resolve_prompt

Same prompt **file** in both arms (`4780613475ef2a62`), different resolved text.
Provenance confirms it: the `-high` runs carry no `prompts.resolved` hash,
because the field did not exist until #220.

One header line in ~4,400 characters is unlikely to move slot-level agreement,
but the note asserted the stronger claim, and a confounded design on top of a
saturated metric establishes nothing in either direction. Recorded in the note
rather than re-run — three more runs would buy a cleaner version of an
experiment that still cannot answer the question.

**Generalises beyond this PR:** absent `prompts.resolved` is the marker that an
arm predates #220 and differs in resolved prompt text from anything generated
after, even where the prompt file is identical. That affects every cross-date
comparison in this corpus.

**Also verified this pass:**

- **The headline numbers survive an independent re-derivation.** The published
  figures require all three replicates to match; computing agreement *pairwise*
  instead gives 3.6% (`-high`) against 6.1% (bare) — same near-zero picture,
  same ~2pp gap, so the result is not an artifact of demanding unanimity.
- Within-arm resolved-hash variation is just `{LABEL}` substitution, as
  expected, not instability.
- The three new runs pass the live-provenance gate; corpus-wide 131 runs, 49
  subject, 0 failing.
- Committed artifacts are the five expected files per run plus the note. **No
  progress files**, no stray output.

PR #240, opened 2026-08-03
Measure replicate agreement by judged equivalence (#169)

realmarcin:
#169 could not be judged adequate or inadequate because the quantity it rests on
was unmeasured. It now is.

## The headline number was wrong by about 18x

"Replicates disagree on 77–98% of the slots they share" — quoted in #229, #176,
#236 and #238 — is **byte equality** over values that are mostly nested objects
of free text. Measured on CHORUS, **zero of ~40 object-valued slots ever
matched**, so the measure returned the same answer regardless of input.

Judged on whether the values *state the same fact*:

| measure | CHORUS, 48 shared slots | rate |
|---|---|---|
| exact byte equality | 1 | 2.1% |
| **judged equivalent** | **18** | **37.5%** |

`title` is the clean case: three different strings, judged equivalent — "all
three name the same CHoRUS dataset, differing only in phrasing". Byte equality
counted that as disagreement.

The generator is substantially more stable than the figure implied. It is still
not stable: 62% of shared slots do carry different facts, which is a real result
rather than an artifact of the metric.

## Embedding similarity does not work as a proxy

Added deliberately to calibrate a cheap measure against the judge and then use
the cheap one at scale:

| slots | mean cosine |
|---|---|
| judged equivalent (n=18) | 0.923 |
| judged different (n=30) | 0.914 |

**A gap of 0.009.** It cannot separate the classes. Every value is schema-shaped
prose about the same dataset, so the embedding captures topic — and topic is
held constant by construction here.

The embeddings are fine in general: 0.865 for a paraphrase against 0.564 for
unrelated text on a hand-built pair. The failure is specific to this population.
Reported because it closes off the cheap route rather than leaving it as an
untested assumption — judged equivalence has to be paid for.

## Design notes

- One judge call **per slot**, not per pair — cheaper, and closer to the
  question ("do the replicates agree here" rather than "does rep1 agree with
  rep2").
- Judgements and embeddings cached content-addressed, keyed on the rubric
  digest, so re-running is free and a rubric edit invalidates rather than
  silently mixes.
- A truncated verdict is salvaged rather than discarded: thinking bills against
  the same budget as the answer, so a long deliberation can cut the closing
  brace off a decision already made. An unparseable verdict is **not** treated
  as agreement — that would inflate the very figure this exists to measure.

## Cost

44 judge calls and 127 embeddings for one project's three replicates.

## Not done

The other three projects and a second config. That is what would let
between-config differences be compared against within-config spread, which is
the actual #169 question. This PR establishes that the instrument exists and
reads sensibly.


realmarcin:
## Review

Read the diff adversarially and independently re-derived the result before reviewing the code.

**The headline claim holds.** I recomputed the matrix from the source records without using `matrix.json`: deltas −6.3 / −17.0 / +4.5 / +6.4, mean −3.1, sd 10.8, SE 5.4, within-config spreads 22.3 and 18.2. Every figure in the note checks out, and the argument — effect a third of the noise, deltas disagreeing in sign, ~100 projects needed — follows from them. The distinction the note draws between this quantity and fitness is the right one and is stated precisely.

Seven findings, filed as #241–#247.

### Fixed in this PR (2d91e356)

| | |
|---|---|
| **#241** | The note said "reproduce with agreement.py" and you could not — nothing called the module, and the note did not say which of three candidate directories held the inputs. I found `claudecode_agent` by trying all three (CHORUS v1/v2 shared slots: 49/43, **55/48** ✅, absent). Added a driver and a provenance table. |
| **#243** | Judge cache was not keyed on model; `Embedder`'s was. Fixed, with legacy records readable. Judge model recovered and recorded: `google/claude-opus-5-high`. |
| **#245** | No tests, in a repo with 923. Now 30, none touching the network. |
| **#246** | Dead `shape` property; now carried into every rows file. |
| **#247** | Per-slot rows existed for CHORUS only; now all four projects × both configs. |

### The substantive one — #244

Values are cut to 4000 chars before the judge sees them, silently. A cut can only hide disagreement, so it biases towards "equivalent", and it is **asymmetric between the configs being compared**: 21 truncated slots in v1 against 9 in v2, judged equivalent 43% against 89%. Across the matrix: 59/1576 values, 30/540 slots, longest value 29,024 chars of which the judge saw 14%.

**It does not overturn anything.** Worst case — every truncated-and-equivalent verdict flipped to "different" — moves the pooled rates 51.1/48.9 → 47.8/45.9, a delta of −1.9 instead of −2.2, still a fraction of the ±10.9 spread. The bias runs against v2, so correcting it shrinks the effect rather than rescues it. Now disclosed in the note with the sensitivity analysis, and flagged per slot.

### Deferred, with reasons on the issues

- **#242** — the cache key had no separator, so `("titl", ["ex"])` and `("title", ["x"])` collided. Scheme fixed going forward; the 434 published verdicts can't be re-keyed (the cache never stored the values) so eliminating the old scheme costs a full paid re-run. The set is frozen and cannot acquire new collisions.
- **#244**'s actual cap — same re-run, same reasoning; do both together when the matrix is next measured.
- **#247**'s `embeddings.jsonl` (1.4 MB) — kept, because it is what makes the 0.923/0.914 negative result reproducible rather than merely asserted.

### Gate

`python -m data_sheets_schema.agreement --offline --embed --write` reproduces all eight cells exactly from cache with zero paid calls. Suite: **953 passed, 3 skipped**.

PR #248, opened 2026-08-03
Remove the judge truncation cap and re-judge the affected slots (#244)

### LINKED ISSUE #244: agreement.py: judge input is silently truncated at 4000 chars, affecting 5.6% of the published slots
From the review of #240.

`EquivalenceJudge.__call__` truncates every value before showing it to the judge:

```python
blocks = "\n\n".join(f"--- value {i + 1} ---\n{t[:4000]}" for i, t in enumerate(distinct))
```

and `Embedder.vector` truncates at `text[:8000]`. Neither records that it happened. A slot whose replicates agree for 4000 characters and diverge afterwards is judged **equivalent** on evidence the judge never saw, and the output carries no flag distinguishing it from a slot judged on its full content.

This is live in the published matrix, not hypothetical. Measured over the eight project-configs actually reported:

| | |
|---|---|
| rendered values across shared slots | 1576 |
| exceeding 4000 chars (judge cap) | 59 (3.7%) |
| exceeding 8000 chars (embedder cap) | 17 |
| longest value | 29,024 chars — the judge saw 14% of it |
| shared slots with at least one truncated value | 30 of 540 (5.6%) |

Worse, the exposure is **asymmetric between the two configs being compared**:

| config | truncated slots | of which judged equivalent |
|---|---|---|
| v1 | 21 | 9 (43%) |
| v2 | 9 | 8 (89%) |

Truncation can only hide disagreement, so it biases toward "equivalent", and it biases v2 more than v1 — i.e. in the direction that would *shrink* the reported −3.1 delta.

**Sensitivity check: the headline conclusion survives.** Flipping every truncated-and-equivalent verdict to "different" (the worst case) moves the pooled rates from 51.1%/48.9% to 47.8%/45.9% — a delta of −1.9 instead of −2.2, still far inside the ±10.9-point between-project spread. The note's claim that the design cannot resolve this effect is unaffected.

Fix: raise or remove the cap, chunk long values, or at minimum set a `truncated` flag on `SlotAgreement` and report the count alongside the rate. Also worth a sentence in the note recording the cap and the sensitivity result above.

realmarcin:
Fixes #244. Follow-up to #240, which disclosed the cap but left it in place.

## What was wrong

Values were cut to 4000 characters before the equivalence judge saw them. A cut can only hide a disagreement that lives past it, so a truncated slot was biased towards "equivalent" — and the exposure was **asymmetric between the two configurations being compared**: 21 affected slots in v1 against 9 in v2.

## The trap that nearly ate the fix

Raising the constant on its own would have changed nothing measurable. The cache keyed verdicts on the **full values**, not on the truncated text actually sent to the judge, so every verdict reached under the 4000-char cap would have been served again under the new one. The correction would have reported a change of exactly zero and looked like vindication.

The key now hashes what the judge saw. Widening the cap moves the key for precisely the slots the cap bit and no others — **30 misses, 510 hits**, matching the 21/9 split measured during review. The legacy index needed the same guard: those records are keyed on full values but were *judged* on the first 4000 characters, so they are only reusable where nothing reached that cap.

## What the re-judge found

**Nine verdicts changed. Every one from "equivalent" to "different"; not one moved the other way** — precisely the direction truncation predicts.

| | published (4000-char cap) | corrected (full text) |
|---|---|---|
| pooled rate | 270/540 = 50.0% | **261/540 = 48.3%** |
| mean v1 | 51.1% | 49.6% |
| mean v2 | 48.1% | 46.7% |
| **mean delta** | **−3.1** | **−2.9** |
| sd of per-project deltas | 10.85 | 10.86 |

| cell | published | corrected | re-judged |
|---|---|---|---|
| v1 AI_READI | 62.0% | 58.2% | 6 |
| v1 VOICE | 39.7% | 37.2% | 9 |
| v1 CM4AI | 48.4% | 48.4% | 5 (no change) |
| v1 CHORUS | 54.5% | 54.5% | 1 (no change) |
| v2 AI_READI | 55.7% | 52.9% | 4 |
| v2 CM4AI | 52.9% | 51.4% | 3 |
| v2 VOICE | 46.2% | 44.9% | 2 |
| v2 CHORUS | 37.5% | 37.5% | 0 — unaffected |

CHORUS v2 had no truncated slots, so the note's headline section (18/48) and the 0.923/0.914 embedding table were measured on full text all along and are unchanged.

**#169's conclusion is unchanged**, and the bound stated before the re-run held: the worst case computed during review (−1.9, assuming every cut hid a real disagreement) bracketed the true −2.9 against the published −3.1. The effect moved 0.2 points; the noise did not move at all.

## Also

- Cap is 100,000 chars rather than absent — above the corpus maximum of 29,024, so it truncates nothing, but an unbounded prompt on a paid call is not a thing to leave lying around. `truncated` is now a property of `max_chars` so the flag tracks the current cap; `truncated_at_legacy_cap` keeps the #244 comparison available.
- Embedder cap stays at 8000 — that one is nomic-embed-text's 8192-token input limit, not an arbitrary choice, and the proxy it feeds was shown not to discriminate here anyway.
- `--embed-online`, off by default: without it a 30-call re-judge would have bought ~1400 embeddings on the way past, because `--embed` alone put the embedder online too.
- Cache records now carry `judge_chars`, so a future cap change is auditable rather than invisible.

## Verification

Canary first — one slot, same launcher, confirmed the verdict reached the cache file and carried both `model` and `judge_chars: 100000` before the other 29 ran. 30 calls, ~136k input tokens, 1.7× what the truncated prompts sent.

`--offline --embed` reproduces all eight cells with `CBORG_API_KEY` explicitly unset. Two new tests guard what was just fixed: no cell in the published matrix may be truncated, and every cell must record the cap and model it was judged under.

**961 passed, 3 skipped** (was 953).


realmarcin:
## Review

Reviewed the diff adversarially after opening it. Three residual findings, filed as #249–#251; two fixed in c7945d18.

- **#249 (fixed)** — the same defect as #244, at its limit: if the cap is narrow enough that two different values become byte-identical after truncation, the judge sees two identical blocks and answers "equivalent". Unreachable at 100k against a 29,024-char corpus maximum, so no published number moves, but it is a trap for anyone lowering the cap. Now raises.
- **#250 (fixed)** — `bool(None)` is False, so an unjudged slot counted as judged-and-different. Now `is True`, with `rate: null` and an `unjudged` count when anything is missing.
- **#251 (filed, not fixed)** — the embedder keys vectors on full text while truncating to 8000 for the request, structurally identical to the judge bug this PR fixes. Left alone deliberately: that cap is nomic-embed-text's 8192-token input limit rather than a choice, so there is no raise-and-re-measure available, and fixing it would orphan the 1.4 MB of vectors that make the "proxy does not work" result reproducible.

`matrix.json` regenerated offline for the new field; all eight cells unchanged. **966 passed, 3 skipped.**

Ready for your call on merging — I won't merge this one without your go-ahead.

realmarcin:
Note for whoever reviews this: #252 is stacked on this branch and fixes #251, which this PR's description says was deliberately left alone. Both reasons given here for leaving it were wrong — it orphans nothing, and the cap's stated justification (nomic-embed-text's 8192-token limit) is not what this endpoint does. It truncates at 2048 tokens and returns 200. Details in #252.

Issue #249, opened 2026-08-03
agreement.py: truncation that collapses two distinct values into one is silent


realmarcin:
From the review of #248.

When the cap is narrow enough that two genuinely different values become byte-identical after truncation, `EquivalenceJudge.__call__` hands the judge two identical blocks and the judge — correctly, on the evidence shown — returns "equivalent".

```python
with mock.patch("...JUDGE_VALUE_CHARS", 5):
    distinct = ["AAAAA_different_tail_X", "AAAAA_different_tail_Y"]
    _sent(distinct)  # -> ['AAAAA', 'AAAAA']
```

This is #244 in its purest form: the cut has hidden *the entire* disagreement rather than part of it, and nothing anywhere records that it happened. The `truncated` flag would be set, but the verdict reads as an ordinary "equivalent" and the rate absorbs it.

**Not reachable at the current cap** — 100,000 chars against a corpus maximum of 29,024 — so no published number is affected. It is a trap for whoever next lowers the cap for cost reasons, or runs this against a corpus with larger values.

`_parse_verdict` already establishes the right posture for this class of thing: refuse rather than quietly return the answer that flatters the measure. Same here — if the cap destroyed the only difference between two values, the measurement cannot be made at that cap and should say so.

realmarcin:
Fixed in c7945d18 (#248). `EquivalenceJudge.__call__` now raises when truncation collapses the distinct values into one, rather than sending the judge two identical blocks. Two tests cover it: the collapse raises without making a call, and values that remain distinct after the cap are judged normally.

Issue #250, opened 2026-08-03
agreement.py: build_matrix reports rate 0.0 when no judge ran, rather than None


realmarcin:
From the review of #248.

```python
"equivalent": sum(bool(r.equivalent) for r in result),
"rate": (sum(bool(r.equivalent) for r in result) / len(result) ...)
```

`SlotAgreement.equivalent` is `None` when no judge was supplied, and `bool(None)` is `False`. So a run without a judge reports every slot as judged-and-different: `equivalent: 0`, `rate: 0.0`. "Nothing was measured" and "nothing agreed" are the two most different possible readings of this instrument, and they serialize identically into `matrix.json`.

Not reachable through the driver — `build_matrix` always constructs a judge — so no published figure is affected. But `compare_records` takes `judge=None` as a documented option, and `matrix.json` is the artifact other things read.

Fix: count `r.equivalent is True` explicitly and return `None` for the rate when any verdict is missing.

realmarcin:
Fixed in c7945d18 (#248). Counting is now `r.equivalent is True`, `rate` is `None` unless every slot carries a verdict, and a new `unjudged` count is written to `matrix.json` so the distinction survives into the artifact. The driver prints `—` rather than crashing on a null rate. A test asserts every published cell has `unjudged == 0` and a non-null rate.

PR #252, opened 2026-08-03
The embedding endpoint truncates at 2048 tokens and returns 200 (#251)

### LINKED ISSUE #251: agreement.py: Embedder has the same key/cap mismatch that #244 was about
From the review of #248.

`Embedder.vector` keys the cache on the **full** text and then truncates for the request:

```python
key = _digest(text)                     # full text
body = ... "input": [text[:EMBED_VALUE_CHARS]]   # truncated
```

This is structurally identical to the judge bug fixed in #248: change `EMBED_VALUE_CHARS` and every vector computed under the old cap is served again under the new one, silently. 17 of the 1576 rendered values exceed the 8000-char cap, so it is not vacuous.

**Deliberately not fixed** in #248, for two reasons worth writing down rather than leaving implicit:

1. Unlike the judge's cap, this one is not arbitrary — nomic-embed-text takes 8192 tokens and 8000 characters sits inside that. There is no "raise it and re-measure" move available.
2. The proxy it feeds was shown not to discriminate at all on this population (0.923 against 0.914). Improving its inputs is not what is limiting it.

So the fix is cheap but buys nothing today, and applying it would orphan 1.4 MB of cached vectors that currently make the negative result reproducible. Filed so that if the embedding measure is ever revived, this is found before it produces a wrong number rather than after.

realmarcin:
Fixes #251. Stacked on #248 — base is `fix-244-judge-truncation`, so this diff shows only the embedder work. It will retarget to `main` when #248 merges.

## I deferred this for a wrong reason

#251 was filed with the note that fixing it "would orphan the 1.4 MB of vectors that make the negative result reproducible." One query settled it: **of the 136 cached vectors, zero were ever truncated.** They are all CHORUS v2, whose longest value is 2,876 characters. Keying on the text sent rather than the value it came from moves no existing key at all.

The second reason was also wrong, and this one matters more. I wrote that the 8000-character cap was "nomic-embed-text's 8192-token input limit, not an arbitrary choice." The endpoint is not that model's documented limit.

## What the endpoint actually does

| input | response | `prompt_tokens` |
|---|---|---|
| 8,000 chars | 200 | 1602 |
| 30,000 chars | 200 | **2048** |
| 40,000 chars | 200 | **2048** |

It truncates at 2048 tokens and says nothing — no error, no field, no warning.

The failure mode is not a degraded number, it is a maximally wrong one:

| | cosine |
|---|---|
| two values contradicting each other **past** the ceiling | **1.000000** — byte-identical vectors |
| the same contradiction, short enough to fit | 0.843 |

**The endpoint reports perfect agreement between two values that contradict each other, provided the contradiction is late enough.** For a module whose entire purpose is detecting whether two values state the same fact, that is the exact inversion of the measurement.

## The fix

A character cap cannot fix this — characters per token vary with the text, so no fixed number is safe for all input. What is reliable is the endpoint's own report: the client reads `usage.prompt_tokens` back and refuses to build a cosine on a prefix.

`PrefixOnlyEmbedding` is raised rather than returned, on the same reasoning as #249 and `_parse_verdict`: never quietly return the answer that flatters the measure. The refusal survives a round trip through the cache, so a stored prefix-vector cannot be reused later either.

Vectors are also now keyed on the text sent (mirroring #244), and cache records carry `chars`, `tokens` and `prefix_only` so a future cap change is auditable.

## The published table is untouched

CHORUS v2's longest value is 2,876 chars ≈ 719 tokens, about a third of the ceiling. Offline reproduction is byte-identical: **48/48 rows still carry similarity, 0.92272 against 0.91442.**

Across the whole corpus 17 of 1619 values would exceed the ceiling, none of them cached — a trap for reviving the measure, not a defect in what is reported.

## Verification

Live, both branches, on the corpus's longest real value (20,335 chars):

- at the 8000-char client cap → endpoint returns a genuine vector
- cap raised to 40,000 → `endpoint kept only 2048 tokens of a 20335-character value; the rest was discarded`

Five new tests pin the client's behaviour without calling the endpoint. **971 passed, 3 skipped.**

### DIFF
--- a/data/evaluation_llm/agreement_cache/matrix.json
+++ b/data/evaluation_llm/agreement_cache/matrix.json
@@ -7,6 +7,7 @@
   "truncated": 0,
   "truncated_at_legacy_cap": 6,
   "rate": 0.5822784810126582,
+  "similarity_absent": 69,
   "judge_model": "google/claude-opus-5-high",
   "judge_chars": 100000,
   "replicates": [
@@ -23,6 +24,7 @@
   "truncated": 0,
   "truncated_at_legacy_cap": 1,
   "rate": 0.5454545454545454,
+  "similarity_absent": 51,
   "judge_model": "google/claude-opus-5-high",
   "judge_chars": 100000,
   "replicates": [
@@ -39,6 +41,7 @@
   "truncated": 0,
   "truncated_at_legacy_cap": 5,
   "rate": 0.4838709677419355,
+  "similarity_absent": 59,
   "judge_model": "google/claude-opus-5-high",
   "judge_chars": 100000,
   "replicates": [
@@ -55,6 +58,7 @@
   "truncated": 0,
   "truncated_at_legacy_cap": 9,
   "rate": 0.3717948717948718,
+  "similarity_absent": 74,
   "judge_model": "google/claude-opus-5-high",
   "judge_chars": 100000,
   "replicates": [
@@ -71,6 +75,7 @@
   "truncated": 0,
   "truncated_at_legacy_cap": 4,
   "rate": 0.5285714285714286,
+  "similarity_absent": 58,
   "judge_model": "google/claude-opus-5-high",
   "judge_chars": 100000,
   "replicates": [
@@ -87,6 +92,7 @@
   "truncated": 0,
   "truncated_at_legacy_cap": 0,
   "rate": 0.375,
+  "similarity_absent": 0,
   "judge_model": "google/claude-opus-5-high",
   "judge_chars": 100000,
   "replicates": [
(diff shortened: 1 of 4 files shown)


realmarcin:
## Review

Three findings, filed as #253–#255, all fixed in 2efb8898.

- **#253** — a null similarity had three unrelated causes (never asked / not cached offline / refused as prefix-only) and `matrix.json` distinguished none. VOICE v1 had 74 of 78 rows null with the artifact silent, which reads as full coverage. Same failure #250 fixed on the judge side; the embedder never got the equivalent. Cells now carry `similarity_absent` — 434 of 540 across the matrix, CHORUS v2 the only complete column.
- **#254** — the error handlers wrote counters onto the injected embedder, so the failure paths demanded attributes the happy path didn't. Any bare embedder raised `AttributeError`. `_bump` uses `getattr` now.
- **#255** — `tokens >= EMBED_MAX_TOKENS` can't tell a complete 2048-token value from a cut one, so the message asserted an inference. Softened; the refusal stays.

Committed artifacts verified to regenerate byte-identically from the final code. **974 passed, 3 skipped.**

PR #256, opened 2026-08-03
Keep the vectors, guard the growth (#247)

### LINKED ISSUE #247: agreement_cache: committed artifacts are asymmetric and include 1.4 MB of embeddings for a rejected proxy
From the review of #240.

`data/evaluation_llm/agreement_cache/` as committed:

| file | size | note |
|---|---|---|
| `embeddings.jsonl` | **1.4 MB** | raw vectors |
| `VOICE_equivalence.jsonl` | 56 KB | |
| `AI_READI_equivalence.jsonl` | 49 KB | |
| `CM4AI_equivalence.jsonl` | 44 KB | |
| `CHORUS_equivalence.jsonl` | 39 KB | |
| `CHORUS_rows.json` | 20 KB | per-slot detail — **CHORUS only** |
| `matrix.json` | 972 B | |

Two things:

1. **`embeddings.jsonl` is 1.4 MB — 94% of the directory — of nomic-embed vectors for a measure the note concludes does not work** (0.923 vs 0.914, "a gap of 0.009"). Keeping the *result* is right; that finding is exactly the kind that should be recorded rather than re-discovered. Keeping the raw vectors is harder to justify, since nothing will read them again. The per-slot cosines needed to reproduce the 0.923/0.914 table are already in `CHORUS_rows.json`.

2. **`CHORUS_rows.json` exists but there is no `rows.json` for AI_READI, CM4AI or VOICE.** So the per-slot detail — similarity, verdict, and the judge's stated reason — is inspectable for one project out of four. The reasons are the most interesting artifact in the whole run (`title`: "all three name the same CHoRUS dataset, differing only in phrasing") and they are discarded for three quarters of it. This asymmetry is an accident of the exploratory run coming first, not a decision.

Suggested: either write `{PROJECT}_rows.json` for all four, or drop the CHORUS one for consistency and rely on the `_equivalence.jsonl` caches (which do retain slot + verdict + reason for every project). Prefer the former. Separately, consider whether `embeddings.jsonl` should be gitignored.

realmarcin:
Closes #247.

I filed #247 worrying about 1.4 MB, then kept the file for a reason that was also wrong. Both are worth correcting.

## The size concern was off by three orders of magnitude

| | |
|---|---|
| working tree | 1.4 MB |
| **packed in git** | **587 KB** |
| share of this repo's 3.25 GiB of objects | **0.017%** |
| rank among largest blobs in history | not in the top 15 (largest is a 3.3 MB source text) |

And the blob has been in history since `a1b3f5ae`, so **deleting it from HEAD reclaims exactly zero bytes**. Only a history rewrite would, and rewriting a shared `main` over half a megabyte is not a trade worth taking.

## The reason for keeping it was wrong too

I wrote that the file is "what makes the negative result reproducible". It isn't. The 0.923 / 0.914 table recomputes from the 24 KB `CHORUS_v2_rows.json` with no vectors involved:

```
from CHORUS_v2_rows.json alone (24 KB):
  equivalent n=18 mean=0.92272
  different  n=30 mean=0.91442
  gap=0.00831   <- the whole finding
```

I had already run exactly this during the #251 review without noticing it refuted my own justification. The vectors let you re-derive the cosines; the rows already carry them.

## What this PR does instead

**Pins the negative result in the suite.** It is the finding someone will want to re-litigate the first time the judge's cost comes up, and a number in a note is easier to wave away than a red test. The assertion also has a direction: if the gap ever exceeds one point, the proxy is worth another look and the test is what says so.

**Bounds the growth, which is the part that was actually unguarded.** The cache holds 136 vectors — one per distinct CHORUS v2 value. Embedding the whole corpus is 1439, about **14 MB at 10,343 bytes a vector**, tracked permanently, and `--embed --embed-online` would do it in a single command. A test caps the count at 200 so crossing it is a decision made in a commit rather than a side effect discovered later.

**Corrects the record** in the note's known-limits section.

## Unrelated, and much bigger than any of this

`.git/objects` is **3.25 GiB across 42 packs**, with 18 prune-packable objects. That smells like it would repack considerably smaller. I have not touched it — flagging rather than acting, since it affects the whole repository and is not mine to decide.

**976 passed, 3 skipped.**

### DIFF
--- a/notes/replicate_agreement_2026-08-02.md
+++ b/notes/replicate_agreement_2026-08-02.md
@@ -243,9 +243,18 @@ Filed rather than fixed, so they are on the record either way:
 - **#243** — those same records do not carry the judge model. It is recovered
   above from the run configuration, not from the cache. The 30 re-judged records
   do carry it, along with the cap they were judged under.
-- **#247** — `embeddings.jsonl` is 1.4 MB of vectors for a proxy this note
-  concludes does not work. Kept, because it is what makes `--offline --embed`
-  reproduce the 0.923 / 0.914 table rather than print nulls.
+- **#247 — kept, and the reasons first given for both keeping and worrying
+  about it were wrong.** The file is 1.4 MB in the working tree but 587 KB
+  packed, which is 0.017% of this repository's 3.25 GiB of objects and does not
+  appear in its fifteen largest blobs. It is also already in history, so
+  deleting it from HEAD would reclaim exactly nothing — only a history rewrite
+  would, which is not a thing to do to a shared `main` over half a megabyte.
+  And it is *not* what makes the negative result reproducible: the 0.923 /
+  0.914 table recomputes from the 24 KB `CHORUS_v2_rows.json`, with no vectors
+  involved. That table is now asserted by the test suite rather than left to be
+  re-derived. What remains real is growth — embedding the whole corpus would be
+  1439 vectors rather than 136, about 14 MB, tracked permanently — so a test
+  bounds the count and makes that a decision instead of a side effect.
 - **#251 — fixed.** Vectors are keyed on the text sent rather than the value it
   came from, mirroring #244, and the client now reads `usage.prompt_tokens` back
   and refuses to build a cosine on a prefix. Nothing was orphaned: no cached
--- a/tests/test_agreement.py
+++ b/tests/test_agreement.py
@@ -464,7 +464,14 @@ class TestJudgeCache(unittest.TestCase):
             self.assertEqual(judge.legacy_hits, 1)
 
 
-@unittest.skipUnless((CACHE / "matrix.json").exists(), "agreement cache not present")
+#: Every file this class reads. Guarding on one of them and reading three meant
+#: a partial cache errored instead of skipping, so "the cache is not here" and
+#: "the cache is here and wrong" arrived as the same red result (#260).
+CACHE_FILES = ("matrix.json", "CHORUS_v2_rows.json", "embeddings.jsonl")
+
+
+@unittest.skipUnless(all((CACHE / f).exists() for f in CACHE_FILES),
+                     "agreement cache not present")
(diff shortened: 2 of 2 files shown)


realmarcin:
## Review

Two findings, filed as #257–#258.

- **#257 (fixed in fa58913b)** — the test added here to pin the embedding finding split its classes on truthiness, so an unjudged (`None`) row would be counted as judged-and-different. That is #250 exactly, reintroduced inside the test whose job is to keep this result honest. Latent today (the rows file holds only `True`/`False`), live the moment that cell is regenerated from a partial run — and silent when it fires. Now splits on identity.
- **#258 (filed, not fixed)** — only `embeddings.jsonl` is size-guarded; the four verdict caches are unbounded. Defensible asymmetry: ~390 bytes per verdict against ~10,343 per vector, so a full re-judge adds ~210 KB where a full embedding sweep adds ~14 MB. Recorded so the choice doesn't read as an oversight.

Also verified while reviewing: no duplicate keys in any of the five caches (511 lines / 511 distinct), because `_load` populates before `_save` appends.

**976 passed, 3 skipped.**

realmarcin:
## Second review pass

Two more findings, filed as #259–#260, both fixed in 09e254bc.

- **#259** — the assertion pinning "the proxy does not discriminate" tested a *signed* difference where the claim is about separation, which is a magnitude. `assertLess(mean_eq - mean_df, 0.01)` passes on `mean_eq=0.40` against `mean_df=0.92` — a gap of −0.52, the proxy discriminating strongly and merely inverted. That is the most interesting result the test could meet and the one it was blind to. `abs()` now.
- **#260** — the class guard checked `matrix.json` while its tests read three files, so a partial cache errored instead of skipping. Same conflation as #250 and #257, in a third place. Guard now matches what is read.

Running total for this PR: four findings (#257, #258, #259, #260), three fixed, #258 filed with its reasoning. **976 passed, 3 skipped.**

Issue #264, opened 2026-08-03
.gitignore: "Re-extract from raw/*.zip" is now ambiguous — raw/ holds two unrelated zips


realmarcin:
From the second review of #261.

The comment above the crate rule reads:

```
# RO-Crate packages: keep the upstream archive (raw/), not the extracted tree.
# CM4AI's extracted crate is ~47 MB ... Re-extract from raw/*.zip.
data/ro-crate_packages/*/crate/
```

That instruction was written when `raw/` held one zip. It now holds two, with unrelated contents and a 3000× size difference:

| zip | size | extracts to |
|---|---|---|
| `cm4ai_release_metadata.zip` | 1.1 MB | `cm4ai_release_metadata/AP-MS/…` — 36 files of RO-Crate HTML. **This is the crate source.** |
| `cm4ai_images_untreated.zip` | 3.2 GB | `untreated/blue/B2AI_1_untreated_A10_R1_z01_blue.jpg` — 15,911 IF images. **Not a crate.** |

So "Re-extract from `raw/*.zip`" now glob-matches an archive that will not reproduce `crate/` and will instead unpack 4.12 GB of microscopy images into it. The glob was accurate when written and #261 is what made it wrong, by recovering the image archive into that directory.

Fix: name the crate source explicitly in the comment rather than globbing, and say what the other archive is so nobody has to open a 3.2 GB zip to find out.

realmarcin:
Fixed in 1159e7ba (#261). The crate source is named explicitly rather than globbed, and the comment now says what the other archive is (3.2 GB, 15,911 IF JPEGs under `untreated/`, not a crate) so nobody has to open it to find out. This PR is what made the glob wrong in the first place, by recovering the image archive into that directory.

Issue #263, opened 2026-08-03
.gitignore: *.tar.* ignores checksum sidecars, which are exactly what an untracked archive needs


realmarcin:
From the second review of #261.

```
data/ro-crate_packages/*/raw/**/*.tar.*
```

That pattern was meant to catch `.tar.gz`, `.tar.bz2`, `.tar.xz`. It also catches anything else with `.tar.` in the name:

| file | ignored? | wanted? |
|---|---|---|
| `images.tar.gz` | ✅ | yes |
| `images.tar.md5` | ✅ | **no** |
| `manifest.tar.sha256` | ✅ | **no** |
| `notes.tar.README` | ✅ | **no** |

The checksum case is the one that matters, and it is self-defeating. The entire reason these archives are untracked is that they are too large to keep in history — which makes a small tracked checksum the only way anyone can verify that the copy sitting in their `raw/` is the right one. This rule ignores precisely that file.

Fix: enumerate the compression suffixes (`*.tar.gz`, `*.tar.bz2`, `*.tar.xz`, `*.tar.zst`) instead of the open-ended `*.tar.*`.

realmarcin:
Fixed in 1159e7ba (#261). Compression suffixes are enumerated — `.tar.gz`, `.tar.bz2`, `.tar.xz`, `.tar.zst` — instead of `*.tar.*`. Verified `images.tar.md5`, `manifest.tar.sha256`, `notes.tar.README` and `images.zip.sha256` are all tracked again, with all ten archive shapes still ignored.

Issue #258, opened 2026-08-03
agreement_cache: only embeddings.jsonl is size-guarded; the verdict caches grow unbounded


realmarcin:
From the review of #256.

#256 bounds `embeddings.jsonl` at 200 records. The four `*_equivalence.jsonl` caches have no bound at all:

| file | records | size |
|---|---|---|
| VOICE_equivalence.jsonl | 154 | 60 KB |
| AI_READI_equivalence.jsonl | 137 | 53 KB |
| CM4AI_equivalence.jsonl | 121 | 46 KB |
| CHORUS_equivalence.jsonl | 99 | 38 KB |
| **total** | **511** | **197 KB** |

They are append-only and never pruned, and they already carry superseded entries by design — the legacy-keyed verdicts for the 30 slots re-judged in #248 are retained deliberately so the old index still resolves.

**Low severity, and the asymmetry is defensible**: a verdict record is ~390 bytes against ~10,343 for a vector, so a full re-judge of every slot adds ~210 KB where a full embedding sweep adds ~14 MB. That is a 26× difference in cost per record and it is why the guard went where it did.

Filing it so the choice is on the record rather than looking like an oversight, and so that if these caches ever start accumulating across many configurations — each new prompt config adds a fresh set of keys to the same files — somebody notices before it matters.

Worth checking at the same time: no duplicate keys exist in any of the five caches today (verified, 511 lines / 511 distinct keys), because `_load` populates before `_save` appends. Concurrent processes writing the same cache would break that assumption.

PR #265, opened 2026-08-03
Guard what enters history: CI size check (#262) and cache budgets (#258)

### LINKED ISSUE #262: No size-based guard against committing a large file; .gitignore only covers *.zip in raw/
From the review of #261.

#261 stops the specific case that bit us — a multi-gigabyte `.zip` landing directly in `data/ro-crate_packages/*/raw/`. Verified coverage:

| path | ignored? |
|---|---|
| `CM4AI/raw/new-image.zip` | ✅ |
| `CHORUS/raw/new-image.zip`, `VOICE/raw/…` | ✅ |
| `CM4AI/raw/images.tar.gz` | ❌ |
| `CM4AI/raw/images.7z`, `.tar`, `.tgz` | ❌ |
| `CM4AI/raw/images.zip.001` (split archive) | ❌ |
| `CM4AI/raw/images/foo.zip` (one level deeper) | ❌ |
| `CM4AI/big.zip`, `CM4AI/processed/big.zip`, `data/big.zip` | ❌ |

So the guard is by extension, by location, and one directory level deep. The actual failure mode — *any* large file reaching a commit — is unguarded, and the next one will simply have a different name.

`.gitignore` cannot express "no file over N megabytes", so this needs a different mechanism:

- **`check-added-large-files`** via the `pre-commit` framework. Idiomatic and exactly designed for this. But this repo has no `.pre-commit-config.yaml`, no active hooks and no `pre-commit` dev dependency, so it means introducing tooling — and it only protects contributors who install it.
- **A CI check** on the PR diff, rejecting blobs over a threshold. Slower to fire but protects everyone without local setup, which matters more here since the damage (a large object entering shared history) is only permanent once pushed.

The CI route is probably the right one, and it is a repo-wide tooling decision rather than something to slip into a `.gitignore` change — hence filed rather than folded into #261.

Worth noting the asymmetry that makes this worth doing at all: locally a stray large object is recoverable, as #261 showed. Once pushed, it is in history permanently and no `git rm` retrieves it.

realmarcin:
Closes #262 and #258. Both are the same shape — bound what can enter shared history — so they share a branch, one commit each.

## #262 — a CI check on the PR diff

A stray large object is cheap to fix locally and impossible to fix once pushed. The 3.19 GB CM4AI archive in #261 was recoverable **because it never reached GitHub**; committed and pushed, rewriting shared history would have been the only remedy.

`.gitignore` can't express "no file over N megabytes" — it matches names, and #263 and #264 were both cases of the next file not being named like the last. A pre-commit hook could express the rule but protects only contributors who install one, which is the wrong side of the boundary. So: CI, on the pull request.

**The limit is 10 MB, chosen against the repository rather than picked round:**

| | |
|---|---|
| largest file tracked at HEAD | 3.94 MB |
| files over 5 MB | 0 (of 3,669) |
| files 1–5 MB | 20 |

That's ~2.5× the real ceiling — source PDFs and concatenated corpora pass untouched.

Two details that are why this is a script and not one line of bash:

- **It diffs against the merge base, not the base tip.** A long-running branch must be judged on what *it* adds; diffing the moving tip of `main` would fail someone's PR for a large file a different PR merged yesterday. There's a test for exactly that.
- **It reports every offender, not the first.** A PR adding three large files should learn that once, and a check showing a subset reads as "this is all of it".

Deletions are excluded — removing a large file is the fix, not the offence. The limit isn't overridable by environment variable: committing something this big should be a decision visible in a diff, not a workflow setting nobody reads.

Canaried against its own PR (passes at 10 MB, fails at a 3 KB limit with the full offender list). The canary also caught two defects in my own output — every sub-MB file printed as `0.00 MB`, and stdout/stderr interleaving so the summary landed *after* the failure it introduces. Both fixed, both tested.

## #258 — budget every cache, not just the expensive one

#247 bounded `embeddings.jsonl` because a vector costs ~10 KB. The four verdict caches had no bound, which made the guard read as an oversight. A verdict is ~395 bytes — a 26× difference in cost per record, which is a reason for different *numbers*, not for guarding one and ignoring the others.

```
embeddings.jsonl              136 -> 200   full-corpus sweep is 1439, ~14 MB
{PROJECT}_equivalence.jsonl 99-154 -> 400  full re-judge of one project ~300
```

Plus two things per-file counts don't give:

- **A total-bytes bound** (1.78 MB today against 4 MB), because a cache can grow by getting *fatter* rather than longer — a larger embedding dimension, or a judge that starts returning paragraphs, sails past a record count.
- **A test that every `*.jsonl` has a budget entry.** That's the gap #258 was actually about: the verdict caches were unguarded and nothing said so. Verified it fires on a hypothetical new cache file.

**989 passed, 3 skipped** (was 976; +11 for the new script, +2 net for the cache guards).

### DIFF
--- a/.github/workflows/large-files.yaml
+++ b/.github/workflows/large-files.yaml
@@ -1,0 +1,38 @@
+# Stop a large file before it reaches shared history.
+#
+# On the pull request rather than in a pre-commit hook, because the boundary
+# worth defending is what enters the shared repository: a stray large object is
+# recoverable locally, and permanent once pushed. See scripts/check_large_files.py
+# and issue #262.
+---
+name: Check for large files
+
+on: [pull_request]
+
+jobs:
+  large-files:
+
+    runs-on: ubuntu-latest
+
+    steps:
+
+      - name: Check out repository
+        uses: actions/checkout@v3
+        with:
+          # The check diffs against the merge base, which needs real history.
+          # The default shallow clone has none and the check would compare
+          # against nothing.
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.12"
+
+      # No dependencies: the script is stdlib and git only, so this job stays
+      # fast and cannot be blocked by an unrelated install failure.
+      - name: Check added and modified files
+        run: |
+          python scripts/check_large_files.py \
+            --base "origin/${{ github.base_ref }}" \
+            --head "${{ github.event.pull_request.head.sha }}"
--- a/scripts/check_large_files.py
+++ b/scripts/check_large_files.py
@@ -44,16 +44,55 @@ def _git(*args: str) -> str:
                           check=True).stdout
 
 
+class BaseRefMissing(RuntimeError):
+    """The branch being merged into is not in this clone."""
+
+
 def changed_files(base: str, head: str = "HEAD") -> list[str]:
     """Paths added or modified between the merge base and `head`.
 
     Merge base rather than the base tip: a long-running branch should be judged
     on what it adds, not on files someone else committed to main meanwhile.
     Deletions are excluded — removing a large file is the fix, not the problem.
+
+    Renames need `--raw` rather than `--diff-filter=AM`. A true rename adds no
+    bytes to history, so skipping it is right; but git *detects* renames by
+    similarity, so deleting `old.zip` and adding a 60%-similar `new.zip` is
+    reported as R and would slip past unweighed even though the destination is
+    a genuinely new blob (#267). Comparing the two blob hashes tells the cases
+    apart: identical means the object already exists, different means it does
+    not.
     """
-    merge_base = _git("merge-base", base, head).strip()
-    out = _git("diff", "--name-only", "--diff-filter=AM", "-z", merge_base, head)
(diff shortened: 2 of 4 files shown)


realmarcin:
## Review

Two findings, #266–#267, both fixed in 2ed7f3d9.

- **#266** — a missing base ref exited with a raw `CalledProcessError` traceback. It **failed closed** (exit 1), which is the half that matters, but a contributor had to read a Python stack to learn the real problem was a ref nobody fetched. Now names it and says what to do.
- **#267** — `--diff-filter=AM` excludes `R`. For a true rename that is right, since the blob is already in history. But git *detects* renames by similarity, so deleting `old.zip` and adding a 60%-similar `new.zip` is reported as `R` — and the destination is a genuinely new object that was never sized. Now compares source and destination blob hashes to tell the two apart.

Writing the true-rename test exposed that my first version tested nothing: it created the file on the branch and renamed it there, which relative to the merge base is just a new file at its final path. Corrected so the file pre-exists on the base.

**Verified in real CI**: the `large-files` job ran on this PR and passed in 11s, resolving `origin/main` — so the workflow works end to end, not just the script.

**992 passed, 3 skipped.**

Issue #166, opened 2026-07-29
Generic condition is incomplete: de_novo and crate_only arms not run


realmarcin:
PR #164 adds the generic condition for the **baseline** arm (4 GCs) and the **healthsheet** arm (AI-READI). Two arms are missing:

- `de_novo` (documents + crate) — CHORUS, CM4AI, VOICE
- `crate_only` — CHORUS, CM4AI, VOICE

18 runs at 3 replicates each. Launch was deliberately halted pending the design rework now merged, so these should use the finalised `generic` / `tuned` conditions rather than the retired `-deprimed` prompts.

### Why it matters

The headline claim — only CHORUS retains a reproducible crate contribution — currently rests on the tuned and de-primed conditions. Without the generic crate arms there is no way to check whether that survives under the controlled prompt, and de-priming already moved VOICE from *marginal* to *not resolvable*.

It also leaves one correction unverified. PR #164 notes that CM4AI's referent pin looked inert on agreement (78.1% vs 77.8%) but is in fact load-bearing (`file_collections` 10 → 0). That was established on the baseline arm only; the crate arms would confirm or refute it.

### Cost note

Via the API path (`d4d api plan`), VOICE de_novo is the most expensive bundle in the study at ~175k tokens per phase.

realmarcin:
Complete. Both arms ran on 2026-07-31 under `2026-07-31_claude-opus-5-api-generic_rep{1,2,3}`:

| arm | method | replicates | projects |
|---|---|---|---|
| `de_novo` | `claudecode_agent_crate` | 3 | CHORUS, CM4AI, VOICE |
| `crate_only` | `claudecode_agent_crate_only` | 3 | CHORUS, CM4AI, VOICE |

That is exactly what this issue asked for — both arms, those three projects, three replicates each. Verified with `d4d runs list --arm de_novo --full-only` and `--arm crate_only`.

Recorded in NEXT_TASKS.md §5 (#268).

PR #268, opened 2026-08-04
Reconcile NEXT_TASKS.md against what actually shipped

realmarcin:
Routine backlog reconciliation. Last verified 2026-07-29; fifteen PRs have merged since. Every pending claim checked against the corpus and git log rather than re-read.

## Two sections were describing work that has shipped

**§6 "Run the generic-v2 arm — staged, not run"** — it ran on 2026-07-31 (under that label, not the staged `2026-07-30`), and `notes/generic_v2_results.md` has the outcome.

Recorded along with the part that matters more than the headline:

| axis | v1 → v2 | |
|---|---|---|
| substance | 40 → 15 | −62% |
| target | 41 → 16 | −61% |
| **form** | **50 → 56** | **worse** |

Rule 1 eliminated collapsed cardinality **27 → 0** and created hollow objects **2 → 33** — one object per entity exactly as instructed, everything crammed into free-text `description` while `name`, `id`, `affiliations` and the date fields go unused. So the promotion decision §6 defers to "only if v2 validates" is live, and it is not a yes/no.

**§5 "de_novo and crate_only are unrun"** — `crate_only` has three replicates under `2026-07-31_claude-opus-5-api-generic`. `de_novo` has no method directory at all, so that half is a first run rather than a top-up.

## One residual, now located

§4 said "the LLM evaluation paths still divide by 84". Still true:

- `constants/evaluation.py: RUBRIC20_MAX_SCORE = 84` — exported and imported by nothing. Dead *and* wrong, which is the combination the next person picks up.
- `scripts/summarize_rubric20_results.py` hardcodes `/84` in its markdown table regardless of `max_points`.

## Two threads that were not logged at all

**§7 — replicate agreement.** #169 settled on evidence (effect −2.9 against sd 10.9, deltas disagreeing in sign, ~110 projects needed). Two instrument defects found and fixed, both the same shape — evidence silently withheld: the judge saw 4000 characters, and the embedding endpoint truncates at 2048 tokens returning HTTP 200.

**§8 — the 3.19 GB orphan blob**, with a warning that is not a task: the recovered archive is the only local copy, it is gitignored so nothing backs it up, it is one of three image conditions, and no download URL is recorded on disk.

Docs-only; no code paths touched.

### DIFF
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -17,7 +17,14 @@ deleted — `d4d runs restore --execute` reverses it.
 
 **Corpus: partial 23 → 0.**
 
-| level | count |
+The level table below is **as of 2026-07-29** and has not been re-derived; the
+corpus has since grown to 123 runs, so treat the counts as a record of that day
+rather than of today. What still holds, re-checked 2026-08-03:
+`d4d runs archive --unattested` reports *"No unattestable runs found — every run
+in the corpus can be placed."* The property the section is about survived the
+growth even though its arithmetic is stale (#270).
+
+| level (2026-07-29) | count |
 |---|---|
 | live | 49 |
 | attested | 33 |
@@ -55,7 +62,13 @@ enforce a rule that postdates them.
 - A label with no parseable date is **subject** to the rule. Exempting anything
   unparseable is an exemption taken by accident.
 
-Corpus today: 98 runs checked, 0 subject to the requirement, 0 failing.
+Corpus on 2026-08-03: **131 runs checked, 49 subject to the requirement, 0
+failing** (`d4d runs check --strict`, exit 0).
+
+The "0 subject" recorded here on 2026-07-29 was true only because no run had yet
+crossed the 2026-07-30 cutoff — it described a gate that had never fired. Every
+run made since is subject to it and every one satisfies it, which is the
+evidence this section was written to be able to claim.
 
 ---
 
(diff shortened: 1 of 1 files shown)


realmarcin:
## Review

Two findings, #269–#270, both fixed in 294dff32. The first is this PR's own reconciliation reaching a wrong conclusion.

- **#269 — §5 was wrong.** I wrote that `crate_only` had run and `de_novo` had not. **Both have run**, and #166 is complete (now closed). I searched for directories whose *names* contain `de_novo`, got zero results, and believed them — but `de_novo` is an arm classification, not a path component, and its method directory is `claudecode_agent_crate`. `d4d runs list --arm de_novo` shows three replicates covering exactly the projects #166 asked for. A grep returning nothing is the easiest kind of evidence to over-trust, and the tool that knew the mapping was never asked.
- **#270 — §2 was stale.** "98 runs checked, 0 subject, 0 failing" is now **131 checked, 49 subject, 0 failing**. The rule applies from 2026-07-30 and every 2026-07-31 run postdates it, so "0 subject" described a gate that had never fired. "49 subject, 0 failing" is the evidence the section was written to be able to claim.

Both errors sat in sections I had either just rewritten or skipped as inert because they were marked *done*. A done conclusion can rest on a count that keeps moving.

§1's level table is stamped with its date rather than re-derived — I found no single command reproducing the live/attested/derived/none breakdown, and guessing would be worse than an honest date stamp. The property it is about was re-checked: `d4d runs archive --unattested` still reports every run can be placed.

Docs-only; no code paths touched.

Issue #243, opened 2026-08-03
agreement.py: judge cache is not keyed on the judge model


realmarcin:
From the review of #240.

`Embedder` filters its cache on the model that produced each vector:

```python
if e.get("model") == self.model:
    self._vecs[e["key"]] = e["vector"]
```

`EquivalenceJudge` does not. It keys only on the rubric digest:

```python
if e.get("rubric") == _digest(EQUIVALENCE_SYSTEM):
    self._memo[e["key"]] = (e["equivalent"], e.get("reason", ""))
```

and `_save` writes `{"rubric", "key", "slot", "equivalent", "reason"}` — no model field. Confirmed across the committed cache: all 481 records carry a single rubric digest and no model.

Consequences:

1. Switching judge model (or effort tier — this repo runs several) silently reuses verdicts from the previous one. There is no cache miss and no warning.
2. The judge model for the published 434 verdicts is **not recorded anywhere** — not in the cache, not in `matrix.json`, not in the note. `_resolve()` falls back to `_model_settings()["name"]`, whatever that was on 2026-08-02.

Fix: add `"model"` to the saved record and to the `_load` filter, mirroring `Embedder`. Separately, back-fill the judge model into the note if it can be recovered from the run logs.

Related: the judge is drawn from the same model family that generated the records it grades. That may be defensible, but it should be stated rather than left implicit.

realmarcin:
Fixed in 2d91e356 (#240). `_save` now writes `model` and `_load` filters on it, mirroring `Embedder`. Records written before the scoping carry no model and are read from a separate legacy index — they are frozen, so keeping them is safe and re-running 434 paid calls to restate identical verdicts would buy nothing.

The judge model for the published verdicts is recovered: `google/claude-opus-5-high`, from `_model_settings()`. Now in the note's provenance table, along with the point that it is the same model family that generated the records it grades.

PR #271, opened 2026-08-04
Split the form class into collapsed-cardinality and hollow-object

realmarcin:
Implements the recommendation `notes/generic_v2_results.md` closed with: *"Any future comparison on this axis should split them."*

## The problem

generic-v2's rule 1 was written to stop the model collapsing several entities into one object. On the axis meant to show it, form went **50 → 56** — apparently worse. The results note read the judge's reasons by hand and concluded the rule had worked while a different defect took the same label. This makes that a measurement.

## Method: don't re-judge what's already paid for

The obvious implementation is to add sub-types to `FITNESS_SYSTEM`. That invalidates **all 1441** cached fitness judgements — the cache is keyed on `(axis, model, rubric, corpus, schema)` precisely so a changed rubric cannot be silently compared against an unchanged one — and discards a v1/v2 comparison already bought.

But the sub-type question doesn't need the fitness judge re-run. Every cached entry already carries the slot and the full value. So a narrower classifier runs over the **106 cached `form` failures only**:

- **106 calls instead of 1441**, 13× less
- published fitness numbers untouched — this is a layer, not a replacement
- each failure attributed to its arm by exact canonical-JSON match against the records (the fitness cache keeps no run label). It lands on **50 v1/ 56 v2 with nothing unattributed** — the published split arriving independently, and a test fails if that stops being exact

## Result

| subtype | v1 | v2 |
|---|---|---|
| collapsed cardinality | 34 | **2** |
| hollow object | **0** | **45** |
| both | 8 | 5 |
| other | 8 | 4 |
| **total** | **50** | **56** |

Counting each defect wherever it appears, `both` included:

| defect | v1 → v2 | |
|---|---|---|
| **collapsed cardinality** | **42 → 7** | **−83%** |
| **hollow object** | **8 → 50** | **+42** |

**The rule did what it was written to do.** What the old axis reported as a 12% regression is one defect nearly eliminated and a second appearing in its place — the model now emits one object per entity, exactly as instructed, and fills each with free-text `description` while `name`, `id`, `affiliations` and the date fields go unused. Correct shape, empty.

## Against the manual read

| subtype | this run | manual |
|---|---|---|
| collapsed | 34 → 2 | 27 → 0 |
| hollow | 0 → 45 | 2 → 33 |
| other | 8 → 4 | 17 → 20 |

Same direction and magnitude, different residual — 12 in `other` against 37. That's the expected shape when a human reading judge reasons in bulk is compared against a judge asked one question about one value with the schema in front of it. The instrument is the one that can be re-run.

## What it changes

- **`form` should not be reported as one number again.** Two defects, opposite responses to the same intervention.
- **For the v2 promotion decision:** rule 1 succeeds on its own terms and should **not** be promoted as written — it needs a companion instruction to populate the declared structured fields rather than `description`, or it trades one form defect for another and leaves the total flat.

## Verification

Canary first (one call, verified it persisted with model and cap recorded) before the other 105. `python -m data_sheets_schema.form_defects --offline` reproduces the whole table from cache with no paid call.

An unparseable answer raises rather than defaulting to `other` — `other` is a real finding, and letting it absorb the classifier's own errors is how one `form` class came to hide two defects to begin with.

**1008 passed, 3 skipped** (+16). Subtype cache is 44 KB.


realmarcin:
## Review

One finding, #277, fixed in 34a49398.

`load_form_failures` selected on the failure class alone, while every fitness entry also carries `rubric` and `model` — the fields `JudgementContext` exists to make a judgement self-describing. Today the cache holds one of each so nothing was mixed, but this module's premise is that `FITNESS_SYSTEM` will not be edited *because* editing it invalidates the cache. If it ever is, the loader would pool both instruments and the sub-type table would look identical. Same defect as #243, and the sub-type classifier's own cache in this PR already filters on both — the loader was the one place that did not.

It raises rather than filtering: a silent skip could halve the corpus and still produce a plausible table.

**1010 passed, 3 skipped.**

PR #272, opened 2026-08-04
Draft the companion rule for v2's rule 1 (generic-v3) — staged, not run

realmarcin:
Follow-on from #271. The form-defect split showed v2's rule 1 doing exactly what it says and handing the failure to a neighbour:

| defect | v1 → v2 |
|---|---|
| collapsed cardinality | 42 → 7 |
| hollow object | 8 → 50 |

The rule is a success on its own terms and is not finished. This drafts the companion. **Nothing is run.**

## The rule

> When a slot's declared range is a class, populate the fields that class declares. Placing the content in a free-text field such as `description` while the declared fields — a name, an identifier, dates, affiliations — stay empty produces an object of the correct shape holding none of the structure it exists to carry. Where the evidence answers a declared field, populate that field rather than restating it in prose.

Stated on the same terms as v2's three: how the schema is used, no fact about any dataset, so it can be said without naming a project.

## v3 is a third file, not an edit

Same reason v2 was one — v2 produced the 2026-07-31 series and is the baseline v3 is measured against. v3's body is **generated from v2's**, so everything outside the marked block is provably identical, and a test asserts it. A second test asserts v2's three rules survive intact, since dropping them would make a v2-vs-v3 difference measure their removal as well as the addition.

## Comparability

`CONDITION_AXES` gains base `v3`, so:

| pairing | differs on | interpretable? |
|---|---|---|
| `generic_v2` vs `generic_v3` | base | ✅ isolates the companion rule |
| `generic_v3` vs `tuned` | base **and** tuned | ❌ flagged confounded |

Same trap v2 had — `tuned` is pinned to the v1 base. A test asserts all three generic bases are distinct, because two conditions sharing one would be silently indistinguishable.

## Genericness

The block carries no digits, no project, no dataset identifier, no expectation phrasing, and no reference to prior runs. That last one is worth naming: the defect it corrects *was* found in a prior run, and saying so would turn a decision rule into a quality warning. The test also bans the word **"hollow"** from the block, so the prompt cannot name the failure class it is about to be measured on.

## The prediction is registered, not in the prompt

`notes/generic_v3_analysis_plan.md`, written before any run, including what would count as failure:

- hollow objects flat or up — the rule is not doing what it says
- collapsed cardinality returning — the rules are not additive, and rule 1 plus the companion needs rewriting as one instruction
- a new sub-type appearing in `other` — the same exchange one level down, now **visible** because `other` is measured rather than inferred (8 → 4 across v1 → v2)

Neither the fitness rubric nor the sub-type classifier's rubric is edited, so the v1/v2 judgements and cached sub-type labels stay valid and all three arms remain comparable.

## Cost, if you decide to run it

Twelve generations (4 projects × 3 replicates) plus fitness scoring and sub-type classification of whatever form failures result.

**1005 passed, 3 skipped** (+13).

### DIFF
--- a/notes/generic_v3_analysis_plan.md
+++ b/notes/generic_v3_analysis_plan.md
@@ -1,0 +1,77 @@
+# generic-v3 analysis plan
+
+Registered before any v3 generation run. The prediction is here rather than in
+the prompt because writing it there would instruct the model to produce the
+result the run is meant to test.
+
+## What v3 changes
+
+One rule, added to v2's three:
+
+> When a slot's declared range is a class, populate the fields that class
+> declares. Placing the content in a free-text field such as `description` while
+> the declared fields — a name, an identifier, dates, affiliations — stay empty
+> produces an object of the correct shape holding none of the structure it
+> exists to carry. Where the evidence answers a declared field, populate that
+> field rather than restating it in prose.
+
+## Why
+
+v2's rule 1 worked and traded one defect for another. Splitting the fitness
+`form` class (`notes/form_defect_split_2026-08-03.md`) measured, over the 106
+form failures already judged:
+
+| defect | v1 → v2 |
+|---|---|
+| collapsed cardinality | 42 → 7 |
+| hollow object | 8 → 50 |
+
+The reported `form` total moved 50 → 56 and hid both.
+
+## The comparison
+
+**v2 against v3**, not v1 against v3. The two differ on one axis (`base`), and
+only that pairing isolates the companion rule. v1 against v3 would measure four
+rules at once.
+
+Same four projects, three replicates, same model and settings as the v1 and v2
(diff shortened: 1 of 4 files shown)


realmarcin:
## Review

No findings. Probed the wiring specifically:

- `resolve_prompt(condition="generic_v3")` resolves and carries **all four rules** — v2's three plus the companion. (My first probe said otherwise; it was a false alarm from a line break inside v2's rule text.)
- `_CONDITIONS` derives from `CONDITION_PROMPTS`, so `--condition generic_v3` is available on every `d4d api` subcommand with no further wiring.
- `generic_v2` vs `generic_v3` differs on one axis; `generic_v3` vs `tuned` on two and is flagged confounded.

**1005 passed, 3 skipped.**

PR #273, opened 2026-08-04
Rubric20 scores out of 88 everywhere, including where the judge is told

realmarcin:
Closes the §4 residual in NEXT_TASKS.

## My earlier diagnosis was incomplete

I reported two places carrying 84. There were **seven** — and the two I found were the two that didn't matter.

| location | effect |
|---|---|
| `constants/evaluation.py` | exported, imported by nothing — dead |
| `summarize_rubric20_results.py` | 3 fallbacks + a hardcoded table denominator |
| **`rubric20_system_prompt.md`** | **the judge's own prompt — the one that actually bit** |
| `batch_evaluate_rubric20_hybrid.py` | `max_total = 84` |
| `render_evaluation_html_rubric20_semantic.py` | fallback |
| `generate_gc_approach_comparison.py` | 3 divisions + 3 display sites |
| `render_interleaved_evaluation.py` | fallback |

## Why fixing the fallbacks alone would have changed nothing

The system prompt handed the model a `"max_points": 84` template to copy, and told it to compute `percentage` as `total_points / max_points`. So **every LLM-path result carried `max_points: 84` in its data**, every downstream `overall.get('max_points', 84)` agreed with it, and the fallbacks never fired. The data itself was wrong.

Its Scoring Summary also decomposed 84 as *16 numeric + 4 pass/fail*. The rubric defines **17 numeric + 3 pass/fail = 88** — one question counted as pass/fail that the rubric defines as numeric. The presence path used 88, the LLM path used 84, and the two were never comparable.

## Per-category maxima removed, not corrected

They summed to 84, and the rubric file **declares no categories at all** — the four-way grouping is the prompt's own, so nothing can derive a per-category maximum from it. Writing four numbers I can't check is how the wrong total survived being read this long.

## The constant is now derived

```python
RUBRIC20_MAX_SCORE = 17 * 5 + 3   # 88
```

Spelled out as its parts, the way `RUBRIC10_MAX_SCORE` already is. A bare `84` gave nothing to check it against.

Tests tie the constant *and* the question count to the rubric file, and sweep every rubric20/evaluation script plus the judge prompt for a hardcoded denominator — matching the shapes a denominator takes rather than the digits, since unrelated 84s exist (an 84-question healthsheet, a percentage).

## Consequence worth stating plainly

**LLM-path rubric20 percentages recorded before this were computed against 84 and are ~4.8% high.** They are not comparable with presence-path scores, and re-running them is the only way to make them so.

**995 passed, 3 skipped.**

### DIFF
--- a/scripts/batch_evaluate_rubric20_hybrid.py
+++ b/scripts/batch_evaluate_rubric20_hybrid.py
@@ -20,6 +20,7 @@ from datetime import datetime
 from typing import Dict, List, Any, Optional, Tuple
 import hashlib
 import re
+from data_sheets_schema.constants import RUBRIC20_MAX_SCORE
 
 # Base directory
 BASE_DIR = Path(__file__).parent.parent
@@ -703,7 +704,7 @@ def evaluate_d4d_file(file_path: Path, project: str, method: str, eval_type: str
     }
 
     total_score = 0
-    max_total = 84  # 16 numeric questions × 5 + 4 pass/fail × 1
+    max_total = RUBRIC20_MAX_SCORE  # 17 numeric x 5 + 3 pass/fail x 1
 
     # Group by category
     categories = {}
@@ -815,7 +816,7 @@ def main():
             json.dump(result, f, indent=2)
 
         score = result["overall_score"]
-        print(f"   ✅ Score: {score['total_points']}/84 ({score['percentage']}%)")
+        print(f"   ✅ Score: {score['total_points']}/{RUBRIC20_MAX_SCORE} ({score['percentage']}%)")
 
         concat_count += 1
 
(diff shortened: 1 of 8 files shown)


realmarcin:
## Review

Three findings, #274–#276. Two fixed in c9a8528f; #275 filed.

- **#274 — a defect this PR introduced.** The individual-summary table went from a hardcoded `/84` to a hardcoded `/88`. The records carry their own denominator and the concatenated table already reads it. That matters now: **167 committed rubric20 evaluations carry `max_points: 84`**, so before this PR the table was accidentally right and after it was confidently wrong. It now reads the set's denominator and prints `MIXED(84/88)` when a set spans both.
- **#276** — the JSON template the model copies still carried `"category_max": 24`, Structural Completeness's maximum under the 16-numeric-plus-4-pass/fail decomposition that produced 84. Removing the prose while leaving the example is the weaker half of the fix. Now a placeholder.
- **#275 — filed, not fixed.** The 167 records scored against 84 cannot be pooled with new ones. They are self-describing, which is what separates them from the bug this PR fixes. Rescaling is arithmetic rather than re-measurement; re-running is the only option that measures under the corrected prompt. Making the incomparability visible is the cheap half and is done here.

**997 passed, 3 skipped.**

Issue #278, opened 2026-08-04
summarize_rubric20_results: create_detailed_report raises NameError on 'denom' (shipped in #273)


realmarcin:
`create_detailed_report` references `denom` at two f-strings and never defines it:

```
261:  report += f"- Average score: {avg_score:.1f}/{denom} ({avg_pct:.1f}%)\n\n"
273:  report += f"- Average score: {avg_score:.1f}/{denom} ({avg_pct:.1f}%)\n\n"
```

`denom` is a local of `create_markdown_table`. Calling `create_detailed_report` with any results raises `NameError`.

**I introduced this in #273**, and it is merged. A blanket replace of `{avg_score:.1f}/{RUBRIC20_MAX_SCORE}` → `{avg_score:.1f}/{denom}` hit the method-comparison section as well as the table the fix was aimed at.

## Why nothing caught it

The test I added for #274 asserts on the **source text**:

```python
self.assertIn("MIXED(", script)
self.assertNotIn("{avg_score:.1f}/{RUBRIC20_MAX_SCORE}", script)
```

Both still pass. A grep over source cannot see an unbound name, and no test in the suite ever *calls* `create_detailed_report` or `create_markdown_table` — the module has no test of its own at all, which is why a 1028-test suite and three green CI runs said nothing.

That is the actual defect here. The denominator bug was worth fixing; fixing it with assertions that read the file rather than run it produced a script that is now broken in a way the old one was not.

Fix: define the denominator in that scope from the records being averaged, the same way `create_markdown_table` does — and add tests that execute these functions against fixture results rather than inspecting their source.

Issue #274, opened 2026-08-04
summarize_rubric20: the individual-summary table prints a constant denominator, ignoring each record's own max_points


realmarcin:
From the review of #273. **This is a defect #273 introduced**, and it is the same shape as the one it was fixing.

The table row was `{avg_score:.1f}/84`. #273 changed it to `{avg_score:.1f}/{RUBRIC20_MAX_SCORE}` — from one hardcoded denominator to another.

The records have their own. Every evaluation JSON carries `overall_score.max_points`, and the concatenated table three functions up already reads it (`max_pts = overall.get('max_points', RUBRIC20_MAX_SCORE)`). The individual-summary table does not, so it labels every row `/88` regardless of what each record was actually scored against.

That matters right now, not hypothetically: **167 committed rubric20 evaluations carry `max_points: 84`**.

```
evaluation JSONs with overall_score.max_points: 326
  values: {84: 167, 50: 159}     # 84 = rubric20, 50 = rubric10
```

So running the summarizer today prints `/88` over averages computed from records scored out of 84. Before #273 it printed `/84` over the same records and was accidentally right; after #273 it is confidently wrong.

The constant is the correct **fallback** for a record that omits `max_points`. It is not a substitute for reading one that has it.

Fix: average the records' own `max_points` alongside the scores, or refuse to aggregate records whose denominators differ — mixing 84-scored and 88-scored records into one average is not meaningful whatever denominator is printed beside it.

realmarcin:
Fixed in c9a8528f (#273). The table reads the set's own denominator and prints `MIXED(84/88)` when a set spans both — averaging across two denominators is not meaningful whatever number is printed beside it. A test asserts the constant is no longer interpolated there.

PR #279, opened 2026-08-04
Refuse to average scores measured against different maxima (#275)

### LINKED ISSUE #275: 167 recorded rubric20 evaluations were scored against 84 and cannot be pooled with new ones
From the review of #273.

#273 corrects the denominator to 88 everywhere it is computed, including the judge's prompt. It does not — and cannot — change results already recorded:

```
evaluation JSONs with overall_score.max_points: 326
  values: {84: 167, 50: 159}     # 84 = rubric20, 50 = rubric10
```

**167 rubric20 evaluations carry `max_points: 84`.** Their `percentage` fields were computed as `total_points / 84`, so each is ~4.8% higher than the same record scored out of 88.

The saving grace is that they are **self-describing**: each file records the denominator it used, so a reader or a script can tell. They are explicitly 84, not silently wrong. That is the difference between this and the bug #273 fixed, where the judge's prompt and every fallback agreed on 84 and nothing recorded a dissent.

What is still unsafe is **pooling**. Any aggregate over a mix of 84-scored and 88-scored records — an average score, an average percentage, a ranking across methods — is meaningless regardless of the denominator printed beside it. #273 leaves that possible.

Options, in order of cost:

1. **Refuse to pool.** Have the summarizers group by `max_points` and fail, or warn loudly, when a set spans more than one. Cheap, and makes the incomparability visible where it matters.
2. **Rescale.** `percentage` can be recomputed as `total_points / 88` without re-running the judge, since `total_points` is unaffected by the denominator. This is arithmetic, not re-measurement — but it silently rewrites recorded results, and the category scores summing to `total_points` were themselves produced under a prompt that declared 84.
3. **Re-run.** The only option that produces results actually measured under the corrected prompt. Paid.

Related: the judge's prompt also told the model each category's maximum (e.g. `"category_max": 24`), and those numbers were derived from the 84 decomposition — see the companion issue.

realmarcin:
Closes #275 and #278.

## The problem

167 committed rubric20 evaluations record `max_points: 84`; everything after the correction records 88. Those records aren't *wrong* — each states the denominator it used, which is exactly what separates them from the bug that produced them, where every layer agreed on 84 and nothing recorded a dissent.

What's wrong is **pooling** them. A mean over both lands somewhere plausible and reads as a result.

## The fix: refusing is the default

`src/data_sheets_schema/rubric_pooling.py`. `common_denominator()` **raises** rather than picking the max, the mode, or a fallback — any of which would let an aggregate over two instruments look like an aggregate over one. Report generators, which should still produce a report, use `group_by_denominator()` and report per group under a warning naming both populations.

The executive summary was the worst offender — one Average Score over every record regardless of maximum. So were the method and project comparisons: averaging a method across two maxima ranks it against itself measured two different ways.

## It also fixes a bug I shipped in #273

`create_detailed_report` referenced `denom`, a local of a *different* function. **Calling it raised `NameError`** — merged, with 1028 tests and three green CI runs behind it.

Nothing caught it because the test I wrote for #274 asserted on the script's **source text**:

```python
self.assertIn("MIXED(", script)
```

A grep cannot see an unbound name, and no test in the suite had ever *called* these functions.

## So the new tests run them

**Mutation-checked**: reintroducing the exact shipped bug turns 5 of them red; restoring the fix turns them green.

Writing them surfaced a second problem — the first version called the generators without redirecting output and **overwrote two committed artifacts with fixture data**, a test damaging the corpus it exists to protect. They now patch `EVAL_DIR` to a temp directory, and the artifacts were restored with `git checkout`.

The source-text assertion in `test_rubric20_scoring` is replaced by a behavioural one; its coverage moved to `test_rubric_pooling`, which calls the code.

**1043 passed, 3 skipped** (+15).

### DIFF
--- a/scripts/generate_gc_approach_comparison.py
+++ b/scripts/generate_gc_approach_comparison.py
@@ -9,6 +9,23 @@ import sys
 from pathlib import Path
 from typing import Dict, List, Tuple
 from data_sheets_schema.constants import RUBRIC20_MAX_SCORE
+from data_sheets_schema.rubric_pooling import MixedDenominators, common_denominator
+
+
+def _r20_cells(records, avg_points):
+    """Score and percentage cells for a set of rubric20 records.
+
+    Refuses rather than dividing by a constant: this table compares Grand
+    Challenge approaches, and if one approach's evaluations predate the
+    denominator correction and another's do not, a pooled average compares an
+    approach against a measurement change (#281). `common_denominator` raises
+    on a mixed set; the cells say so instead of printing a number.
+    """
+    try:
+        denom = common_denominator(records)
+    except MixedDenominators as mixed:
+        return (f"{avg_points:.1f}/MIXED{tuple(mixed.seen)}", "n/a (mixed maxima)")
+    return f"{avg_points:.1f}/{denom}", f"{(avg_points / denom) * 100:.1f}%"
 
 # Project root
 BASE_DIR = Path(__file__).parent.parent
@@ -53,6 +70,7 @@ def generate_comparison_table():
     for gc in GRAND_CHALLENGES:
         gc_r10_total = 0
         gc_r20_total = 0
+        gc_r20_records = []
         gc_count = 0
         
         for method in METHODS:
@@ -65,7 +83,7 @@ def generate_comparison_table():
                 row = [
                     f"{gc} - {method}",
                     f"{r10['total_points']}/50",
-                    f"{r20['total_points']}/{RUBRIC20_MAX_SCORE}",
+                    f"{r20['total_points']}/{r20.get('max_points', RUBRIC20_MAX_SCORE)}",
                     f"{r10['percentage']}%",
                     f"{r20['percentage']}%"
                 ]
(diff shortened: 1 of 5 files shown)


realmarcin:
## Review

Two findings, #280–#281, both fixed in 5f4279da. Both are pooling paths this PR left alone — leaving them would have undercut what it claims.

- **#280** — "Top Performing D4Ds" applied a ≥80% cut and a sort across the whole corpus. **71/84 is 84.5% and 71/88 is 80.7%**: identical raw scores separating purely by which instrument measured them, with the 167 older records winning systematically in a table whose entire purpose is ranking. Each row already printed its own denominator so a careful reader could see it; the threshold and the ordering could not be seen that way. Now one table per denominator.
- **#281** — `generate_gc_approach_comparison` summed `total_points` and divided by the constant 88 in all three aggregation blocks, never reading `max_points`. That script compares Grand Challenge approaches, so if one approach's evaluations predate the correction and another's do not, the comparison is between an approach and a measurement change. Now prints `n/a (mixed maxima)` rather than a number.

Both tested by **running** the paths, per the standing lesson from #278.

**1046 passed, 3 skipped.**

PR #282, opened 2026-08-04
Correct the number `d4d runs select` argues from (#176)

realmarcin:
`d4d runs select` justified itself with *"replicates disagree on 77-98% of the slots they share"* — byte equality over values that are mostly nested objects of free text, which is the measure #169 showed returns the same answer regardless of input.

## The two measures, on the same generic-v2 runs

| project | byte equality | judged equivalence |
|---|---|---|
| AI_READI | 83% | **47%** |
| CHORUS | 98% | **62%** |
| CM4AI | 77% | **49%** |
| VOICE | 88% | **55%** |

The 77–98% was real — it is exactly what byte equality gives on these runs. It just doesn't mean what the sentence claimed, because two records describing one collection method in different words are not two facts. The honest figure is **47–62%**.

Both help texts now say so, and a test derives the range from `matrix.json` rather than restating it, so they can't drift apart again.

## The decision is unchanged

It never rested on the magnitude. Splicing mixes referents whether half the shared slots differ in fact or nearly all of them, and the coverage merging buys is 1–5 slots either way.

What the old number did was make the case look **better supported than the evidence for it**. That matters most given where this reasoning is headed — a rationale resting on a disproved figure does not survive review, even when the conclusion happens to be right.

## A gap the correction exposed

`select` and `merge` recommend opposite things and **neither named the other**, so a reader believed whichever help text they opened. `merge` now carries the coherence objection and points to `select` as the recommended route to a shippable record.

Both cases are true — one is about coverage, one about coherence. Stating only one of them per command was the actual problem.

Tests assert each command names the other, and that neither offers 77–98% as a disagreement rate again.

**1049 passed, 3 skipped** (+3).

### DIFF
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -412,7 +412,7 @@ def merge_cmd(method, project, labels, config, out_label, unguarded, execute):
 
     **`d4d runs select` is the recommended way to get a shippable record, and
     it argues against this command.** Replicates state different facts on
-    47-62% of the slots they share, so a union splices across referents — one
+    47-63% of the slots they share, so a union splices across referents — one
     replicate's participant count beside another's DOI — to gain 1-5 slots.
     That objection is about coherence and this command's case is about
     coverage; both are true, and the two commands existing without naming each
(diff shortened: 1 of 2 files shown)


realmarcin:
## Review

One finding, #283, fixed in 1b94c25e.

The test derived the range with `round()`, and CHORUS lands on **exactly 62.5%**. Python rounds half to even, so `round(62.5)` is 62 — and the help text read `47-62%`, a range **excluding one of the four values it claimed to span**. A figure whose whole point is that the previous one was wrong should not have a rounding rule doing load-bearing work.

Now floors the min and ceils the max: 47.1, 48.6, 55.1, 62.5 → **47-63%**, which provably bounds every project. The assertion message names the per-project values and says which file to edit.

**1049 passed, 3 skipped.**

PR #284, opened 2026-08-04
Pin the semantic agents' grounding, and audit what the records were scored under (#162)

### LINKED ISSUE #162: Both -semantic agents: 'program of origin' scoring is ungrounded (no program/project field in D4D), breaking determinism
## Context

Found by adversarial review of PR #154 (`rubric-review`). Not covered by the existing follow-ups (#155–#160).

## Problem

PR #154 adds to **both** semantic agents the check: *"Is the description semantically appropriate for the claimed dataset type **and program of origin**?"* (`.claude/agents/d4d-rubric20-semantic.md:58-61`; same edit in `d4d-rubric10-semantic.md`).

But there is **no `program`/`project` field in the D4D datasheet** to ground this on. `D4D_Core.yaml` defines a 95-field exchange schema with no program/project attribute (grep for `program|project` in `src/data_sheets_schema/schema/D4D_Core.yaml` returns nothing). The only `project` concept lives in the **batch evaluation summary** metadata (`D4D_Evaluation_Summary.yaml`), not in the datasheet under evaluation.

## Impact

With no field to read, the evaluator will infer "program of origin" from filenames, keywords, publisher, or prior knowledge — so the **same datasheet can score differently depending on invocation context** (filename, who's asking). This undermines the agents' stated temp=0.0 / "same file → same score" reproducibility guarantee.

## Recommendation

Either (a) name the exact datasheet fields that establish program of origin and require quoted evidence (e.g. inferred *only* from `keywords` / `publisher` / `funders`), or (b) drop "program of origin" from semantic scoring and keep `project` strictly as batch metadata.

## Evidence
- `.claude/agents/d4d-rubric20-semantic.md:58-61`
- `.claude/agents/d4d-rubric10-semantic.md` (same "program of origin" edit)
- `src/data_sheets_schema/schema/D4D_Core.yaml` (no program/project slot)

realmarcin:
Closes #162's code half and establishes what the data half costs.

## The prompt fix was already in

Both agents were corrected on **2026-07-22** (`0e19e85f`): program inference is restricted to quoted values in `keywords`, `publisher` or `funders`, and the filename, invocation context and prior knowledge are forbidden by name — exactly recommendation (a) from the issue.

Nothing pinned it, and it is one careless edit from returning. Five tests now do, including the premise: if the schema ever gains a `program` slot the test fails and the workaround can go.

## The audit is the bigger half

Since semantic scores are going in the manuscript, I audited what's actually on disk. **Three independent defects, and every recorded evaluation has at least one.**

| set | files | recorded dates |
|---|---|---|
| `rubric10_semantic` | 36 | 2025-12-10 (12), 2026-05-13/14 (8), 2026-07-22 (16) |
| `rubric20_semantic` | 24 | 2026-05-13/14 (8), 2026-07-22 (16) |

**1. Ungrounded scoring.** 8 rubric20 and 20 rubric10 records predate the fix. The 16 dated 2026-07-22 fall on the same day as the commit and **cannot be placed either side of it** from the data alone.

**2. Wrong denominator.** All 24 `rubric20_semantic` records carry `max_points: 84` against the 88 the rubric's questions define — every semantic percentage ~4.8% high. `rubric10_semantic` is unaffected at 50 throughout.

**3. A split project name.** Three records carry `project: "AI"` and `method: "READI_claudecode"` — `AI_READI` split on its underscore. The parsing bug was fixed in `69aa303d` on **2025-12-08**; these records are dated **2025-12-10**, two days later, so the fix didn't cover the path that produced them.

## Why rescaling won't do

The denominator could be corrected arithmetically. **A score produced under an ungrounded prompt cannot** — it is not a score of the record, it is a score of the record plus its filename.

So the semantic path needs **re-running (60 evaluations across both rubrics)**, or dropping from reported results with the reason stated. That's a decision about the paper rather than the code, and I'm not making it here.

## What is safe

Presence and fitness are untouched by all three: presence has always scored rubric20 against 88, fitness has its own rubric and a cache keyed on `(axis, model, rubric, corpus, schema)`, and neither reads a `program` field. The agreement and form-defect measurements likewise.

Full write-up in `notes/semantic_evaluation_corpus_2026-08-03.md`.

**1054 passed, 3 skipped** (+5).

### DIFF
--- a/notes/semantic_evaluation_corpus_2026-08-03.md
+++ b/notes/semantic_evaluation_corpus_2026-08-03.md
@@ -1,0 +1,71 @@
+# The recorded semantic evaluations cannot go in the manuscript as they stand
+
+The semantic rubric agents are cited in `publication_summary_2026-07-22_fable-5.md`
+and `three_way_comparison_2026-07-22.md`. Auditing what is actually on disk
+turned up three independent problems, and every recorded evaluation has at least
+one of them.
+
+## What is on disk
+
+| set | files | recorded dates |
+|---|---|---|
+| `rubric10_semantic` | 36 | 2025-12-10 (12), 2026-05-13/14 (8), 2026-07-22 (16) |
+| `rubric20_semantic` | 24 | 2026-05-13/14 (8), 2026-07-22 (16) |
+
+## Problem 1 — ungrounded scoring (#162)
+
+Both agents were asked whether a description suits "the claimed dataset type
+**and program of origin**". No `program` or `project` slot exists anywhere in
+the D4D schema, so the inference falls back to the filename, the invocation
+context or prior knowledge — and the agents promise temperature 0.0 with
+"same file → same score".
+
+**Fixed in the prompts on 2026-07-22** (`0e19e85f`), which restricted the
+inference to quoted values in `keywords`, `publisher` or `funders` and forbade
+the filename explicitly. `tests/test_evaluation/test_semantic_agent_grounding.py`
+now pins that, including the premise — if the schema ever gains a `program`
+slot the test fails and the workaround can go.
+
+**The records predate or straddle the fix.** 8 rubric20 records and 20 rubric10
+records are from before it. The 16 dated 2026-07-22 fall on the same day as the
+commit and cannot be placed either side of it from the data alone.
+
+## Problem 2 — wrong denominator
+
+All 24 `rubric20_semantic` records carry `max_points: 84`. The rubric's own
+questions define 88 — 17 numeric at 5 plus 3 pass/fail — and the presence path
+has always used 88. So every semantic percentage is computed against a maximum
+the rubric does not have, and is roughly 4.8% high.
+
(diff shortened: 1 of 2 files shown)


realmarcin:
## Review

One finding, #285, fixed in ba8fd83d.

The test pinning #162's premise read `D4D_Core.yaml` for the absence of a `program` slot — **wrong scope twice over**:

- Records are evaluated as `Dataset`, which draws from the whole schema. A `program` slot added to any other module would have left the test green while the workaround it guards became unnecessary.
- The fields the prompt *may* use aren't in that file either: `publisher` and `keywords` live in `D4D_Base_import.yaml`, and `funders` is an **attribute** rather than a top-level slot — so `grep '^  funders:'` reports it missing when `schema_digest` reports it present.

It also never checked the three permitted fields **exist**, which is #162 in miniature: a test that stops a prompt grounding on nothing didn't check that what it grounds on is something. All three do (94 Dataset slots), but nothing said so.

Both halves now derive from `schema_digest.build("Dataset")`.

**1055 passed, 3 skipped.**

Issue #287, opened 2026-08-04
Scope semantic evaluation of the current corpus (there is nothing to re-run)


realmarcin:
Follow-on from #162. Scoping the "re-run 60 semantic evaluations" question turned up that it is not a re-run.

## The targets are gone

| set | files | target still on disk | target gone |
|---|---|---|---|
| `rubric20_semantic` | 24 | **0** | 24 |
| `rubric10_semantic` | 36 | 9 | 27 |

They point at two dead things: the `2026-04-10_sonnet-4.6` label, archived as unattestable, and flat pre-run-label paths removed when outputs moved into dated run directories. See #286 — the published summaries have the same problem across all four rubrics.

So the three defects in `notes/semantic_evaluation_corpus_2026-08-03.md` (ungrounded prompt, `max_points: 84`, split project name) are the lesser problem. Correcting a denominator on a score of an archived run produces a corrected number about nothing.

**One correction to that note**: it says the 16 records dated 2026-07-22 "cannot be placed either side of" the grounding fix. They can — the fable-5 exemplar manifest records `git_head` as `0e19e85f`, which *is* the grounding-fix commit. Those 16 were produced with the fix in place. It does not rescue them, because their inputs are archived, but the note overstates the uncertainty.

## The real question: what to evaluate, and how much

| coverage | evaluations | buys |
|---|---|---|
| canonical record per project, both rubrics | **8** | one semantic score per project |
| all 3 replicates, full only | 24 | replicate spread, comparable to fitness |
| all 3 replicates, full + core | 48 | matches the shape of the existing corpus |
| add v1 for a config comparison | 96 | semantic v1-vs-v2 |

**Recommend 8 or 24.** #169 established that four projects cannot resolve between-config differences near the noise floor, so the 96 option buys a comparison already known to be underpowered.

## Two things to settle first, or it gets done twice

1. **No canonical record has been marked** on the 2026-07-31 runs. Until `d4d runs select --execute` has run per project, "the semantic score for CHORUS" has no referent.
2. **Which prompt config.** v1, v2, or v3 — and v3 is drafted but unrun (#272). Evaluating v2 now means evaluating again if v3 lands.

Suggested order: select canonical → decide v3 → evaluate once.

## Preconditions already met

- The agents' grounding is fixed and pinned (#284).
- The denominator is 88 everywhere it is computed (#273), and mixed-maxima pooling is refused (#279), so new results cannot be silently averaged with the old ones.

[...]

Issue #286, opened 2026-08-04
Both published evaluation summaries score the archived 2026-04-10 sonnet baseline


realmarcin:
Found while scoping the semantic re-run (#162 follow-on). This is larger than the semantic path and affects all four rubrics.

## What they scored

**`publication_summary_2026-07-22_fable-5.md`** — titled *"D4D Publication Summary"*, presented as the exemplar run. Its `run_manifest_2026-07-22_fable-5.json` names the inputs:

```
data/d4d_concatenated/claudecode_agent/2026-04-10_sonnet-4.6/{AI_READI,CHORUS,CM4AI,VOICE}_d4d.yaml   GONE
data/d4d_concatenated/claudecode_agent/{AI_READI,CHORUS,CM4AI}_d4d.yaml                               GONE
```

**`three_way_comparison_2026-07-22.md`** — states it outright in its own header: *"Same D4D inputs (**2026-04-10 Sonnet 4.6 full + core outputs**)"*.

## Why that matters

`2026-04-10_sonnet-4.6` was **archived as unattestable** by `d4d runs archive --unattested --execute` (NEXT_TASKS §1). Its bundles were first committed after the runs executed, so the bytes those runs consumed cannot be verified. The whole point of that archival was to remove such runs from every analysis path at once.

So the document called "Publication Summary" reports 32 evaluations of records the study has excluded. Every number in its table — rubric10, rubric20, rubric10_semantic, rubric20_semantic, full and core — is a score of an archived record.

The remaining input paths are the pre-run-label flat layout (`claudecode_agent/AI_READI_d4d.yaml`), removed when outputs were reorganised into dated run directories.

## Nothing is lost

The scored records are preserved at `data/ATTIC/d4d_concatenated_archived/claudecode_agent/2026-04-10_sonnet-4.6/`, so the evaluations remain traceable to their inputs. This is a provenance and framing problem, not a data-loss one.

## What needs deciding

1. **These summaries cannot be cited as current results.** Either re-base them on the live corpus or retitle them as historical, with the archival status stated in the document rather than inferable from a manifest.
2. It follows that the semantic scoping question (#162 follow-on) is not "re-run 60 evaluations" — there is nothing to re-run against. See the companion issue.
3. Worth checking whether any other document in `data/evaluation_llm/` cites these tables.

Related: the memory of this repo already records these as "superseded runs" with template-derived model stamps, so the supersession was known; what was not recorded is that the summaries built on them are still presented as publication output.

PR #288, opened 2026-08-04
The semantic evaluations score records that are no longer in the study

realmarcin:
Corrects `notes/semantic_evaluation_corpus_2026-08-03.md`, written yesterday. Scoping the re-run found its headline was wrong and one claim in it was lazy.

## The targets are gone

| set | files | target on disk | gone |
|---|---|---|---|
| `rubric20_semantic` | 24 | **0** | 24 |
| `rubric10_semantic` | 36 | 9 | 27 |

They evaluate the `2026-04-10_sonnet-4.6` label — **archived as unattestable** — and flat pre-run-label paths removed when outputs moved into dated run directories.

That subsumes the three defects the note was about. Correcting a denominator on a score of an archived record produces a corrected number about nothing.

## And it isn't confined to the semantic path

`publication_summary_2026-07-22_fable-5.md`, titled **"D4D Publication Summary"**, scored the archived label across all four rubrics. `three_way_comparison_2026-07-22.md` states it in its own header. Filed as **#286**.

The records survive in `data/ATTIC/`, so the evaluations remain traceable to their inputs — a framing problem, not data loss.

## The lazy claim

I wrote that the 16 records dated 2026-07-22 *"cannot be placed either side of"* the grounding fix *"from the data alone"*. They can. `run_manifest_2026-07-22_fable-5.json` records `git_head` as `0e19e85f` — the grounding-fix commit itself — so those 16 were produced with the fix in place.

The manifest was the data, and I hadn't read it. It doesn't rescue them, since their inputs are archived, but the note overstated the uncertainty and that's worth correcting rather than leaving as a hedge that happened to be safe.

## Consequence

The "re-run 60 evaluations" framing is dropped. There is nothing to re-run against, so the question is what to **generate** against the current corpus — a scoping decision with options and a recommendation in **#287**.

Docs-only. **1055 passed, 3 skipped.**

### DIFF
--- a/notes/semantic_evaluation_corpus_2026-08-03.md
+++ b/notes/semantic_evaluation_corpus_2026-08-03.md
@@ -1,9 +1,17 @@
-# The recorded semantic evaluations cannot go in the manuscript as they stand
+# The recorded evaluations score records that are no longer in the study
 
-The semantic rubric agents are cited in `publication_summary_2026-07-22_fable-5.md`
-and `three_way_comparison_2026-07-22.md`. Auditing what is actually on disk
-turned up three independent problems, and every recorded evaluation has at least
-one of them.
+**The finding that matters is Problem 4 below: the evaluation targets are gone.**
+Not one of the 24 `rubric20_semantic` records points at a file that still
+exists, and the same is true of both published summaries across all four
+rubrics — not only the semantic ones (#286).
+
+The other three problems are real and were found first, which is why they are
+numbered first. They are worth less than they look: correcting a denominator on
+a score of an archived record produces a corrected number about nothing.
+
+Auditing what is actually on disk turned up **four** problems. Every recorded
+semantic evaluation has at least one of the first three, and all of them have
+the fourth.
 
 ## What is on disk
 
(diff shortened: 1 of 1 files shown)


realmarcin:
## Review

Two findings, #289–#290, both fixed in 9c082076. Both come from adding Problem 4 without revisiting the frame.

- **#289 — the one with consequence.** "What is safe to keep" said presence and fitness were *"untouched by all three problems"*, written before Problem 4 existed. Presence does not survive it: **8 of its 19 recorded targets are gone**, every one in the flat pre-run-label layout. That section exists to tell a reader which numbers they can still cite, so as written it invited citing eight figures describing files that aren't there. Now split — fitness clean on all four *with the evidence* (106 form failures attributing exactly to the current runs, 50 v1 / 56 v2, nothing unattributed), presence clean on 1–3 and partly affected by 4 with the count.
- **#290** — the opening still said "three independent problems" while the body ran to four, and Problem 4 is described as subsuming the others, so a reader trusting the count would stop before the finding that matters. Now leads with Problem 4, and the title says "recorded evaluations" rather than "semantic evaluations" since #286 showed the same defect across all four rubrics.

**1055 passed, 3 skipped.**

Issue #234, opened 2026-08-02
Regenerating the structural SSSOM mapping drops 10 rows


realmarcin:
Found while reviewing #232.

`make gen-sssom-structural` does not reproduce the committed
`data/semantic_exchange/d4d_rocrate_structural_mapping.sssom.tsv`. It emits
**150 rows against the committed 160**, silently dropping 10:

    d4d:CoreDataset
    d4d:CoreDatasetCollection
    d4d:CoreDistribution
    d4d:DatasetCollection/resources
    d4d:DataSubset
    d4d:File/file_type
    d4d:FileCollection/resources
    ... and 3 more

So the committed file cannot be regenerated from its declared inputs
(`$(D4D_SCHEMA_ALL) $(ROCRATE_JSON) $(SSSOM_STRUCTURAL_SCRIPT)`). Either the
generator has drifted since the file was produced, or the file was edited by
hand afterwards. Both are worth knowing: a mapping nobody can rebuild is a
mapping nobody can safely change.

The `Core*` losses are conspicuous — `CoreDataset` and friends live in a
separate merged schema (`data_sheets_schema_core_all.yaml`), and the target
declares only the full schema as an input, so a generator reading just that
input would never see them. That would make the committed file a *merge* of two
runs, which the Makefile does not express.

## Why this is filed rather than fixed

I hit it because #226 left two mapping files describing
`FileCollection.resources` as a `File | FileCollection` union, which the schema
no longer permits. Regenerating looked like the clean fix and instead lost ten
unrelated rows, so #232 corrects the two stale cells directly and leaves the
generator alone.

## Worth checking when addressed

- whether `Core*` rows need the core schema added as an input
- whether the committed file has hand edits that regeneration would discard
- adding a `--check` mode so drift between the file and its generator is caught
  rather than discovered

realmarcin:
Investigated. The report stands, and the root cause is now much narrower than
when I filed it — but there is also a **trap in verifying it** that I fell into
first, which matters more than the original finding.

## `make gen-sssom-structural` is usually a no-op

The target is a file rule:

    $(SSSOM_STRUCTURAL): $(D4D_SCHEMA_ALL) $(ROCRATE_JSON) $(SSSOM_STRUCTURAL_SCRIPT)

so `make` does nothing when the output is newer than its inputs — which it
normally is. Running it and diffing gives **byte-identical output and a false
clean bill**. I nearly closed this issue on exactly that basis. A forced rebuild
[...]

PR #291, opened 2026-08-04
Make the structural mapping's drift detectable, and diagnose why it drifted (#234)

realmarcin:
No regeneration of anything, and the committed mapping is untouched.

## The issue's hypothesis is wrong

#234 guessed the `Core*` rows were lost because the core schema isn't a declared input. Reproduced the 160 → 150 drop exactly, and the ten rows account for themselves **by capability, not by schema**:

| dropped | target | reason |
|---|---|---|
| 4 class-level rows | `schema:` | the generator emits **no class-level rows at all** |
| 2 `…/resources` | `schema:hasPart` | it produces no `schema:` targets |
| 3 file/collection attrs | `d4d:` | 21 committed, 18 regenerated |
| 1 `total_bytes` | `dcat:byteSize` | it produces no `dcat:` targets |

`class_uri` is parsed into `SchemaClass` and **never used to emit anything** — every `StructuralMapping` is built from `slot.parent_class`. So `DataSubset` is missing for the same reason `CoreDataset` is, and `DataSubset` is in the full schema. **Adding the core schema as an input would recover nothing.**

`_map_slot_uris` only emits a row when the RO-Crate input carries a matching property, and `fileType`, `collectionType`, `fileCount`, `byteSize` are not in `full-ro-crate-metadata.json`.

So the committed file is the output of a more capable generator than the one in the tree, or was partly written by hand. The ten rows are assertions no declared input supports. `total_bytes` is the clearest case: the schema declares `slot_uri: d4d:total_bytes` and the committed row maps it to `dcat:byteSize`.

## What this does instead

Restoring the missing capability is real feature work that would move other rows. This does the thing that makes the file safe to change in the meantime — the issue's own third suggestion.

`--check` regenerates to a temp file, diffs on `(subject, predicate, object)`, and exits non-zero on any difference. Exposed as `make check-sssom-structural`.

- **Identity is the triple, not the whole line.** Confidence and structural notes move without meaning changing, and a check that fires on those is one nobody keeps running.
- **It writes nothing.** A check that regenerates in place becomes the thing it detects; a test asserts the committed bytes are unchanged after running it.

## The gap is pinned, not blessed

`KNOWN_UNDERIVABLE` enumerates the ten with what is wrong with each. **New drift fails; the documented gap does not** — a check that has been red since the day it was written is a check nobody reads. Shrinking the set is progress; growing it needs a reason.

The other direction is asserted too: rows the generator produces that the file lacks would mean the file is *stale* rather than hand-extended, which is a different problem with a different fix.

## Answers to the issue's "worth checking" list

- **Do `Core*` rows need the core schema as an input?** No — see above.
- **Does the committed file have hand edits regeneration would discard?** All ten carry `semapv:StructuralMapping`, the generator's own justification, so they look generated — but the generator in the tree cannot produce them. Earlier generator or hand-authored; either way undeclared.
- **Add a `--check` mode?** Done.

**1059 passed, 3 skipped** (+4).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -366,7 +366,7 @@ SSSOM_URI_COMPREHENSIVE = src/data_sheets_schema/semantic_exchange/d4d_rocrate_s
 SSSOM_COMPREHENSIVE = src/data_sheets_schema/semantic_exchange/d4d_rocrate_sssom_comprehensive.tsv
 SSSOM_STRUCTURAL = data/semantic_exchange/d4d_rocrate_structural_mapping.sssom.tsv
 
-.PHONY: gen-core-schema validate-core lint-core gen-sssom gen-sssom-full gen-sssom-subset gen-sssom-uri gen-sssom-uri-comprehensive gen-sssom-comprehensive gen-sssom-structural gen-sssom-all clean-sssom
+.PHONY: check-sssom-structural gen-core-schema validate-core lint-core gen-sssom gen-sssom-full gen-sssom-subset gen-sssom-uri gen-sssom-uri-comprehensive gen-sssom-comprehensive gen-sssom-structural gen-sssom-all clean-sssom
 
 gen-core-schema: $(D4D_CORE_SCHEMA_ALL) ## Generate merged core exchange schema (data_sheets_schema_core_all.yaml)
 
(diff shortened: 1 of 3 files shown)


realmarcin:
## Review

Two findings, #295–#296, both fixed in e1dffd4f.

- **#295 — the substantive one.** `make gen-sssom-structural` writes **two** artifacts and the check read one. Regenerating shows they're in different states: the summary regenerates byte-for-byte, the `.tsv` doesn't (160 → 150). So the summary is current with the generator while the mapping isn't, and **it doesn't describe the file beside it** — it never mentions `CoreDataset`, `total_bytes` or `file_count`, which the `.tsv` asserts. That's the confusing direction: the artifact that's right is the one nobody thinks to distrust. It matters for #294, where fixing the `.tsv` could leave the summary silently correct for the wrong reason. `--check` now covers both and still writes nothing.
- **#296** — `read_sssom_rows` indexed its three columns directly, so a malformed file failed with a bare `KeyError` from inside `csv`, sending the reader to the code rather than the input. A headerless file returned zero rows, making `check_drift` report every committed row as lost — loud, but a misleading way to say "no header". Both now raise naming the file and the column. Reachable by whoever points `check_drift` at a hand-edited file, which is exactly what fixing #294 involves.

**1064 passed, 3 skipped.**

Issue #292, opened 2026-08-04
No VOICE replicate is shippable: all three fail on related_datasets, each differently


realmarcin:
Found while establishing the canonical records #287 needs.

`d4d runs select --project VOICE --config 2026-07-31_claude-opus-5-generic-v2`:

> Error: no replicate of VOICE is eligible. Nothing here can be canonical, and picking the least-bad would ship a record known to be broken.

**VOICE is the only project that populates `related_datasets` at all** — the other three omit it — and every failure is in that one slot:

| replicate | `target_dataset` | `relationship_type` | fails on |
|---|---|---|---|
| rep1 | **dict** (schema declares `string`) | `has_part` ✓ permitted | `target_dataset` |
| rep2 | str ✓ | **`related_to`** — not in the enum | `relationship_type` |
| rep3 | str ✓ | **`References`** — capitalised; `references` is permitted | `relationship_type` |

`DatasetRelationshipTypeEnum` has 36 permissible values after the DataCite alignment (#223). `related_to` is not among them; `references` is, and rep3 emitted `References`.

## Two distinct defects

**1. A structure where a scalar is declared.** rep1 put a whole nested Dataset object — id, name, title, description, doi, version, licence, keywords — into `target_dataset`, whose range is `string` with `slot_uri: dcterms:relation`. This is the same family as the form defects in #271: correct content, wrong shape.

**2. Enum values outside the vocabulary**, one of them a pure casing difference from a permitted value.

## Why it matters beyond VOICE

- **It blocks #287.** "One canonical record per project" is not available: AI_READI, CHORUS and CM4AI each select cleanly, VOICE has nothing eligible. Any semantic evaluation of canonical records covers three projects, not four.
- It is invisible in the other three projects **only because they omit the slot**, so this is not evidence that generation handles relationships correctly elsewhere — it is evidence nothing else exercised it.

## Worth considering

- Whether `target_dataset` should accept an inline Dataset as well as a reference. VOICE genuinely has a companion pediatric dataset with its own DOI, and a bare string loses everything the model found about it.
- Whether the enum should accept case-insensitive matches, or whether generation should be constrained to emit permissible values — `References` failing while `references` passes is a validation failure with no semantic content.
- A generation rule for "a slot whose range is a scalar takes an identifier, not the object it identifies" — the companion to the v3 rule drafted in #272, and the same defect family.

realmarcin:
Re-validated against the current schema. All three replicates still fail, by
[...]

PR #293, opened 2026-08-04
Mark the canonical records, and label the historical summaries as historical (#286, #287)

realmarcin:
Addresses #287's actionable precondition and #286. No API calls, nothing regenerated.

## #287 — canonical records

`d4d runs select --execute` on the generic-v2 config:

| project | canonical | slots |
|---|---|---|
| AI_READI | rep1 | 70 |
| CHORUS | rep2 | 50 (1 ahead of the runner-up) |
| CM4AI | rep2 | 72 |
| **VOICE** | **nothing eligible** | — |

Each winner's provenance gains a `canonical` block naming all three candidates, the criterion and the margin. Nothing was moved or copied.

### VOICE has no canonical record, and it's one slot

VOICE is **the only project that populates `related_datasets` at all**, and all three replicates fail there — each differently:

| replicate | failure |
|---|---|
| rep1 | a whole nested `Dataset` where the range is `string` |
| rep2 | `related_to` — not among the 36 DataCite-aligned enum values |
| rep3 | `References` — `references` is permitted, this is casing |

Filed as **#292**. It changes #287's arithmetic: "one canonical record per project" covers **three** projects, so the 8-evaluation option is **6** until VOICE is fixed.

It's also worth noting the other three projects pass here only because they *omit* the slot — that's not evidence relationships are handled correctly, it's evidence nothing else exercised them.

## #286 — the summaries

All three scored the archived label. `publication_summary_2026-07-22_fable-5.md` is titled *"D4D Publication Summary"*; `three_way_comparison` says so in its own header; the opus-4.8 summary's manifest names the same inputs.

Each now opens with `[HISTORICAL]` and a banner stating what it scored, that the label was archived as unattestable and why, that the records survive in ATTIC so the numbers stay traceable, and not to cite them as current. The generating script carries the same note — including that re-pointing it at the current corpus means re-running the **evaluations**, not re-running it.

**Retitled rather than re-based**, because re-basing requires evaluations that don't exist yet. That's #287, and it shouldn't be done twice.

The `.tsv` companions are left alone: machine-read, with the warning on the `.md` a reader actually opens.

**1055 passed, 3 skipped.**

### DIFF
--- a/data/d4d_concatenated/claudecode_agent_core/2026-07-31_claude-opus-5-generic-v2_rep1/AI_READI_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-07-31_claude-opus-5-generic-v2_rep1/AI_READI_provenance.yaml
@@ -150,3 +150,21 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-07-31_claude-opus-5-generic-v2_rep1/AI_READI_d4d_core.yaml
       md5: 0e9f0914e5f028f0944e724afb4e50bb
   recorded_by: api_runner.execute
+canonical:
+  criterion: full and core both validate against the current schema, then most slots,
+    then lowest label
+  selected_from:
+  - label: 2026-07-31_claude-opus-5-generic-v2_rep1
+    slots: 70
+    validation: valid
+    validation_recorded_at_run_time: valid
+  - label: 2026-07-31_claude-opus-5-generic-v2_rep2
+    slots: 69
+    validation: invalid (full)
+    validation_recorded_at_run_time: invalid
+  - label: 2026-07-31_claude-opus-5-generic-v2_rep3
+    slots: 70
+    validation: invalid (full)
+    validation_recorded_at_run_time: invalid
+  margin_over_runner_up: null
+  selected_by: d4d runs select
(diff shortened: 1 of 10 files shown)


realmarcin:
## Review

Two findings, #305–#306. #305 fixed in b38460c9; #306 filed.

**Rebased onto 13 commits of main** and re-verified the selection is unchanged: AI_READI rep1, CHORUS rep2, CM4AI rep2, and no eligible VOICE replicate. The schema and normaliser changes that landed meanwhile did not move it.

- **#305** — I marked the three summaries `[HISTORICAL]` in their `.md` and left the `.tsv` companions alone, reasoning they were "machine-read, with the warning on the `.md` a reader actually opens". **That has it backwards.** A person opens the `.md` and sees the banner; a script reads the `.tsv` and sees archived-run numbers with nothing to say what they are — and a script is the likelier consumer of a long-format table. These are the files someone pulls into a notebook for the manuscript, and their rubric20 rows are both ~4.8% high *and* about excluded records. Banner added to all three, and to the writer so regeneration cannot drop it.
- **#306 — filed, not fixed.** Nothing consumes the `canonical` block this PR writes. It is well-formed, and #176 called selection the answer to *"which record **is** the CHORUS datasheet"* — but the only reader in the tree is the test for the writer. Every analysis path still treats all three replicates alike. That is the missing other half rather than a defect here, and it is first genuinely needed by #287's scoping: "one canonical record per project" is 8 evaluations only if something can enumerate them.

**1095 passed, 3 skipped.**

Issue #297, opened 2026-08-04
LinkML cannot express a string-or-inline-object range; target_dataset stays a reference


realmarcin:
From #292, where the decision was to widen `DatasetRelationship.target_dataset` to accept an inline `Dataset` as well as an identifier. It is not expressible, and the evidence is worth keeping so nobody tries again from scratch.

## What was tried

`any_of` with both branches, in three arrangements. The compiled JSON Schema in each case:

**No explicit range** (`any_of: [string, Information]`):
```json
{"anyOf": [{"type": "string"}, {"type": "string"}], "type": "string"}
```
A class range inside `any_of` compiles to `{"type": "string"}` — LinkML treats it as a reference by identifier unless the branch is `inlined`.

**With `inlined: true` on the object branch:**
```json
{"anyOf": [{"type": "string"}, {"$ref": "#/$defs/Information"}], "type": "string"}
```
Correct branches, but the residual `"type": "string"` — from `default_range: string` in `data_sheets_schema.yaml` — **ANDs** with the `anyOf`, so an object still fails.

**With `range: Information, inlined: true` at slot level:**
```json
{"$ref": "#/$defs/Information", "anyOf": [{"type": "string"}, {"$ref": "#/$defs/Information"}]}
```
The inverse: the top-level `$ref` now forces an object, so a **string** fails.

Confirmed on the real records:

| replicate | `target_dataset` | validates under the third arrangement |
|---|---|---|
| VOICE rep1 | object | ✅ |
| VOICE rep2 | string | ❌ |

LinkML always emits a top-level type or `$ref` from the slot's range, and JSON Schema requires it *and* the `anyOf` to pass. Whichever branch the range names, the other fails. There is no third option — the slot always has a range, explicit or defaulted.

## What was done instead

`target_dataset` stays `range: string` and its **description was corrected to match** — it used to promise "identifier, URL, or Dataset object" while permitting only the first, which is the self-contradiction that made this look like a simple widening.

A related dataset with its own DOI, protocol and ethics approval **is its own datasheet** and is referenced from here by identifier. That is the resolution for VOICE, whose pediatric companion has all three.

## If this is revisited

- `Dataset` cannot be named as a range from `D4D_Composition.yaml` at all — it is declared in the importing schema, so the reference would be circular. `Information` was used as the object branch because `Dataset is_a Information`.
- Removing `default_range: string` would change every slot in the schema that relies on it, which is not a trade worth making for one slot.
[...]

PR #298, opened 2026-08-04
Honour the enum aliases the schema declares; the range union is not expressible (#292)

realmarcin:
Two of the three #292 decisions, plus evidence the third can't be done as asked. The VOICE fork is deliberately not here — reasons at the end.

## Widening `target_dataset` — not expressible in LinkML

Tried `any_of` three ways:

| arrangement | compiled JSON Schema | result |
|---|---|---|
| `any_of: [string, Information]` | `{"anyOf": [{"type":"string"}, {"type":"string"}], "type":"string"}` | class range → string unless `inlined` |
| `+ inlined: true` | `{"anyOf": [{"type":"string"}, {"$ref":"…Information"}], "type":"string"}` | right branches, residual `type` ANDs — objects fail |
| `range: Information, inlined` | `{"$ref":"…Information", "anyOf": […]}` | **inverts** — objects pass, strings fail |

Confirmed on real records: VOICE rep1 (object) and rep2 (string), one passing each way. LinkML always emits a top-level type or `$ref` from the slot's range, and there's no arrangement without one. Filed as **#297**.

**What was actually wrong was subtler than the range.** The description promised *"identifier, URL, or Dataset object"* while permitting only the first — the schema contradicted itself, which is what made this look like a simple widening. The description now matches the range, and the comments record why a reference is the right shape.

## Case-insensitive enum matching — too narrow, so it does more

Measured across the 24 generic and generic-v2 records, **14 enum values fall outside their vocabulary**:

| value | count | kind |
|---|---|---|
| `IsNewVersionOf` | 6 | **declared alias** |
| `is a later release in the same series as` | 4 | prose, no equivalent |
| `HasPart` | 2 | **declared alias** |
| `related_to` | 1 | no equivalent |
| `References` | 1 | casing only |

**Case-insensitivity alone fixes 1 of 14.** The schema already declares DataCite's CamelCase spellings as `aliases` — and those are what generation emits, because that's what the vocabulary is called everywhere else. `linkml-validate` matches on `text` alone, so **the schema declared a name valid and validation rejected it**.

`normalise_enum_aliases` honours the declared aliases and folds case: **9 of 14**. The table is read from the schema, not listed in code — a second copy of a vocabulary is one to keep in step. It runs beside `normalise_temporal`, text-level for the same reason (re-dumping would drop the `#` provenance header).

The remaining five are **left to fail**. Only bare word-shaped scalars are touched, so prose stays visible — normalising `is a later release in the same series as` into something valid would hide a defect.

## The VOICE fork — a study-structure decision, not an implementation one

The evidence is there: the VOICE bundle has 100 mentions of "pediatric", 13 of `b2ai-voice-pediatric`, 11 of the DOI. It is generatable.

What stops me doing it unilaterally is the cost:

- **`PROJECTS = ["AI_READI", "CHORUS", "CM4AI", "VOICE"]`**, and **26 files** hardcode that four-way list.
- The pipeline keys one record per project (`{project}_d4d.yaml`), so a second VOICE dataset is not expressible without either a fifth project or a change to that assumption.
- Every "four projects" claim in the study changes, **including #169's power analysis**, which concluded four cannot resolve effects near the noise floor.

Three shapes, and they are not equivalent:

1. **`VOICE_PEDIATRIC` as a fifth project** — cleanest for the pipeline, but it is not a fifth Grand Challenge, and the arms/comparisons are organised by GC.
2. **A second record under VOICE** — truest to the science, needs the one-record-per-project assumption relaxed.
3. **A standalone record outside the study corpus**, referenced by VOICE — cheapest, keeps the four-project claims intact, but the pediatric datasheet is then not in any analysis.

It also needs a generation run, so it is not free either.

**1075 passed, 3 skipped** (+11).

### DIFF
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -283,6 +283,38 @@ a budget on every tracked cache under `agreement_cache/`.
 
 ---
 
+## 9. Split VOICE into two datasets — registered, not generated
+
+`VOICE_PEDIATRIC` is a project as of #298. The companion pediatric dataset has
+its own DOI (10.13026/h995-bt35), protocol and Research Ethics Board approval,
+and was being represented as a nested object inside VOICE's `related_datasets` —
+which is why no VOICE replicate validated (#292).
+
+**Done:** the registry, the shared-corpus declaration, and the schema correction
+that made the nested form wrong rather than ambiguous. Adding the fifth project
+broke nothing; the four-project list in 26 files is prose, not logic.
+
+**Not done, and each needs a decision:**
+
+- **A bundle and manifest line.** Both datasets are documented in the same VOICE
+  corpus, so they share a bundle and must be distinguished by `{MANIFEST_LINE}`,
+  which is already a substitution field. Scoping via the manifest keeps the
+  frozen prompt bodies untouched — the four-project list lives in the preamble,
+  not the body, so no existing arm's resolved text changes.
+- **A generation run.** Twelve phases for one project, paid.
+- **Whether `VOICE` becomes `VOICE_main`.** Not done: 198 of the ~580 affected
+  paths are archived records, archived because their provenance could not be
+  verified, and rewriting their project field is the opposite of what archiving
+  them was for. Renaming only the live corpus splits the series instead. The
+  live-only migration is available if the asymmetry matters more than the split.
+
+**Consequence for #169.** The power argument counts four projects. These two
+share a source corpus, so they are not two independent samples, and
+`SHARED_CORPUS_GROUPS` records that for any analysis that would otherwise
+count them as such.
+
+---
+
 ## Standing constraints
 
 - **No gold standard exists.** `curated` was generated through a ChatGPT chat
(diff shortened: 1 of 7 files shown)


realmarcin:
## Review

One finding, #299, fixed in 25925426. **The normaliser I added here shipped a worse bug than the one it fixed.**

`_enum_aliases` flattened every alias of every enum into one table and applied it to any `key: Value` line, without checking the slot's range was that enum — or an enum at all:

```
title: References       ->  title: references
name: HasPart           ->  name: has_part
description: Collects   ->  description: collects
```

A dataset legitimately titled `References` was silently retitled **on the write path, in the committed record, by the pass meant to be fixing validity.** The enum failures it was written for were at least loud — `linkml-validate` rejected them and `d4d runs select` refused to make the record canonical. This was not.

The table is now keyed by slot and built only from enum-ranged slots, so `title`, `name` and `description` are unreachable rather than unlikely. A slot ranged on different enums in different classes is dropped rather than guessed at — text alone cannot say which class a line sits in — which also resolves the `'other'` collision across three enums.

Two smaller faults from the same review: the value pattern was letters-only, so **41 permissible values carrying digits** (`BZ2`, `Big5`, `GB2312`) could never be normalised while their enums appeared covered.

Still **9 of 14** against the same corpus counts.

**1083 passed, 3 skipped.**

PR #300, opened 2026-08-04
Scope VOICE_PEDIATRIC by its bundle (NEXT_TASKS §9)

realmarcin:
The free half of §9, plus a correction to what I wrote there.

## `{MANIFEST_LINE}` was the wrong mechanism

I claimed the two datasets could be distinguished by that field. **They can't** — it is substituted into the *output record's* header block as a provenance comment, not into the instruction.

The frozen body says *"DECLARED INPUT BUNDLE — your only source of dataset facts: {BUNDLE}"*, so **the bundle is the scope**. Scoping there needs no change to any frozen prompt, which is the property I wanted from the manifest line and didn't get.

## The pediatric selection

Six sources, led by the pediatric PhysioNet record — which carries **all 11** references to the dataset's DOI — plus the IRB protocol, project documentation, NIH RePORTER entry, DUA and docs repository, all of which cover the pediatric cohort explicitly.

| | files | size | `h995-bt35` | adult version pages |
|---|---|---|---|---|
| `VOICE` | 11 | 378 KB | 11 | present |
| `VOICE_PEDIATRIC` | 6 | 204 KB | 11 | **0** |

**Excluded**: the three adult PhysioNet version pages and the two adult-cohort papers. They carry zero "pediatric" mentions, and including them would import adult facts into a pediatric datasheet — the failure the fitness axis calls `target`, which is the thing the study measures.

**The VOICE bundle is untouched and byte-identical** — md5 `e637eb75`, what the 49 runs subject to the live-provenance rule attest. A test pins it, because the whole reason VOICE keeps its identifier is that its attested artifacts must not move.

## The rename — decided against, on evidence

You chose option C, and having attempted it I can now say why it was the right call rather than just arguing it.

The 567 path moves were clean: no clashes, no `VOICE_PEDIATRIC` mangling. **The content rewrite broke attestation two ways that care does not avoid:**

1. Record bodies carry `VOICE` in titles, ids and prose, and provenance records their **byte counts** — rewriting them failed **33 runs** outright.
2. `source_manifest.yaml` is keyed by project and its md5 is recorded by **115 runs**, so renaming the project necessarily changes a file those runs attest.

Renaming converts attested runs into reconstructed ones — the same property that got the 2026-04-10 series archived. Reverted; nothing was pushed.

## What remains in §9

The generation run — twelve phases for one project, paid. Everything it needs is now in place.

**1092 passed, 3 skipped** (+9).

### DIFF
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -296,17 +296,31 @@ broke nothing; the four-project list in 26 files is prose, not logic.
 
 **Not done, and each needs a decision:**
 
-- **A bundle and manifest line.** Both datasets are documented in the same VOICE
-  corpus, so they share a bundle and must be distinguished by `{MANIFEST_LINE}`,
-  which is already a substitution field. Scoping via the manifest keeps the
-  frozen prompt bodies untouched — the four-project list lives in the preamble,
-  not the body, so no existing arm's resolved text changes.
+- **A scoped bundle — done.** `{MANIFEST_LINE}` was the wrong mechanism and the
+  claim here was wrong: it is substituted into the *output record's* header
+  block as a provenance comment, not into the instruction. The frozen body says
+  "DECLARED INPUT BUNDLE — your only source of dataset facts: {BUNDLE}", so the
+  bundle is the scope. `VOICE_PEDIATRIC` now has a manifest selection of six
+  sources — the pediatric PhysioNet record plus the programme documents that
+  cover it — and its own 204 KB bundle. The three adult version pages and the
+  two adult-cohort papers are excluded: they carry no pediatric content, and
+  including them would import adult facts into a pediatric datasheet, which is
+  the failure the fitness axis calls `target`.
+
+  The VOICE bundle is byte-identical (md5 `e637eb75`, what 49 runs attest), and
+  a test pins it.
 - **A generation run.** Twelve phases for one project, paid.
-- **Whether `VOICE` becomes `VOICE_main`.** Not done: 198 of the ~580 affected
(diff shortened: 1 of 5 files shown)


realmarcin:
## Review

One finding, #302, fixed in 8106a255. **I had recreated #234 — one week after filing it.**

`d4d download concatenate --project VOICE_PEDIATRIC` failed outright:

```
❌ Error: Input directory not found: data/preprocessed/individual/VOICE_PEDIATRIC
```

because `VOICE_PEDIATRIC` is a *selection over VOICE's downloads*, not a project with its own directory. The committed bundle had been built with a bespoke `--input-dir` override recorded **nowhere** — not the Makefile, not the manifest, not the bundle header. So the artifact was exactly what #234 is about, and it matters more here: a generation run will attest this bundle's md5, and an input nobody can regenerate is one nobody can check.

The source directory is now declared in the manifest beside the selection — the manifest already says *which* files a project selects, so the directory they come from belongs in the same place. The CLI reads it only when `--input-dir` is at its default, so an explicit override still wins.

**Verified:** the documented command reproduces the committed bundle byte-for-byte, and all four existing projects still reproduce theirs. VOICE's md5 is still `e637eb75` — what its 49 attested runs record.

Also checked while reviewing: the manifest append assumed `projects:` was the last top-level key. It was, so the block landed correctly — but that was luck rather than a check, and it is worth knowing the file's shape carried it.

Not fixed, unrelated: `Makefile:84` advertises `make concat-preprocessed` and no such target exists. Noted in #302 — it is why I went to the CLI in the first place.

**1095 passed, 3 skipped.**

PR #301, opened 2026-08-04
Defer the VOICE_PEDIATRIC run to v3 (NEXT_TASKS §9)

realmarcin:
Records the decision to wait for v3 rather than generate `VOICE_PEDIATRIC` under v2, and one consequence that is easy to miss.

**Stacked on #300** (base `voice-pediatric-manifest`), so this diff shows only the deferral. Both PRs edit NEXT_TASKS §9 — branching from `main` meant this one carried the stale text #300 corrects, and merging both would have restored the wrong claim about `{MANIFEST_LINE}`.

## The deferral

Generating under v2 now means generating again when v3 lands. `VOICE_PEDIATRIC` waits for whichever config the v2-vs-v3 comparison settles on. v3 is drafted, wired and tested (#272) but unrun.

It **cannot join that comparison itself** — no v1 or v2 baseline exists for it, so it is generated once under the winning config rather than standing as a third arm.

## Regenerate VOICE in the same run

Its three 2026-07-31 replicates are why no VOICE record is canonical (#292). Both causes are now addressed **for future runs but not for those records**:

| cause | status |
|---|---|
| `References` (casing) | normalised on the write path → `references` |
| `related_to` | still fails — correctly, it names nothing in the vocabulary |
| inline object in `target_dataset` | should not recur: the pediatric dataset is now its own datasheet, so VOICE references it by DOI — a string, which validates |

So **a v3 run is the first opportunity for VOICE to have a canonical record at all.** Until then it has none, and any "canonical record per project" count is three of four rather than four.

Docs-only. **1092 passed, 3 skipped**, with #300 underneath.

### DIFF
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -309,7 +309,29 @@ broke nothing; the four-project list in 26 files is prose, not logic.
 
   The VOICE bundle is byte-identical (md5 `e637eb75`, what 49 runs attest), and
   a test pins it.
-- **A generation run.** Twelve phases for one project, paid.
+- **A generation run — deferred to v3, by decision.** Running under v2 now would
+  mean running again when v3 lands, so `VOICE_PEDIATRIC` is generated with
+  whichever config the v2-vs-v3 comparison settles on. v3 is drafted, wired and
+  tested (#272) but unrun.
+
+  `VOICE_PEDIATRIC` cannot join that comparison itself — it has no v1 or v2
+  baseline, so it is generated once under the winning config rather than as a
+  third arm.
+
+  **Regenerate VOICE in the same run.** Its three 2026-07-31 replicates are the
+  reason no VOICE record is canonical (#292), and both causes are now addressed
+  for future runs but not for those records:
+
+  - the enum values: `References` normalises to `references` on the write path,
+    and `related_to` still fails, which is correct — it names nothing in the
+    vocabulary;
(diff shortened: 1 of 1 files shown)


realmarcin:
## Review

Two findings, #303–#304, both fixed. Both are the same error — describing a state that does not exist yet as though it does.

- **#304 hid a real question.** The text said *"with the pediatric dataset now its own datasheet, VOICE should reference it by DOI"* — **in the bullet that defers generating that datasheet.** Zero `VOICE_PEDIATRIC` records exist; it is a registered project with a bundle. The reasoning survives, but the tense was doing work it shouldn't: **the pediatric record has to exist before VOICE can point at it.** Generate `VOICE_PEDIATRIC` first, or a regenerated VOICE either inlines the dataset again and fails again, or drops the relationship. That ordering is a live constraint on the v3 run, and the original wording concealed it.
- **#303** — *"three of four"* holds only if #293 merges. On `main` it is **zero of four**: no provenance record carries a `canonical` block, and the three marks live in that still-open PR. Now states both.

Also corrected: the branch had been sitting on `voice-pediatric-manifest` as it stood *before* the #302 fix, so it measured 1092 while I wrote 1095. Rebased onto `main`; it is genuinely 1095 now. Second time this turn I quoted a count from the wrong base.

Docs-only. **1095 passed, 3 skipped.**

Issue #306, opened 2026-08-04
Nothing consumes the canonical mark, so 'which record is the CHORUS datasheet' still needs a human


realmarcin:
From the review of #293.

That PR records a `canonical` block in the winning replicate's provenance for AI_READI, CHORUS and CM4AI. The block is well-formed — criterion, margin over runner-up, all three candidates named — and #176 called selection the answer to *"which record **is** the CHORUS datasheet"*.

But nothing reads it. Searching the tree for a consumer finds one hit, and it is the test for the writer:

```
tests/test_cli/test_runs_select.py:131:  canon = prov.get("canonical")
```

So the question #176 posed is answered in a file, not by the pipeline. Anyone asking which record to ship still has to know that `d4d runs select` was run, find the right provenance file, and read it. Every analysis path — `compare`, `arm_delta`, the evaluation scripts — still treats all three replicates alike.

This is not a defect in #293, which deliberately marks rather than moves ("nothing is moved or copied"). It is the missing other half: a mark with no reader.

The cheapest useful consumer would be a resolver — `d4d runs canonical --project CHORUS` printing the path, or a `--canonical-only` filter on `runs list`. That turns the mark into something the semantic-evaluation scoping in #287 can point at, which is where it is first actually needed: "one canonical record per project, both rubrics" is 8 evaluations only if something can enumerate them.

realmarcin:
Fixed — `d4d runs canonical` is the resolver this asked for, and its help text
cites this issue:

```
Usage: d4d runs canonical [OPTIONS]

  Which record is the datasheet, per project.

  `select` marks a replicate canonical and nothing read the mark (#306), so
  the question it answers was answerable only by someone who knew to open
  provenance. This reads it.
```

`--paths-only` enumerates the canonical records for an evaluation sweep, and
`--missing` names the projects that have none, which is the count #287 needs.

Verified against the corpus: three of four projects resolve, VOICE absent while
no replicate validates (#292).

Issue #299, opened 2026-08-04
normalise_enum_aliases rewrites values in slots that are not enums


realmarcin:
From the review of #298. This one corrupts records.

`_enum_aliases()` flattens every alias of every permissible value across **all** enums into one table, and `normalise_enum_aliases` applies it to any line matching `key: Value`. It never checks that the slot's range is the enum the alias came from — or that it is an enum at all:

```
    title: References        ->  title: references
    name: HasPart            ->  name: has_part
    description: Collects    ->  description: collects
```

`title`, `name` and `description` are plain strings. A dataset legitimately titled `References`, or a `name` that happens to read `HasPart`, is silently rewritten on the write path, in the same pass that is supposed to be *fixing* validity.

This is worse than the problem it was added for. The enum failures were loud — `linkml-validate` rejected them and `d4d runs select` refused to make the record canonical. This is silent, and it lands in the committed record.

## Two smaller faults in the same function

**Cross-enum collision.** `'other'` is a permissible value of `CreatorOrMaintainerEnum`, `FileCollectionTypeEnum` *and* `FileTypeEnum`. The flat table keeps whichever was built last, so a value can be rewritten to another enum's spelling. Harmless today because all three spell it identically, but the table cannot express that it is three different values.

**Values with digits are unreachable.** The value pattern is `[A-Za-z][A-Za-z_]*`, and 41 permissible values contain digits — `BZ2`, `Big5`, `GB2312`, `application/x-bzip2`. None can ever be normalised, so the function silently does nothing for a whole class of enums while appearing to cover them.

## Fix

Resolve the slot's range before rewriting: build `(class, slot) -> enum` from the schema, track the enclosing class while walking, and only consult that enum's aliases. That is what makes it a *normaliser* rather than a global find-and-replace, and it fixes the collision case for free.

The digit gap then matters less, but the pattern should still admit them.

Note this changes the approach from text-level to structure-aware, which conflicts with the reason it is text-level: re-dumping the YAML drops the `#` provenance header. Resolving the range needs to know which class a line sits in, which text alone does not tell you. Worth deciding whether a narrower rule — only rewrite lines whose key is a slot with an enum range **anywhere** in the schema — is enough, since that would exclude `title`, `name` and `description` outright while staying text-level.

realmarcin:
[...]

Issue #277, opened 2026-08-04
form_defects: load_form_failures ignores the rubric and model recorded on each fitness judgement


realmarcin:
From the review of #271.

`load_form_failures` selects on the failure class alone:

```python
if entry.get("failure") != "form":
    continue
```

Every fitness cache entry also carries `rubric` and `model` — the fields `JudgementContext` exists to make a judgement self-describing, and which `JudgementContext.mismatch()` uses to say *which* dimension makes a stored entry unusable.

Today the cache holds exactly one of each, so nothing is mixed:

```
form failures by fitness rubric: {'6c5b10778ade': 106}
                       by model: {'google/claude-opus-5-high': 106}
```

But the module's whole premise is that `FITNESS_SYSTEM` will not be edited, *because* editing it invalidates the cache. If it ever is edited — or the fitness judge is run under a second model — this loader silently pools form failures judged under both, and the sub-type table would mix two instruments while looking exactly the same.

That is the defect fixed for the equivalence judge in #243 and for the sub-type classifier's own cache in this same PR, which does filter on rubric and model. The loader is the one place that does not.

Fix: accept `rubric` and `model` as arguments, default them to the values the published table used, and skip entries that do not match — or fail loudly when the loaded set spans more than one of either, which is arguably better since a silent skip could quietly halve the corpus.

realmarcin:
Fixed in 34a49398 (#271). `load_form_failures` now raises when the loaded set spans more than one fitness rubric or model, rather than filtering — a silent skip could halve the corpus and still produce a plausible table. Two tests: the real cache holds exactly one of each (the premise the design rests on), and a hand-built mixed cache is refused.

PR #307, opened 2026-08-04
Make the canonical mark readable, so #287 can be sized (#306)

realmarcin:
#287's blocker was never the decision — it was that nothing could enumerate what to evaluate.

## The mark had no reader

`d4d runs select` wrote a `canonical` block and the only reader in the tree was the test for the writer. #176 called selection the answer to *"which record **is** the CHORUS datasheet"*, but that was an answer only for someone who already knew to open provenance and which file to open. Every analysis path still treated all three replicates alike.

`canonical_runs()` returns `project -> {label, method, criterion, candidates, full, core, provenance}`, and `d4d runs canonical` reads it:

```
$ d4d runs canonical --paths-only
data/d4d_concatenated/claudecode_agent/…rep1/AI_READI_d4d.yaml
data/d4d_concatenated/claudecode_agent_core/…rep1/AI_READI_d4d_core.yaml
data/d4d_concatenated/claudecode_agent/…rep2/CHORUS_d4d.yaml
…
6 records

$ d4d runs canonical --missing
VOICE
VOICE_PEDIATRIC
```

A project with no eligible replicate is **absent rather than guessed at**, and asking for one by name exits non-zero rather than printing nothing — silence would read as "no canonical needed" instead of "none exists".

## It also corrects #287's arithmetic

The table there called *"canonical record per project, both rubrics"* **8 evaluations** — 4 projects × 2 rubrics, implicitly full records only. But the existing semantic corpus evaluates **both variants**: `rubric20_semantic` is 12 full and 12 core.

With three canonical projects:

| shape | evaluations |
|---|---|
| full only, 2 rubrics | **6** |
| full + core, 2 rubrics | **12** — matches the existing corpus |

Neither is wrong; they measure different things, and the table stated one number without saying which. The resolver prints the record set, so the count now follows from the set rather than from a number in a note.

## What this does not do

No evaluations are run. Generation is deferred to v3 (NEXT_TASKS §9), and evaluating v2 canonicals now would mean evaluating again when v3 lands — which is the reason #287 listed the preconditions in the first place.

**1105 passed, 3 skipped** (+10).

### DIFF
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -152,14 +152,20 @@ def canonical_cmd(project, config, paths_only, missing):
     replicate validates (#292).
     """
     from data_sheets_schema.constants import PROJECTS
-    from data_sheets_schema.runs import canonical_runs
+    from data_sheets_schema.runs import AmbiguousCanonical, canonical_runs
 
-    found = canonical_runs(config=config)
+    try:
+        found = canonical_runs(config=config)
+    except AmbiguousCanonical as exc:
+        # Naming the configurations, because --config is the answer and the
+        # user cannot pass it without knowing what to pass (#308).
+        click.echo(f"{exc}", err=True)
+        raise SystemExit(2)
     if project:
         found = {k: v for k, v in found.items() if k == project}
 
     if missing:
-        gap = [p for p in PROJECTS if p not in canonical_runs(config=config)]
+        gap = [p for p in PROJECTS if p not in found]
         if project:
             gap = [p for p in gap if p == project]
         for p in gap:
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -130,6 +130,68 @@ def restore(labels, projects, execute):
         click.echo("\nDry run. Re-run with --execute.")
 
 
+@runs.command("canonical")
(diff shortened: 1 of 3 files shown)


realmarcin:
## Review

One finding, #308, fixed in a263aea7 — and it was live for the very next step.

`canonical_runs` built `out[project] = …` while walking, so a project marked canonical under two configs kept whichever label **sorted last**, silently.

That is not hypothetical: NEXT_TASKS §9 plans to re-select canonical after a v3 run, and `select --execute` does not clear a previous mark — so AI_READI, CHORUS and CM4AI would each carry two. `--paths-only` would then feed an evaluation sweep whichever sorted last, and **since v3 labels sort after v2 ones it would have been right by accident**, which is worse than wrong. It would work until a label broke the pattern.

Now raises `AmbiguousCanonical` naming the project and competing labels; the command prints them and exits 2. Same posture as #277 and #299 — refuse rather than guess.

Writing the test for that exposed a second thing: `concat_dir` was a **default argument**, so `CONCAT_DIR` was frozen at import and the corpus root was unpatchable. The function was testable only against the real corpus — which has exactly one mark per project, so it could never have shown the bug. Resolved at call time now.

Left open in #308: whether `select --execute` should clear a prior mark. Worth deciding before the v3 re-select.

**1109 passed, 3 skipped.**

Issue #308, opened 2026-08-04
canonical_runs silently drops a second canonical mark, which the v3 run will create


realmarcin:
From the review of #307.

`canonical_runs()` builds `out[project] = …` while walking provenance, so when a project carries a `canonical` block under more than one config, **the last one seen wins and the other disappears without a word**:

```
two configs, both marked canonical for CHORUS
  projects returned: 1
  CHORUS resolved to: 2026-09-09_config-b_rep1
  the other mark is silently dropped — no warning, no error
```

Ordering is by sorted path, so which one survives is a property of the label string rather than of anything meaningful.

## This is not hypothetical — it is the next step

The plan in NEXT_TASKS §9 is: run v3 → generate `VOICE_PEDIATRIC`, then VOICE → **re-select canonical across all five projects**. `d4d runs select --execute` writes a `canonical` block into the winner's provenance and does not clear any previous one, so after that step AI_READI, CHORUS and CM4AI each carry two marks — one under generic-v2, one under generic-v3.

At which point `d4d runs canonical --paths-only` returns one of them, chosen by string sort, and an evaluation sweep consumes whichever that is. Given v3 labels sort after v2 ones the answer would happen to be right, which is worse than it being wrong — it would work by accident until a label broke the pattern.

## Fix

Refuse rather than pick, the way `load_form_failures` refuses a cache spanning two rubrics (#277) and `_enum_aliases` drops a slot ranged on two enums (#299). Raise naming the project and the competing labels, and point at `--config` as the disambiguator — it already exists and already narrows correctly.

Worth deciding at the same time whether `select --execute` should clear a prior mark for the same project. Keeping both is defensible — the v2 choice remains a fact about the v2 config — but then the resolver needs a config to be meaningful, and callers should be made to say which.

realmarcin:
Fixed in the same PR (#307). `canonical_runs` raises `AmbiguousCanonical` naming the project and the competing labels; the command prints them and exits 2. `--config` disambiguates, and it is named in the message because nobody can pass it without knowing what to pass.

Writing the test exposed a second problem: `concat_dir` was a default argument, so `CONCAT_DIR` was frozen at import and the corpus root was unpatchable. The function was only testable against the real corpus — which has exactly one mark per project, so it could never have exhibited the bug. Resolved at call time now.

[...]

PR #309, opened 2026-08-04
Re-selecting supersedes the prior canonical mark, in both records (#308)

realmarcin:
Closes the half of #308 left open: `select --execute` now clears a prior mark.

## Why it matters now

NEXT_TASKS §9 re-selects canonical after the v3 run. Without this, AI_READI, CHORUS and CM4AI would each carry two marks — one under generic-v2, one under generic-v3 — and `canonical_runs` would refuse to resolve them.

## Cleared, and named

The winner's block gains `supersedes: [label]`. The superseded record gains `canonical_superseded_by` with the label, a timestamp and the tool.

A mark that simply vanishes leaves no trace of what the project used to ship. **"This replaced that" is the fact a reader of either record wants** — from whichever end they arrive at it.

## Verified against the actual scenario

A fixture reproducing the v3 case: CHORUS's generic-v2 replicates cloned under a generic-v3 label so both carry the mark.

```
before: AMBIGUOUS — more than one canonical record per project (CHORUS: …v2_rep2, …v3_rep2)
→ 2026-09-01_claude-opus-5-generic-v3_rep2  (50 slots)
  cleared the prior mark on 2026-07-31_claude-opus-5-generic-v2_rep2
after:  unambiguous — CHORUS -> …generic-v3_rep2
        supersedes: ['2026-07-31_claude-opus-5-generic-v2_rep2']
old record: has canonical=False  superseded_by=…generic-v3_rep2
```

The real corpus is untouched — still three canonical records, six paths.

**1112 passed, 3 skipped** (+3).

### DIFF
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -1,9 +1,9 @@
 """Run-tracking commands for the D4D CLI."""
 
-import click
-
 from datetime import datetime, timezone
 from pathlib import Path
+
+import click
 
 
 @click.group()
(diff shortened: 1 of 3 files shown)


realmarcin:
## Review

Two findings, #310–#311, both fixed in f8a80458.

**#310 — the clear could silently fail to clear.** The scan swallowed every error, so a provenance file that would not parse kept its `canonical` block and `select --execute` reported nothing:

```
select output: (no clearing reported)
corrupt file still carries a canonical mark: True
```

The contract is one canonical record per project; a file it cannot read is a file whose mark it cannot clear. And the failure did not surface *here* — it surfaced later in `canonical_runs` as `AmbiguousCanonical`, in a different command, with nothing connecting it back to the selection meant to resolve it. **That is the worst possible placement for the error.** Now fails the selection, naming the file and what goes wrong if it proceeds.

**#311 — pre-existing, surfaced by the test for #310.** `_prov` was annotated `dict | None` and returned whatever `safe_load` produced, so a list-shaped record reached callers calling `.get` on it: a bare `AttributeError` pointing at `runs.py` rather than at the file. `validation_status` had the same shape. Both now treat a non-mapping as unusable — which is what `check_provenance` already did for its own path.

Also checked: rewriting a provenance file through `ProvenanceRecord` round-trips **byte-identically**, so clearing a mark produces a minimal diff rather than reformatting the record.

Left open, noted in #310: the clear is not transactional — the winner is written before the priors are cleared, so a process dying between leaves two marks. Recoverable by re-running `select`, but the window exists.

**1114 passed, 3 skipped.**

PR #312, opened 2026-08-04
There is no ordering constraint between VOICE and VOICE_PEDIATRIC

realmarcin:
Corrects a claim I introduced in §9 and repeated twice.

## What it said

> **Order matters: the pediatric record has to exist before VOICE can point at it.** Generate `VOICE_PEDIATRIC` first, or a regenerated VOICE either inlines the dataset again and fails again, or omits the relationship entirely.

## Why it is wrong

`target_dataset` is `range: string` with `slot_uri: dcterms:relation` — an **identifier**, not a pointer to a record we hold. The evidence was already in the corpus:

| replicate | `target_dataset` | that slot |
|---|---|---|
| rep2 | `https://physionet.org/content/b2ai-voice-pediatric/1.1.0/` | **passed** |
| rep3 | `https://doi.org/10.13026/mf9s-5r03` | **passed** |

Both failed on `relationship_type` alone. And the pediatric DOI appears **11 times in VOICE's own bundle**, so VOICE can reference the dataset whether or not a pediatric datasheet exists.

The claim conflated two things: referencing a dataset **by identifier**, which needs only the DOI in the input documents, and referencing **a D4D record we produced**, which is not what this slot holds.

## What follows

**VOICE and `VOICE_PEDIATRIC` can be generated in either order, or independently.** The run sequence has one fewer constraint than I said.

The fork stands on its own reasoning — a dataset with its own DOI, protocol and ethics approval is its own datasheet (#292) — and never needed VOICE to depend on it. Asserting a dependency that does not exist would have constrained the run order for no reason, and made the fork look like a prerequisite rather than a modelling decision.

**Still true:** a v3 run is the first opportunity for VOICE to have a canonical record, because it is the first run under the corrected enum handling and the clarified slot description.

Docs-only. **1114 passed, 3 skipped.**

### DIFF
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -325,16 +325,26 @@ broke nothing; the four-project list in 26 files is prose, not logic.
   - the enum values: `References` normalises to `references` on the write path,
     and `related_to` still fails, which is correct — it names nothing in the
     vocabulary;
-  - the inline object in `target_dataset`: once the pediatric dataset **is** its
-    own record, VOICE references it by DOI, which is a string and validates.
-    Note the tense — it is a registered project with a bundle, not yet a
-    datasheet, and this bullet is what defers generating it. **Order matters:
-    the pediatric record has to exist before VOICE can point at it.** Generate
-    `VOICE_PEDIATRIC` first, or a regenerated VOICE either inlines the dataset
-    again and fails again, or omits the relationship entirely.
+  - the inline object in `target_dataset`: the slot is `range: string` with
+    `slot_uri: dcterms:relation` — an **identifier**, not a pointer to a record
+    we hold. rep2 and rep3 already put a URL there and that slot passed; they
+    failed on `relationship_type` alone. The pediatric DOI appears 11 times in
+    VOICE's own bundle, so VOICE can reference it whether or not a pediatric
+    datasheet exists.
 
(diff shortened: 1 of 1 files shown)


realmarcin:
## Review

One finding, #313, fixed in 6762369f — **the sentence I wrote while removing one false dependency introduced another.**

The correction ended *"a v3 run is the first opportunity for VOICE to have a canonical record"*. Nothing about the enum normaliser or the slot description is tied to v3: the first is on the write path, the second is in the schema, so any new run picks both up. A v2 re-run would serve equally. v3 is what is planned, not what is required.

And "opportunity" overstated it. Of VOICE's three failures:

| failure | fixed for a new run? |
|---|---|
| `References` (casing) | yes — normalised on the write path |
| inline object in `target_dataset` | probably — the description no longer promises "or Dataset object" |
| `related_to` | **no** — names nothing in the vocabulary, nothing stops it recurring |

**Verified while reviewing, and it also confirms this PR's premise:** correcting rep2's `related_to` to a permitted value makes that record validate **completely**. So the string `target_dataset` was never a blocker — the enum was the whole of it — which is exactly what the PR claims.

§9 now says chance rather than guarantee and names which failure is the risk.

Also checked: the ordering claim survives nowhere else in the repo.

Docs-only. **1114 passed, 3 skipped.**

PR #316, opened 2026-08-05
Score the rubric the repository publishes (#314)

### LINKED ISSUE #314: rubric20-semantic scores 5 questions the rubric does not define, including a Q20 that is a different question
Found by an external Codex review of rerun readiness (`notes/codex_rerun_readiness_review_2026-08-04.md`), verified here. This is a validity problem, not a scaling one, and it is on the path that matters for the paper.

## The semantic agent and the rubric disagree on 5 of 20 questions

Comparing `.claude/agents/d4d-rubric20-semantic.md` against the canonical `data/rubric/rubric20.txt`:

| | agent scores | rubric defines |
|---|---|---|
| Q10 | Interoperability and Standardization | Interoperability, Standardization, and Cross… |
| Q13 | Version History Documentation | Version History, Maintenance, and Sustainabi… |
| Q18 | Reusability (License Clarity) | Reusability, Use Guidance, and Social Impact |
| Q19 | Data Integrity and Provenance | Data Integrity, Provenance Graph, and Qualit… |
| **Q20** | **Interlinking Across Platforms** | **Bias Documentation and Responsible AI Alignment** |

Q10, Q13, Q18 and Q19 look like the rubric broadening scope while the agent kept the older, narrower question. **Q20 is not a rename at all** — the agent scores a retired interlinking question, and the rubric's bias / responsible-AI question **is not scored by the semantic path at all**.

So a semantic evaluation does not measure what the rubric says it measures, and the omission is specifically the RAI content.

## The denominator is enforced, not merely stale

`src/download/prompts/rubric20_semantic_schema.json` has:

```
const 84 at .properties.overall_score.properties.max_points
```

It is a `const`, so the schema **rejects** an output reporting 88. Correcting the agent without correcting this would make every semantic evaluation fail validation.

Four files carry 84 and none carry 88:

| file | occurrences of 84 |
|---|---|
| `.claude/agents/d4d-rubric20-semantic.md` | 5 (incl. `max_points`, `adjusted_max_points`, `max_score`) |
| `src/download/prompts/rubric20_semantic_schema.json` | 1 (`const`) |
| `src/download/prompts/rubric20_output_format.json` | 1 |
| `src/data_sheets_schema/schema/D4D_Evaluation_Summary.yaml` | 1 |

The canonical rubric is **17 numeric + 3 pass/fail = 88**.

## Why #275 missed it

#275 corrected the scoring path — constants, the summariser scripts, the non-semantic system prompt — and its tests assert against those. Nothing compared the *semantic agents* to `data/rubric/rubric20.txt`, so the semantic path kept both the old denominator and the old question set.

## Fix

1. Align the agent's 20 questions with the rubric, Q20 especially.
2. Change the `const: 84` to the computed maximum, and the three other files with it.
3. Add a test comparing the semantic agents' question numbers, names and score types **directly against `data/rubric/rubric20.txt`**, so the two cannot drift again. That is the check whose absence let this survive.

## Blocking

The semantic evaluation is going in the paper, so this blocks the rerun's evaluation step. Running it as-is produces numbers against a 20-question instrument that differs from the published rubric in a quarter of its questions.

## Not checked

`rubric10-semantic` could not be compared the same way — `data/rubric/rubric10.txt` nests under `d4d_complex_proxy_rubric` with a different shape, so the element names need a separate comparison. Its denominator is right (50 everywhere), but that is not evidence its questions match.

realmarcin:
Closes #314.

## What was wrong

Three rubric20 scorers each carry their own copy of the 20 questions. All three
had drifted from `data/rubric/rubric20.txt`:

| scorer | questions wrong | denominator |
|---|---|---|
| `.claude/agents/d4d-rubric20-semantic.md` | 5 | 84 |
| `.claude/agents/d4d-rubric20.md` | 6 | 84 |
| `scripts/batch_evaluate_rubric20_hybrid.py` | 5 | table summed to 84, divided by 88 |

Four drifted questions were the rubric broadening scope while the scorers kept
the narrower name (Q9, Q10, Q13, Q18, Q19 vary by file). The fifth was not a
rename: **all three scored a retired "Interlinking Across Platforms" as
pass/fail worth 1** where the rubric's Q20 has been "Bias Documentation and
Responsible AI Alignment", numeric worth 5, since `schema_version: 2.0`.

So the responsible-AI question was not scored by any path, and `84` against
`88` is exactly that one substitution — 17 numeric x 5 + 3 pass/fail x 1.

`src/download/prompts/rubric20_system_prompt.md` is the shape that did not
drift: it interpolates `{RUBRIC_SPECIFICATION}` from the rubric file at run
time. Untouched here, and the reason its questions could not go stale.

## The hybrid scorer, which the last fix made worse

#275 corrected `RUBRIC20_MAX_SCORE` to 88, and
`batch_evaluate_rubric20_hybrid.py`'s `max_total = RUBRIC20_MAX_SCORE` followed
it. Its question table still summed to 84. A record scoring every available
point was therefore reported as **95.5%**, and no percentage the script emitted
could reach 100.

`max_total` is now `derived_maximum()` — computed from the table, raising when
it disagrees with the constant, called at import so a drifted table stops the
run rather than surfacing partway through a batch.

## Also corrected

A judge shown a worked example that contradicts the specification follows the
example.

- **Every worked example.** The agents' category maxima decomposed to
  24/22/25/13 and `rubric20_output_format.json` still demonstrated Q20 as
  pass/fail. Categories are 21/21/25/21; every derived percentage recomputed.
- **`rubric20_semantic_schema.json`** pinned `max_points` with `const: 84`, so
  it would have *rejected* an evaluation reporting the right maximum.
- **Q20's applicability gate.** It sat behind "datasets shared & available for
  reuse", which made sense for cross-platform links and does not for bias: a
  dataset that is never released still has documentable biases, and the rubric
  applies Q20 to all four programs. Q10 and Q13 keep the field-based gates the
  rewrite dropped.
- **A Q20 scorer in the hybrid path.** Reads `bias_type` structurally against
  `BiasTypeEnum` rather than grepping for the word "bias", so prose naming a
  selection bias scores 3 and a categorised bias with a mitigation strategy or
  affected subsets scores 5 — the rubric's own distinction.

## Testing

29 tests in two files, comparing each scorer to `rubric20.txt` rather than to
each other. `1143 passed, 3 skipped`.

Mutation-checked: 19 reintroductions of the defects above, each confirmed to
fail. One found a hole in the test itself — `**Maximum Possible Score:** 88` was
invisible to a separator class that did not admit markdown bold markers, so
reverting the headline figure to 84 left every test green. Fixed and re-checked.

## Not in scope, will be filed

The field lists and scoring-band wording have drifted much further than the
names — the hybrid scorer and both agents disagree with the rubric's `field:`
on most questions. That needs judgement per question (some are path spellings
of the same slot, some are real substitutions) and is not mechanical.

## Second commit

`notes/codex_rerun_readiness_review_2026-08-04.md`, written 2026-08-04 and left
uncommitted. Unrelated to #314; carried here so it stops living only in a
working tree.

### DIFF
--- a/.claude/agents/d4d-rubric20-semantic.md
+++ b/.claude/agents/d4d-rubric20-semantic.md
@@ -90,7 +90,7 @@ Read the provided D4D YAML file and perform a **semantic quality assessment** th
        | Condition | Satisfied when… | Gates |
        |---|---|---|
        | Human subjects | `description` or `keywords` reference human participants, patients, or clinical research, OR `collection_mechanisms` describes human participant recruitment — checked via Q1/Q2/Q3/Q12 fields only, never Q8's own fields | Q8, Q15 |
-       | Datasets shared & available for reuse | `distribution_formats` populated OR `download_url`/`page` links to accessible data OR license explicitly permits reuse | Q10, Q17, Q20 |
+       | Datasets shared & available for reuse | `distribution_formats` populated OR `download_url`/`page` links to accessible data OR license explicitly permits reuse | Q10, Q17 |
        | Software tools produced as dataset output | `external_resources` references a code repository, OR `description`/`purposes` explicitly identifies software production as a dataset output — checked via Q2/Q7/Q14 fields only, never Q11's own fields | Q11 |
        | Data collection identified AND datasets shared | Collection fields populated (`acquisition_methods`, `collection_mechanisms`, `data_collectors`, or `collection_timeframes`) AND the datasets shared condition above is met | Q12, Q13 |
        | Publication identified AND datasets shared | `citation` or `external_resources` includes at least one publication reference AND the datasets shared condition above is met | Q14 |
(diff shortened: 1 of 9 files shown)


realmarcin:
## Review pass

Six findings, filed as #317–#322. Four addressed in `8d77ae0c`; two left filed
because they need per-question judgement against the schema rather than a
mechanical edit.

**Fixed here**

- **#318** — `typed` and `fairness` were computed independently over the whole
  record, so nine untyped biases beside one typed one scored 5, and so did a
  record where one bias supplied the taxonomy and a different one supplied the
  mitigation. Fairness is now required *of the categorised biases*, and
  "comprehensive" means every documented bias categorised.
- **#317** — `BIAS_TYPE_ENUM` was a second copy of the schema's vocabulary with
  nothing keeping it in step. Pinned by a test against `BiasTypeEnum`.
- **#321** — rubric10 now has the same parity comparison. Green on the first
  run; that is the point.
- **#322** — one rule for the maximum, and the summary schema is read as YAML.

**Left filed**

- **#319** — the scorers' `field:` lists and scoring-band wording drift much
  further than the names. Three different causes (path spellings, superseded
  slot names, genuine substitutions) needing different answers. The superseded
  slot names are the ones that silently cost points and are worth resolving
  before the rerun.
- **#320** — `same_as` and `project_website` are no longer read by any hybrid
  question. Follows from the rubric, but it is a silent coverage change.

**Mutation-checking the review fixes found three more holes, all in the tests**

- The maximum assertions rejected `84` specifically and accepted anything else
  — `50 → 45` in both rubric10 agents passed clean.
- No assertion had ever matched `max 88 points` (only `maximum`), which is how
  `D4D_Evaluation_Summary.yaml` writes it.
- The summary-schema check filtered for lines containing `rubric20`, and the
  maximum is on the `description:` line *below* that key — so it had never
  examined the value it existed to examine.

23 mutants across the two test files, no survivors. `1150 passed, 3 skipped`.

Issue #323, opened 2026-08-05
Every recorded semantic evaluation fails its own JSON schema


realmarcin:
Found while working #319.

`scripts/validate_evaluation_schema.py` validates the recorded semantic
evaluations against `rubric10_semantic_schema.json` and
`rubric20_semantic_schema.json`. Run today:

```
SUMMARY:
  ✅ Valid:   0
  ❌ Invalid: 28
  📊 Total:   28
```

Not one passes. Four distinct causes:

```
8x  'percentage' is a required property        (rubric20-semantic)
8x  'summary_scores' is a required property    (rubric10-semantic)
4x  'element_scores' is a required property    (rubric10-semantic)
3x  'description' is a required property
```

**The rubric20 case is the clearest.** `rubric20_semantic_schema.json` declares
`overall_score` as exactly `{total_points, max_points, percentage}` and requires
all three. The agent instructs emitting the N/A convention's shape instead:

```
total_points, max_points, excluded_max_points, adjusted_max_points,
normalized_percentage, questions_not_applicable
```

`percentage` is absent, and required. All 24 rubric20-semantic records on disk
have the agent's shape, so the schema has been rejecting every one of them since
the N/A convention landed. The schema was never updated with it.

**This is on the critical path for the rerun.** Twelve semantic evaluations are
planned; as things stand the validator rejects all twelve, and there is no
in-repo signal short of running the validator by hand — it is not wired into
`make test` or CI.

**Deciding what to fix requires a call I did not want to make silently:**

1. Is `normalized_percentage` the reported figure and `percentage` retired? The
   agent says the normalized one "is the only percentage reported", which reads
   like yes — in which case the schema should require `normalized_percentage`
   and drop `percentage`.
2. Or does `percentage` survive as the un-adjusted `total_points / max_points`,
   in which case the agent has to emit both.

The same question applies to rubric10's `summary_scores`/`element_scores`.

Related: #316 corrected this schema's `const: 84`, which is how the file came up
for review at all. `tests/test_evaluation/test_rubric20_fields_resolve.py`
carries `UNDECLARED_OUTPUT_KEYS` listing the N/A vocabulary explicitly, with a
pointer here, rather than widening its exclusion silently.

**Worth wiring the validator into the test suite once the contract is settled** —
a schema nothing runs is how this survived 28 files.

PR #324, opened 2026-08-05
Score fields the schema actually has (#319)

### LINKED ISSUE #319: rubric20 scorers disagree with the rubric's field lists and scoring bands, not just its names
Found reviewing #316, and deliberately left out of it.

#316 aligned the *names*, *score types* and *denominator* of the three rubric20
scorers with `data/rubric/rubric20.txt`. Their `field:` lists and scoring-band
wording are still their own. Measured against the rubric:

**`scripts/batch_evaluate_rubric20_hybrid.py`** and **`.claude/agents/d4d-rubric20.md`**
disagree on nearly every question. Examples from the plain agent:

```
Q1  missing: confidentiality_level, creators, doi, instances, page,
             parent_datasets, purposes, resources, variables
Q8  missing: data_protection_impacts, ethical_reviews, human_subject_research,
             informed_consent, is_deidentified, participant_compensation,
             participant_privacy, reidentification_risk, vulnerable_populations
    extra:   deidentification_and_privacy, ethics
Q12 missing: acquisition_methods, collection_mechanisms, collection_timeframes,
             data_collectors, raw_data_sources
    extra:   collection_process
```

**`.claude/agents/d4d-rubric20-semantic.md`** drifts on about ten questions,
more mildly:

```
Q8  missing: reidentification_risk, vulnerable_populations
    extra:   at_risk_populations, participant_privacy.reidentification_risk
Q15 missing: is_subpopulation, missing_data_documentation, vulnerable_populations
    extra:   DataSubset.is_data_split, DataSubset.is_subpopulation
Q16 missing: doi, id      extra: download_url, external_resources
```

**Why this is not mechanical.** The three categories need different answers:

1. *Path spellings of the same slot* — `participant_privacy.reidentification_risk`
   against `reidentification_risk`. The scorer's form is arguably more precise.
2. *Superseded slot names* — `deidentification_and_privacy`, `collection_process`,
   `data_characteristics.data_formats`. Need checking against the current schema;
   a field the schema no longer has scores 0 forever.
3. *Genuine substitutions* — `at_risk_populations` for `vulnerable_populations`.
   One of the two is wrong and the schema decides which.

Category 2 is the one that silently costs points, and it is worth resolving
before the rerun; 1 and 3 can follow.

**Also**: verify every field named by any scorer exists as a `Dataset` slot,
the same premise check `test_the_fields_the_prompt_may_use_actually_exist` does
for the grounding fields in `tests/test_evaluation/test_semantic_agent_grounding.py`.
A scorer looking for a slot that does not exist reports a 0 that means nothing.

realmarcin:
Closes #319. **Stacked on #316** — will retarget to `main` before that merges.

## What was wrong

#316 aligned the three rubric20 scorers' question *names* with
`data/rubric/rubric20.txt`. Their `field:` lists were still written against a
schema the project no longer has. Of the 93 distinct paths named by the rubric
or a scorer, **29 could not resolve from `Dataset` at all**.

A field that does not resolve does not fail. It reads as absent, forever.

## What it cost, measured

Two questions were unscoreable on every record ever produced. Q10 carried a
second, independent defect: `formats_str = str(has_formats)` stringified a
*bool*, so no format name was ever a substring of `"True"` — its 5 points were
unreachable even where the fields resolved.

Six of 88 points could not be earned by any record, and nothing reported an error.

Across the 25 current records, before → after:

| | before | after |
|---|---|---|
| Q5 Data File Size Availability | **0.0/1 constant** | 0.2/1 |
| Q10 Interoperability… | **0.0/5 constant** | 4.8/5 |
| Q8 Ethical and Privacy Declarations | 4.8/5 | **4.4/5** |
| Q11 Tool and Software Transparency | 1.6/5 | 2.0/5 |
| Q15 Human Subject Representation | 4.1/5 | 4.8/5 |
| **total** | 62.7/88 | 68.9/88 |
| sd | 10.53 | 11.04 |

**Q8 going down is the point.** `has_ethical` rested on
`get_nested_value(data, 'ethics.ethical_position')` — never resolvable — *or*
the substring `"ethic"` appearing anywhere in `purposes`. With the dead half
inert, the live half was a word search, and it was passing records documenting
no ethics fields at all.

## The fix

`rubric20.txt` first, because `rubric20_system_prompt.md` interpolates it as
`{RUBRIC_SPECIFICATION}` — a dead field there reaches the API judge directly.
The two agents and the hybrid scorer are then brought up to it, keeping their
own additions where those resolve.

Two kinds of replacement, kept distinct:

- **A leaf that only needed its path.** `confidentiality_level` is a real field,
  on `ExportControlRegulatoryRestrictions`, reachable as
  `regulatory_restrictions.confidentiality_level`. Same for `hipaa_compliant`,
  `other_compliance`, `governance_committee_contact`, `reidentification_risk`,
  `is_data_split`, `is_subpopulation`, `bytes`.
- **A slot the schema no longer has**, replaced by current equivalents:
  `ethics`, `deidentification_and_privacy`, `collection_process`,
  `data_characteristics`, `access_and_licensing`, `software_and_tools`,
  `funding_and_acknowledgements`, `release_notes`, `files`, `motivation`,
  `citations`/`references`, `composition.population`. `rrid` is dropped — not
  in the schema at all, and `doi`/`id`/`page` already carry that check.

`vulnerable_populations` resolved the other way: the semantic agent's
`at_risk_populations` was right and the rubric was wrong. The rubric moved.

## The distinction the tests rest on

**Q9 still scores 3 on all 25 records, and now that is a true statement.**
`access_and_licensing.access_policy` never resolved, so band 5 was unreachable
by construction. It now reads `regulatory_restrictions`, which every record
leaves empty — a real gap in generation rather than an artefact of the scorer.

So the tests assert **resolution, never population**. Requiring every named
field to appear in some record would forbid the rubric from asking for anything
the generator does not yet produce, which is most of what a rubric is for.

## Testing

15 tests. Mutation-checked with 11 reintroductions, no survivors. One found a
gap the Fields-line tests could not see: the agents also name record fields in
their cross-field consistency rules ("IF `human_subject_research.…` THEN EXPECT
`ethical_reviews` populated"), where a dead name instructs the judge to compare
against something that cannot exist. Now covered.

`1165 passed, 3 skipped`.

## Found on the way, filed not fixed

**#323 — every recorded semantic evaluation fails its own JSON schema.**
`scripts/validate_evaluation_schema.py` reports `Valid: 0, Invalid: 28`. The
rubric20 schema requires `overall_score.percentage`; the agent emits
`normalized_percentage` and the N/A vocabulary instead. This is on the critical
path for the rerun's twelve semantic evaluations, and fixing it means deciding
which percentage is the reported one — a call I did not want to make silently.

### DIFF
--- a/.claude/agents/d4d-rubric20-semantic.md
+++ b/.claude/agents/d4d-rubric20-semantic.md
@@ -119,7 +119,7 @@ Read the provided D4D YAML file and perform a **semantic quality assessment** th
 #### Question 1: Field Completeness
 **Description:** Proportion of mandatory schema fields populated (id, title, description, keywords, license).
 
-**Fields:** `id`, `title`, `description`, `keywords`, `license_and_use_terms`
+**Fields:** `id`, `title`, `description`, `keywords`, `license_and_use_terms`, `doi`, `page`, `creators`, `purposes`, `instances`, `resources`, `parent_datasets`, `variables`, `regulatory_restrictions.confidentiality_level`
 
 **Scoring (numeric 0-5):**
 - **0:** ≤40% fields populated
@@ -161,7 +161,7 @@ Read the provided D4D YAML file and perform a **semantic quality assessment** th
 #### Question 4: File Enumeration and Type Variety
 **Description:** Number of distribution formats and file type diversity.
 
-**Fields:** `distribution_formats`, `format`, `media_type`
+**Fields:** `distribution_formats`, `conforms_to_schema`, `total_size_bytes`, `file_collections.total_bytes`
 
 **Scoring (numeric 0-5):**
 - **0:** 1 file type only
@@ -175,7 +175,7 @@ Read the provided D4D YAML file and perform a **semantic quality assessment** th
 #### Question 5: Data File Size Availability
 **Description:** Presence of file size or instance count metadata.
 
-**Fields:** `bytes`, `instances`
+**Fields:** `total_size_bytes`, `file_collections.total_bytes`, `instances`, `subsets.is_data_split`, `splits`, `subsets.is_subpopulation`, `subpopulations`
 
 **Scoring (pass/fail):**
 - **Pass:** Numeric file size or instance count found
@@ -190,7 +190,7 @@ Read the provided D4D YAML file and perform a **semantic quality assessment** th
 #### Question 6: Dataset Identification Metadata
 **Description:** Presence of unique identifiers such as DOI, RRID, or persistent URLs.
 
-**Fields:** `doi`, `rrid`, `page`
+**Fields:** `doi`, `page`, `id`, `publisher`
 
 **Scoring (pass/fail):**
 - **Pass:** At least one persistent ID found
@@ -233,7 +233,7 @@ Read the provided D4D YAML file and perform a **semantic quality assessment** th
 #### Question 9: Access Requirements and Governance Documentation
 **Description:** Whether access policy, license, IP restrictions, regulatory restrictions, and confidentiality level are clearly defined.
 
-**Fields:** `license_and_use_terms`, `ip_restrictions`, `regulatory_restrictions`, `confidentiality_level`, `data_protection_impacts`, `regulatory_restrictions.governance_committee_contact`
+**Fields:** `license_and_use_terms`, `ip_restrictions`, `regulatory_restrictions`, `regulatory_restrictions.confidentiality_level`, `data_protection_impacts`, `regulatory_restrictions.governance_committee_contact`, `regulatory_restrictions.hipaa_compliant`, `regulatory_restrictions.other_compliance`
 
 **Scoring (numeric 0-5):**
 - **0:** No license or access info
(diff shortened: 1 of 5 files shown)


realmarcin:
## Review pass

Three findings. One was a defect in this PR and is fixed in `2a5f15ae`; two are
filed.

**Fixed here — #326: the remap was global, and the right replacement is not.**

`format` in Q10 (Interoperability, Standardization) means schema conformance.
In Q4 — "File Enumeration and Type Variety", scored `0: 1 file type only`,
`3: 2–3 file types`, `5: >3 file types` — it means the *kinds* of file. One map
gave both the same answer, so Q4 ended up with `conforms_to_schema` (a schema
URL, silent on how many) and two byte-size fields, while its own `method:` still
read "Count total formats and unique media types."

Every path resolved. Every test passed. The question had stopped measuring what
it says it measures — the failure mode this PR is about, reintroduced by the fix
for it. Q4 is now `["distribution_formats", "file_collections",
"total_file_count"]`. Scores are unchanged (2.4/5, still discriminating): the
list was wrong, not the arithmetic.

`TestFieldsMatchWhatTheQuestionMeasures` pins Q4 and Q5 in both directions — Q4
may not score sizes, Q5 must — so the rule cannot be satisfied by emptying both.

**Filed — #325: six questions score identically on all 25 records.**

Q9 is the one that matters. It is pinned at 3/5 because band 5 needs
confidentiality classification and a governance contact, and all four
`regulatory_restrictions.*` sub-fields are empty in **0/25** records. Before this
PR that was invisible — Q9 read `access_and_licensing.access_policy`, not a slot,
so band 5 was unreachable by construction. The constant is now a true measurement
of a real generation gap, and it is cheap to address before the rerun.

The other five (Q2, Q6, Q7, Q12, Q16 — 11 of 88 points) are saturated at the top:
awarded to every record unconditionally. Not wrong, but they inflate the total
without separating methods.

**Filed — #323: every recorded semantic evaluation fails its own JSON schema.**

`scripts/validate_evaluation_schema.py` reports `Valid: 0, Invalid: 28`. On the
critical path for the rerun's twelve semantic evaluations. Fixing it means
deciding which percentage is the reported one, which I did not want to settle
silently.

**On the evidence standard.** The resolution test is mechanical and complete. The
27 replacements it does not judge were checked by hand against each question's
`method:` and scoring bands — that is a weaker standard than the test around it,
and #326 records why.

`1168 passed, 3 skipped`. 13 mutants across the new file, no survivors.

Issue #317, opened 2026-08-05
rubric20 hybrid scorer keeps its own copy of BiasTypeEnum


realmarcin:
Found reviewing #316.

`scripts/batch_evaluate_rubric20_hybrid.py` hardcodes the enum that decides
Q20's 3-versus-5 band:

```python
BIAS_TYPE_ENUM = frozenset({
    'selection_bias', 'measurement_bias', 'historical_bias', ...
})
```

The schema declares the same nine values in
`src/data_sheets_schema/schema/D4D_Base_import.yaml` under `BiasTypeEnum`.
They agree today. Nothing keeps them agreeing.

**Failure**: `BiasTypeEnum` gains a tenth value — say `temporal_bias`. A record
that categorises a bias as `temporal_bias` with a mitigation strategy is a
5-point record by the rubric, and the scorer awards 3, reporting "documented
without taxonomy categorisation". The record is penalised for using the
vocabulary the schema told it to use. Silent: no error, just a lower score.

This is the shape of #292/#299 — a second copy of a vocabulary is one to keep
in step, and `tests/test_enum_alias_normalisation.py` has
`test_it_is_read_from_the_schema_not_hardcoded` for exactly this reason.

**Fix**: read the permissible values via `SchemaView`, as `_enum_aliases()`
does. If the import cost at module scope is unwanted, keep the literal and add
a test asserting it equals the schema's set — the point is that drift fails
rather than silently rescoring.

realmarcin:
Fixed in #316 (commit 8d77ae0c).

PR #327, opened 2026-08-05
Describe the output the agents actually instruct (#323)

### LINKED ISSUE #323: Every recorded semantic evaluation fails its own JSON schema

realmarcin:
Closes #323.

## What was wrong

`scripts/validate_evaluation_schema.py` reported `Valid: 0, Invalid: 28`. Not one
recorded semantic evaluation conformed to its own JSON schema. Nothing runs the
validator, so nothing said so — and the rerun's twelve semantic evaluations would
have been rejected the same way.

Six independent drifts, all the same shape: the schemas describe the 2025-12
output and the agents moved on.

| | drift |
|---|---|
| rubric20 | `overall_score` required `percentage`; the agent emits the N/A vocabulary and says the normalized one is *the only percentage reported* |
| rubric10 | required `summary_scores`/`element_scores`; the agent has instructed `overall_score`/`elements` since 2026-07 |
| both | `score` typed `number`, so `score: null` — "not assessed" — failed. Already on disk 6 times. |
| rubric10 | `semantic_insights` typed `string` against an array; `severity` enumerated `critical\|warning\|info` against the agent's `low\|medium\|high`; `consistency_checks` typed as arrays against three counters; `grant_format` against `grant_number_format` |
| rubric10 | `method` enumerated a hand-written subset that rejected `claudecode_agent_core`, a method the pipeline has produced for months |
| rubric10 | `element_max` fixed at `const: 5`, contradicting the N/A convention — records already carry `element_max: 3` |

## Decisions I made rather than leaving open

**`normalized_percentage` is the reported figure.** The agent says so
explicitly ("This is the only percentage reported; it is comparable across
datasheets regardless of how many questions are excluded"), and all 24 records
emit it. `percentage` is gone rather than kept as an alias — two percentages
with different denominators is how this kind of thing goes wrong again.

**Two vocabularies dropped rather than widened.** The issue `type` and
rubric10's `correctness_validations` enumerated values the agents never
documented and records outgrew (`content_accuracy`, `not_present`,
`mostly_valid`). rubric20 had already taken this line for the latter. A schema
that rejects real work over a taxonomy nobody maintains is worse than no
constraint; the vocabulary is now documented in the description.

**`method` mirrors `constants.METHODS`**, pinned by a test — the same shape as
#317's `BiasTypeEnum`.

## Twenty records still fail, and should

    ✅ Valid:      8
    🕐 Superseded: 12   pre-2026 rubric10 shape; nothing to be done to them
    ❌ Invalid:     8   `max_points: 84`, a maximum the rubric never had (#314)

`Invalid: 28` said nothing about which records were merely old. Weakening the
schema to admit the 84-denominator records would be the opposite of the fix, so
`classify()` separates a superseded *shape* from a wrong *value*, and only the
second fails the run.

## Testing

13 tests. The load-bearing one builds a record from `rubric20.txt` — all 20
questions in the four categories the schema requires, one carrying the N/A
convention — and asserts it validates. That is what the rerun will produce, and
it is the single test that would have failed for all six drifts at once.

Its first version used one category: it passed every check that mattered while
failing `minItems: 4`, which was the schema being right and the fixture being
lazy.

Mutation-checked with 12 reintroductions, no survivors. `1181 passed, 3 skipped`.

### DIFF
--- a/scripts/render_evaluation_html_rubric10_semantic.py
+++ b/scripts/render_evaluation_html_rubric10_semantic.py
@@ -5,6 +5,7 @@ Render rubric10-semantic evaluation JSON files to HTML
 import json
 from pathlib import Path
 from datetime import datetime
+from data_sheets_schema.rubric_pooling import reported_percentage
 
 
 def generate_evaluation_html(eval_data, output_path):
@@ -22,7 +23,7 @@ def generate_evaluation_html(eval_data, output_path):
     summary = eval_data.get("summary_scores") or {
         "total_score": eval_data.get("overall_score", {}).get("total_points", 0),
         "total_max_score": eval_data.get("overall_score", {}).get("max_points", 50),
-        "overall_percentage": eval_data.get("overall_score", {}).get("percentage", 0),
+        "overall_percentage": reported_percentage(eval_data),
     }
     semantic = eval_data.get("semantic_analysis", {})
     assessment = eval_data.get("assessment", {})
--- a/scripts/render_evaluation_html_rubric20_semantic.py
+++ b/scripts/render_evaluation_html_rubric20_semantic.py
@@ -17,6 +17,7 @@ import json
 from pathlib import Path
 from datetime import datetime
 from data_sheets_schema.constants import RUBRIC20_MAX_SCORE
+from data_sheets_schema.rubric_pooling import reported_percentage
 
 
 def generate_evaluation_html(eval_data, output_path):
@@ -34,7 +35,7 @@ def generate_evaluation_html(eval_data, output_path):
     overall = eval_data.get("overall_score", {})
     total_score = overall.get("total_points", 0)
     max_score = overall.get("max_points", RUBRIC20_MAX_SCORE)
-    percentage = overall.get("percentage", 0)
+    percentage = reported_percentage({'overall_score': overall})
 
     # Calculate grade
     if percentage >= 95:
--- a/scripts/summarize_rubric10_results.py
+++ b/scripts/summarize_rubric10_results.py
@@ -18,6 +18,7 @@ import csv
 from pathlib import Path
 from typing import List, Dict
 from datetime import datetime
+from data_sheets_schema.rubric_pooling import reported_percentage
 
 # Base directory
 BASE_DIR = Path(__file__).parent.parent
@@ -74,7 +75,7 @@ def create_csv_summary(results: List[Dict]):
             overall = result.get('overall_score', {})
             total_score = overall.get('total_points', 0)
             max_score = overall.get('max_points', 50)
-            percentage = overall.get('percentage', 0)
+            percentage = reported_percentage({'overall_score': overall})
 
             # Count elements passing (score >= 3/5)
             elements = result.get('elements', [])
@@ -134,7 +135,7 @@ def create_markdown_table(results: List[Dict]):
                 overall = result.get('overall_score', {})
                 total = overall.get('total_points', 0)
                 max_pts = overall.get('max_points', 50)
-                pct = overall.get('percentage', 0)
+                pct = reported_percentage({'overall_score': overall})
 
                 elements = result.get('elements', [])
                 elements_passing = sum(1 for el in elements if el.get('element_score', 0) >= 3)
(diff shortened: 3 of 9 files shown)


realmarcin:
## Review pass

Two findings. One was a defect this PR left in place and is fixed in `8f634528`;
one is filed.

**Fixed here — the contract fix was only half of it.**

The schemas described `overall_score.percentage`; so did every consumer:

```python
percentage = overall.get('percentage', 0)
```

The N/A convention renamed that key, and the default of `0` turned the absence
into a score:

    16 of 16 current-shape records reported 0.0%
    their real scores ran 72.0 – 98.8

Both summarisers, both HTML renderers. `render_evaluation_html_rubric20_semantic`
grades on that number, so every current record rendered as an **F at 0.0%**.

Silent, because `0` is a plausible reading of a bad evaluation — which is exactly
why it cannot double as "absent". `reported_percentage` raises instead, falls back
to `percentage` for pre-2026 records, and returns `0.0` when a record genuinely
reports zero. 29 call sites across four scripts now go through it.

**Filed — #328: `fix_evaluation_scores.py` writes the superseded shape.**

It computes a percentage against the *unadjusted* maximum and writes it under a
key the schema no longer declares, turning a conformant record non-conformant.
Left alone here because its `get(..., 0)` is load-bearing — it compares current
against recomputed to decide whether to rewrite — and changing that without
knowing when the script is meant to run would be guessing. Nothing in the
pipeline invokes it.

**Two things the full suite caught that targeted runs did not**, both worth
recording because both were mine:

- The import I inserted into `summarize_rubric20_results.py` landed inside an
  open parenthesised import and made the module unparseable. Seven tests in
  `test_rubric_pooling.py` failed on it — the file that exists because that same
  script had a denominator bug before.
- My first no-defaulting test matched any `get('percentage', 0)`, including an
  element-level read in a legacy branch of the rubric10 renderer — a different
  structure with its own denominator. Narrowed to the overall-score read and
  mutation-checked by reverting a consumer.

`1187 passed, 3 skipped`. 12 mutants on the schemas and validator, no survivors.

PR #329, opened 2026-08-05
Name the nested keys a record may use (#325)

### LINKED ISSUE #325: rubric20 questions saturate: six score identically on all 25 records
Found reviewing #324.

With the dead fields fixed (#319), the hybrid scorer discriminates on 14 of 20
questions across the 25 current records. Six do not — every record gets the same
score:

| | score on all 25 | why |
|---|---|---|
| Q2 Entry Length Adequacy | 5/5 | every record's `description` exceeds 200 chars |
| Q6 Dataset Identification Metadata | 1/1 | every record has `doi` or `page` |
| Q7 Funding and Acknowledgements | 5/5 | every record has `funders` with an award-shaped string |
| Q12 Collection Protocol Clarity | 5/5 | every record fills all three collection groups |
| Q16 Findability (Persistent Links) | 1/1 | every record has `page` |
| **Q9 Access Requirements and Governance** | **3/5** | **no record fills `regulatory_restrictions.*`** |

**Q9 is the different one and the one worth acting on.** The other five are
saturated at the top: the questions are answered well by everything, so they
carry no information but are not wrong. Q9 is pinned at 3 because band 5 needs
confidentiality classification and a governance contact, and
`regulatory_restrictions.confidentiality_level`,
`.governance_committee_contact`, `.hipaa_compliant` and `.other_compliance` are
empty in **0/25** records.

Before #324 this was invisible: Q9 read `access_and_licensing.access_policy`,
which is not a slot of `Dataset`, so band 5 was unreachable by construction and
the constant 3 was an artefact. It is now a true measurement of a real gap —
generation does not populate the governance sub-fields at all.

**Two separable questions:**

1. *Generation*: should the v3 prompt ask for `regulatory_restrictions`
   sub-fields? All four are real slots and all four are empty in every record,
   which suggests the prompt or the schema digest is not surfacing them.
2. *Measurement*: 11 of 88 points (Q2, Q6, Q7, Q12, Q16) are awarded to every
   record unconditionally. They inflate the total without separating methods,
   which is the saturation seen in #169 from a different direction.

Worth settling (1) before the rerun, since it is cheap and changes what the
records contain. (2) is an analysis decision for the manuscript.

realmarcin:
Closes #325's actionable half. The measurement half is answered below and left
for the manuscript.

## Root cause, which is not where the issue guessed

#325 asked whether the v3 prompt should ask for `regulatory_restrictions`
sub-fields. It should not need to: the model was never told they exist.

`NestedClass` has carried an `optional` list since the first version. `build()`
populated it. `render()` never emitted it. So the digest said which nested keys
were **mandatory** and never which were **available**:

- **54 of 66** nested classes rendered as a bare `required: none`
- **367** attributes were named nowhere in the text the model receives

`ExportControlRegulatoryRestrictions` requires nothing, so a run saw exactly:

```
- **ExportControlRegulatoryRestrictions** — required: none
```

**What the records do instead is the tell.** Every one writes prose into that
object's `description`, and the prose *contains the answers*: "FDA Regulated:
No", "De-identified Samples: Yes", a license clause about data security and
privacy standards. The model had the facts and no structured place it had been
told about.

## Why the digest and not the prompt

The digest is the layer that says what shape an answer takes, it is shared by
every arm, and fixing it there keeps generic-v3 a controlled condition rather
than one carrying a field-specific instruction. It was also simply incomplete —
it drops half its own data structure.

## The by-reference shortcut, and its risk

Three classes mostly repeat the top-level slot listing — `Dataset` (100%),
`DataSubset` (98%), `FileCollection` (85%) — and are the only ones large enough
to truncate. Listing their alphabetically-first 24 would be redundant *and*
arbitrary, so they are described by reference **plus** what makes them
different:

```
- **DataSubset** — required: `id`
    - also accepts the same slots as the top-level listing above plus
      `is_data_split`, `is_subpopulation`
```

Those two are exactly what rubric20 Q5 and Q19 score. A shortcut that swallowed
them would have removed the only reason to read the entry, so a test asserts
every non-overlapping attribute survives it.

## Cost

`24,273 → 29,114` chars, **+19.9%**, roughly +1,200 tokens in a cached prefix.
Still an eighth of the 254 KB merged schema. A regression that enumerated the
three mirroring classes in full would roughly double the digest; a test bounds it.

**No attestation is affected.** `schema_digest_md5` is written by `api_runner`
but appears in **0** existing run provenances, so nothing verifies against the
old fingerprint. The rerun records the new one.

## What these tests do not claim

They assert the *listing*, not the outcome. Making a field addressable is what
the digest can do; whether a run then fills it is a question for the next
generation. Two of the four Q9 fields were already visible via their enum
vocabularies and still went unfilled in 25/25 records — so I am not claiming
this fixes Q9, only that it removes the one cause that was certain.

## #325's other half, not addressed here

Five questions (Q2, Q6, Q7, Q12, Q16 — 11 of 88 points) are awarded to every
record unconditionally. That is saturation, not a defect, and what to do about
it is an analysis decision for the manuscript rather than a code change. Left in
#325.

10 tests, mutation-checked with 5 reintroductions, no survivors.
`1197 passed, 3 skipped`.

### DIFF
--- a/src/data_sheets_schema/schema_digest.py
+++ b/src/data_sheets_schema/schema_digest.py
@@ -41,6 +41,21 @@ DESCRIPTION_CHARS = 300
 # vocabulary costs 38 bytes more than the clipped one, in a cached prefix.
 MAX_ENUM_VALUES = 60
 
+# Attribute names every class inherits. Listing them on all 66 nested classes
+# would be 264 tokens saying nothing — the model already reaches for
+# `description`, which is the problem this listing exists to solve.
+UNIVERSAL_ATTRIBUTES = frozenset({"id", "name", "description", "used_software"})
+
+# Optional attributes listed per nested class before truncating.
+MAX_OPTIONAL_SHOWN = 24
+
+# Above this overlap with the top-level slot listing, a nested class is
+# described by reference instead of enumerated. Exactly three classes qualify —
+# `Dataset` (100%), `DataSubset` (98%) and `FileCollection` (85%) — and they are
+# the three large enough to be truncated, so listing the alphabetically-first 24
+# would spend the tokens on a redundant and arbitrary slice.
+MIRRORS_TOP_LEVEL = 0.8
+
 
 @dataclass
 class SlotDigest:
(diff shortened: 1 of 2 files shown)


realmarcin:
## Review pass

Two findings, both defects in this PR, both fixed in the follow-up commit.

**#330 — the truncation safeguard had never run.** `MAX_OPTIONAL_SHOWN`, the
slice and the `(+N more)` tail are unreachable against both target classes:

    Dataset:      3 by-reference,  0 truncated
    CoreDataset:  1 by-reference,  0 truncated

The three classes large enough to truncate all exceed `MIRRORS_TOP_LEVEL` and
take the by-reference path instead. Dead code in a prompt-building path reads as
a verified guard against an oversized digest and is not — the first run of that
branch would otherwise be inside a paid generation run. Kept (a 90-attribute
non-mirroring class is a real future case) and now exercised against a synthetic
`ClassDigest`.

**#331 — the tests covered `Dataset` only.** `CoreDataset` is its own merged
artifact, is the `*_core` arm of every run and half the rerun's evaluations, and
takes a *different branch*: its smaller top-level listing means one nested class
mirrors it against three. Every general invariant now runs over both entries in
`CLASS_SCHEMA`, and the four governance fields are asserted on both since the
core arm scores the same Q9.

Mutation-checked with 8 reintroductions, no survivors — including three the
Dataset-only, cap-untested version would have missed: an off-by-one in the cap,
a dropped `(+N more)`, and a reworded by-reference line.

`1202 passed, 3 skipped`.

PR #333, opened 2026-08-05
Refuse to repair records newer than the repair tool (#328); measure the saturation (#325)

### LINKED ISSUE #325: rubric20 questions saturate: six score identically on all 25 records

### LINKED ISSUE #328: fix_evaluation_scores.py writes the superseded overall_score shape
Found reviewing #327.

`scripts/fix_evaluation_scores.py` recomputes an evaluation's totals and
rewrites `overall_score`. It reads and writes `percentage`:

```python
current_percentage = eval_data.get("overall_score", {}).get("percentage", 0)
...
calculated_percentage = (calculated_total / calculated_max * 100) if calculated_max > 0 else 0
```

#327 established that the current contract is `normalized_percentage` over
`adjusted_max_points` — the denominator *after* N/A exclusions — and removed
`percentage` from both semantic schemas. So this script, run on a current
record, would:

1. compute a percentage against the **unadjusted** maximum, which is a different
   number from the one the record reports whenever anything is excluded; and
2. write it under a key the schema no longer declares, turning a conformant
   record into a non-conformant one.

**Left alone deliberately in #327.** Its read of `percentage` defaulting to `0`
is load-bearing there rather than a bug — it compares current against recomputed
to decide whether a rewrite is needed, and "absent" genuinely should trigger one.
Changing that without understanding when the script is meant to run would be
guessing.

**What needs deciding:** is this script still live? It appears to be a one-off
repair for the 2025-12 corpus. If so it should say so and refuse to touch
records in the current shape. If it is meant to stay usable, it needs the N/A
vocabulary — `excluded_max_points`, `adjusted_max_points`,
`normalized_percentage`, and the per-rubric not-applicable counter — and to
recompute against the adjusted denominator.

No urgency for the rerun: nothing invokes it in the pipeline. Filing so it does
not get run later on the assumption that it is current.

realmarcin:
Closes #328. Closes #325.

Two small pieces of evaluation-reporting hygiene before the rerun, on one branch
because each is a few lines and both concern the same corpus.

## #328 — the repair tool now refuses the current shape

`scripts/fix_evaluation_scores.py` recomputes a percentage as `total / max` and
writes `overall_score.percentage`. Both halves are wrong for a record written
after the N/A convention:

- the reported figure is `normalized_percentage` over `adjusted_max_points` —
  the maximum *after* exclusions — so `total / max_points` is a different number
  whenever anything is excluded;
- `percentage` is not declared by either semantic schema (#323), so writing it
  turns a conformant record non-conformant.

Nothing in the pipeline invokes it, and everything it could still legitimately
repair — the 12 superseded and 8 invalid records — is being regenerated. So it
refuses rather than being migrated: that would be work spent on a corpus about to
be replaced. The guard runs *before* the rubric-type check and before any
rewrite, and `dry_run` defaults to `False`, so a refusal placed after the rewrite
would be no guard at all — there is a test for exactly that.

Against the live corpus: **8 refused, 0 modified.**

A refusal is reported separately from a fault. A traceback here would read as a
problem with the corpus rather than a tool that is out of date.

## #325 — the saturation, measured, and a correction

I reported "11 points across five questions". It is **six questions and 20
points** — the count omitted Q9, and the arithmetic did not add up either way.

| | question | score | of |
|---|---|---|---|
| Q2 | Entry Length Adequacy | 5 | 5 |
| Q6 | Dataset Identification Metadata | 1 | 1 |
| Q7 | Funding and Acknowledgements | 5 | 5 |
| **Q9** | **Access Requirements and Governance** | **3** | **5** |
| Q12 | Collection Protocol Clarity | 5 | 5 |
| Q16 | Findability (Persistent Links) | 1 | 1 |

Q9 differs in kind. The other five saturate at their maximum — the corpus
answers them uniformly well. Q9 is pinned at 3 because its top band needs a
confidentiality classification and a governance contact, and **0 of 25** records
populate those fields; until #329 the digest never named them.

**Why this confounds nothing, stated precisely rather than asserted:**

```
full 88-point score      mean 68.9   sd 11.0413
66-point subscore        mean 48.9   sd 11.0413
```

Identical to four decimal places, by construction. Every point of between-record
variation comes from the 14 discriminating questions. What saturation costs is
the absolute percentage: 22.7 of 78.3 points are awarded before a datasheet is
read.

`notes/rubric20_saturation_2026-08-04.md` carries the numbers, the recommended
manuscript wording, and the argument **against** dropping the questions —
rubric20 is the published instrument, a 66-point variant would not be comparable
to anything already scored, and saturation is a property of a four-project corpus
rather than of the questions. A fifth project omitting its funders would
un-saturate Q7 immediately.

## Testing

8 tests, mutation-checked with 7 reintroductions. One survivor first time round,
**in the test rather than the code**: it iterated over
`module.CURRENT_SHAPE_MARKERS`, so narrowing that tuple to one entry made it pass
vacuously — a test deriving its expectation from the code it checks. The markers
are now named independently, with a complementary test that the module
recognises exactly those three.

`1210 passed, 3 skipped`.

### DIFF
--- a/notes/rubric20_saturation_2026-08-04.md
+++ b/notes/rubric20_saturation_2026-08-04.md
@@ -1,0 +1,82 @@
+# Six rubric20 questions score identically on every record
+
+**Recommended wording for the manuscript is in the last section.** The rest is
+the measurement behind it.
+
+Measured 2026-08-04 over the 25 current `claudecode_agent` records
+(`2026-07-31`, generic and generic-v2), scored with the hybrid scorer as
+corrected in #319. **These figures are not comparable to any rubric20 number
+reported before that date** — six of 88 points were unearnable by any record
+until it landed.
+
+## What saturates
+
+Six of twenty questions award the same score to all 25 records:
+
+| | question | score | of |
+|---|---|---|---|
+| Q2 | Entry Length Adequacy | 5 | 5 |
+| Q6 | Dataset Identification Metadata | 1 | 1 |
+| Q7 | Funding and Acknowledgements Completeness | 5 | 5 |
+| Q9 | Access Requirements and Governance Documentation | **3** | 5 |
+| Q12 | Collection Protocol Clarity | 5 | 5 |
+| Q16 | Findability (Persistent Links) | 1 | 1 |
+
+**20 of 88 points are awarded unconditionally.** An earlier note in #325 said
+"11 points across five questions"; both figures were wrong — it omitted Q9 and
+the arithmetic did not add up. The measured figure is six questions and 20
+points.
+
+Q9 is the one that differs in kind. The other five saturate *at their maximum*:
+every datasheet in the corpus genuinely answers them well. Q9 saturates at 3 of
+5 because its top band requires a confidentiality classification and a governance
+contact, and no record populates `regulatory_restrictions.confidentiality_level`,
+`.hipaa_compliant`, `.other_compliance` or `.governance_committee_contact` —
+0 of 25. Until #329 the digest never named those fields, so no run was told they
+existed. Whether the next generation fills them is open.
+
+## Why this does not affect comparisons
+
+The saturated questions contribute a constant to every record, so they move the
+mean and leave the variance untouched. Removing them changes nothing about how
+records rank or how far apart they are:
+
+```
+full 88-point score      mean 68.9   sd 11.0413
+66-point subscore        mean 48.9   sd 11.0413
+```
+
+The standard deviations are identical to four decimal places — not
+approximately, but by construction. **Every point of between-record variation
+comes from the 14 discriminating questions.**
+
+So the saturated questions are not a confound for any between-method or
+between-condition comparison. What they do is inflate the absolute percentage:
+68.9/88 is 78.3%, of which 22.7 percentage points are awarded before any
+datasheet is read.
+
+## Recommended wording
+
+> Scores are reported against the full 88-point rubric20 instrument. Six of its
+> twenty questions (Q2, Q6, Q7, Q9, Q12, Q16) awarded an identical score to all
+> 25 records in this corpus, contributing 20 points unconditionally. Because
+> these contribute a constant, they do not affect any comparison reported here —
+> the standard deviation of the 66-point discriminating subscore is identical to
+> that of the full score (11.04 in both cases) — but absolute percentages should
+> be read with the 20-point floor in mind. Five of the six saturate at their
+> maximum, indicating questions the corpus answers uniformly well; Q9 saturates
+> at 3 of 5 because no record populates the governance sub-fields its top band
+> requires.
+
+## What not to do
+
+**Do not drop the saturated questions from the instrument.** rubric20 is the
+published rubric; a 66-point variant would not be comparable to anything already
+scored, including the presence path and the manuscript's own earlier figures, and
+the saturation is a property of *this corpus* rather than of the questions. Four
+projects is a small sample: a fifth that omits its funders would un-saturate Q7
+immediately.
+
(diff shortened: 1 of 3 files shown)


realmarcin:
## Review pass

Two findings, both fixed in the follow-up commit. Nothing filed — see the last
section for the one I deliberately did not raise.

**The refusal was invisible to a caller.** `main()` returned 0 unconditionally,
so a directory in which every record was refused reported success having done
nothing — the same silence the guard exists to break. It now returns 1 when
everything was refused and nothing fixed, and still 0 when there was something to
repair, so the guard does not turn the tool into one that always fails.

**`--input-dir` outside the working tree crashed.** Pre-existing:
`eval_dir.relative_to(Path.cwd())` raises for a path outside cwd, so the
documented flag failed before doing any work. Surfaced by writing a test that
runs the CLI against a temp directory.

**How that one surfaced is worth recording.** The crash *also* exits 1, so a test
asserting only `returncode != 0` would have passed for entirely the wrong reason.
It was asserting on stdout as well that caught it. Both exit-code tests now check
the output rather than just the status.

## Noted, not filed

The default invocation auto-detects `rubric20_semantic/concatenated` and never
visits the rubric10 directory — so the "8 refused, 0 modified" figure in the PR
body covers only the rubric20 set. The 8 current-shape rubric10 records are
equally refusable and are covered by `TestTheLiveCorpus`, which globs both.

Not filed, on the judgement that this is a pre-existing quirk of directory
discovery in a tool that this PR is effectively retiring: it prints which
directory it chose, `--input-dir` reaches the other, and every record either
directory holds is being regenerated. Filing it would be noise against a frozen
script. Flagging it here so the decision is on the record rather than an
oversight.

`1212 passed, 3 skipped`. 11 mutants across both commits, no survivors.

Issue #330, opened 2026-08-05
schema_digest's optional-attribute truncation branch is unreachable


realmarcin:
Found reviewing #329.

#329 renders `NestedClass.optional`, capped by `MAX_OPTIONAL_SHOWN = 24` with a
`(+N more)` tail. Measured against both target classes, **no input reaches it**:

```
Dataset:      3 by-reference classes,  0 truncated
CoreDataset:  1 by-reference class,    0 truncated
```

The three classes large enough to truncate — `Dataset`, `DataSubset`,
`FileCollection` — all exceed the `MIRRORS_TOP_LEVEL` threshold and take the
by-reference path instead. So the cap, the slice, and the `(+N more)` string
have never been rendered.

Dead code in a prompt-building path is worse than usual: it reads as a verified
safeguard against an oversized digest, and it is not. If the schema later gains
a large class that does *not* mirror the top-level listing, the first time that
branch runs will be in a paid generation run.

**Not simply removable.** The guard is genuinely wanted — a 90-attribute
non-mirroring class would otherwise be enumerated in full into every prompt.

**Fix:** exercise it against a synthetic `ClassDigest` with one large
non-mirroring nested class, asserting both the cap and the `(+N more)` tail.
That costs nothing and makes the safeguard real. Doing this in #329.

realmarcin:
Fixed in #329 — see the commit addressing the review of that PR.

PR #334, opened 2026-08-05
Measure enumeration depth, and correct #332's finding

### LINKED ISSUE #332: API records enumerate far less than agent-runtime records, and rubric20 cannot see it
Prompted by the observation that recent generations look worse than earlier ones.
Measured across the corpus with the current (post-#319) hybrid scorer.

## The headline drop is a sampling artifact

Run-level means fall from ~74/88 to ~51/88 between 2026-07-27 and 2026-08-02.
Almost all of that is **which projects each run produced**: the 08-02 `bare`
runs made CHORUS only, and CHORUS scores 51–61 in *every* run while AI_READI,
CM4AI and VOICE score 72–82.

    CHORUS only:   agent 58–61   |   API-high 51–53   |   API-default 51

## No model changed

Every run is opus-5. What varied is the runtime and the effort tier:

    2026-07-27/28   claude-opus-5[1m]           Claude Code agent
    2026-07-29/31   google/claude-opus-5-high   CBORG API, high tier
    2026-08-02      google/claude-opus-5        CBORG API, default tier

**Effort tier shows no effect**: CHORUS scores 52.5 and 50.7 on API-high against
51.3 on API-default.

## Runtime, paired on the same prompt, is inside the noise

The one clean contrast is `generic` (agent, 07-28) against `api-generic` (API,
07-31) — same prompt, n=3 each:

| project | agent | API | delta |
|---|---|---|---|
| AI_READI | 73.7 | 76.0 | **−2.3** |
| CHORUS | 59.3 | 53.0 | +6.3 |
| CM4AI | 79.7 | 72.3 | +7.3 |
| VOICE | 78.3 | 77.7 | +0.7 |

Mean **+3.0/88, sd 4.6**, and it flips sign on AI_READI. Four projects cannot
resolve a difference this size — the same wall as #169, which put the requirement
at ~110 projects.

## What is real, and what the rubric misses

API records are **30–40% shorter**, and the loss is concentrated in the
*enumerative* slots:

| slot | agent | API | change |
|---|---|---|---|
| `resources` | 14,126 | 3,113 | **−78%** |
| `variables` | 11,448 | 2,579 | **−77%** |
| `creators` | 7,016 | 2,803 | −60% |
| `collection_mechanisms` | 4,538 | 2,130 | −53% |
| `instances` | 2,856 | 1,254 | −56% |

(chars, mean over 12 records per side)

A datasheet listing 30 variables instead of 200 reads as *much* worse to a human
and scores the same, because no rubric20 question rewards enumeration depth: Q3
counts keywords, Q4 counts file types, and nothing counts variables, resources
or instances.

So the perception is well founded and the instrument is blind to it.

## Why this matters before the rerun

The rerun is planned on the API path. If enumeration depth is part of what the
manuscript claims about generation quality, then:

1. **the API arm needs to enumerate like the agent arm did** — plausibly a phase
   or iteration difference rather than a model one, since the `bare` runs made
   6–8 phase calls against 30–78 for the 07-31 runs; and
2. **rubric20 needs a question that sees enumeration depth**, or the result will
   report parity where a reader sees a large difference.

## Caveats

- These scores come from the scorer corrected in #319 today. They are **not**
  comparable to any figure reported earlier in this project.
- The scorer is presence- and heuristic-based. It cannot see correctness, and
  there is no gold standard (#177). "More enumerated" is not "more accurate" —
  a longer `variables` list could be better coverage or could be invention.
  Distinguishing those needs the grounding axis, not this one.

realmarcin:
Closes #332.

## The issue's headline was wrong

#332 reported the agent runtime enumerating "about twice as many items" as the
direct API. Measured properly, **the API enumerates more items on all four
projects**:

| project | items agent | items API | chars agent | chars API |
|---|---|---|---|---|
| AI_READI | 170 | **226** | 121,512 | 87,704 |
| CHORUS | 129 | **140** | 52,881 | 34,539 |
| CM4AI | 139 | **249** | 104,836 | 70,011 |
| VOICE | 126 | **227** | 120,071 | 87,704 |

The API enumerates *more* and writes *less*. The two measures point in opposite
directions, which is exactly why the original claim went wrong: it came from
per-slot character counts, which measure elaboration, and from an item count
taken over the minority of items carrying a structured label. Both were read as
depth.

Paired runs — identical prompt file, identical four-phase mode, differing only in
`Agent runtime`. Three replicates each.

What a reader notices as "thinner" records is **elaboration per entry**, not how
many entries there are.

## Coverage before grounding

`enumeration.py` measures grounding by finding *anchors* — DOIs, ORCIDs, RRIDs,
URLs, grant numbers, tokens mixing letters and digits — verbatim in the source
bundle. Prose is deliberately unmatched: every English word appears somewhere in
a 200 KB bundle, so counting those would drive every rate to 100% and make the
arms indistinguishable.

```
agent   items 1690   checkable 858 (51%)   anchors 2514   found 1799 (72%)
API     items 2528   checkable 829 (33%)   anchors 1367   found 1236 (90%)
```

**Half the agent's items and two thirds of the API's carry nothing verifiable**,
so neither 72% nor 90% is a statement about the record as a whole. `grounding`
raises `NothingCheckable` where no anchor existed — `0.0` would read as invented
and `1.0` as verified, and the measurement is absent rather than zero.

Read together the numbers resist a simple story, and the note says so rather than
picking the flattering reading. The agent puts **1.8× as many checkable anchors
into half as many items** — more falsifiable content per entry — and a larger
share of them are not found in the sources. That is consistent with inventing
more, and equally consistent with paraphrasing identifiers or citing across
document boundaries. Substring matching cannot separate these.

## What this changes about #332's two proposals

**Do not add an enumeration question to rubric20.** The judge receives only the
D4D record, so it cannot see sources and any question would score a count. On
these numbers a count question would have ranked the API arm *above* the agent
arm, while the 28% of the agent's anchors that are unverified went unexamined.

**Do not change the API arm before the rerun.** The gap is a finding about
runtimes, and generic-v3's value is being a clean prompt condition comparable to
v1 and v2. If elaboration depth matters for the manuscript, the agent arm already
has data on disk.

## Also removed

A `MIN_ANCHOR_CHARS` constant no input could reach: `\b` keeps trailing
punctuation out of a `code` match, so the strip never shortens one, and the
identifier patterns all exceed five characters. Found by mutation — changing it
from 5 to 1 left every test green. The regex quantifier is now the only floor,
pinned by a test at the boundary. Same shape as #330, but deleted rather than
exercised: unlike that truncation cap there is no future case it guards.

20 tests, mutation-checked with 8 reintroductions, no survivors.
`1232 passed, 3 skipped`.

### DIFF
--- a/notes/enumeration_depth_2026-08-04.md
+++ b/notes/enumeration_depth_2026-08-04.md
@@ -1,0 +1,86 @@
+# The API writes fewer, longer entries — not fewer entries
+
+**This corrects the finding in #332.** That issue reported the agent runtime
+enumerating "about twice as many items" as the direct API. Measured properly, the
+API enumerates **more** items on all four projects. What the agent produces is
+more *text per item*.
+
+The original claim came from per-slot character counts, which measure
+elaboration, and from an item count taken over the minority of items that carry a
+structured label. Both were read as depth. They are not.
+
+## The measurement
+
+Paired runs, `2026-07-28_claude-opus-5-generic` against
+`2026-07-31_claude-opus-5-api-generic`: identical prompt file
+(`d4d_generic_arm_prompt.md`), identical four-phase mode, differing only in
+`Agent runtime` (Claude Code against Claude API direct). Three replicates each.
+
+| project | items agent | items API | chars agent | chars API |
+|---|---|---|---|---|
+| AI_READI | 170 | 226 | 121,512 | 87,704 |
+| CHORUS | 129 | 140 | 52,881 | 34,539 |
+| CM4AI | 139 | 249 | 104,836 | 70,011 |
+| VOICE | 126 | 227 | 120,071 | 87,704 |
+
+**The API enumerates more and writes less.** Both consistently, across all four
+projects — the two measures point in opposite directions, which is why reporting
+either alone is misleading.
+
+Per slot the item counts are mixed rather than uniform: the agent lists more
+`variables` on AI_READI (45 against 22) and more `resources`, while the API lists
+more `external_resources` everywhere and nearly three times the `creators` on
+CM4AI. The character counts are not mixed — the agent writes more in almost every
+slot, up to 13× on CM4AI `resources`.
+
+So the reader's impression of "thinner" records is about **elaboration per
+entry**, not how many entries there are.
+
+## How much of this can be checked at all
+
+`src/data_sheets_schema/enumeration.py` measures grounding by looking for
+*anchors* — DOIs, ORCIDs, RRIDs, URLs, grant numbers and other tokens mixing
+letters and digits — verbatim in the project's source bundle. Ordinary prose is
+deliberately not matched: every English word appears somewhere in a 200 KB
+bundle, so counting those would drive every rate to 100%.
+
+    agent   items 1690   checkable 858 (51%)   anchors 2514   found 1799 (72%)
+    API     items 2528   checkable 829 (33%)   anchors 1367   found 1236 (90%)
+
+**Read the coverage column before the grounding column.** Half the agent's items
+and two thirds of the API's carry nothing verifiable at all, so neither 72% nor
+90% is a statement about the record as a whole.
+
+The two numbers have to be read together, and doing so is what makes the
+comparison hard rather than easy:
+
+- The agent puts **1.8× as many checkable anchors** into half as many items —
+  its entries carry substantially more specific, falsifiable content.
+- A larger share of those anchors are **not found** in the sources (72% against
+  90%). That is consistent with the agent inventing more, and equally consistent
+  with it paraphrasing identifiers, citing things stated across document
+  boundaries, or elaborating from context. Substring matching cannot separate
+  these.
+
+**What this does not establish.** It measures verifiability, not truth. An anchor
+found in the source shows the record did not invent that token; it says nothing
+about whether the surrounding claim is right. Both arms are measured identically,
+so the comparison is sound even though each absolute rate is a lower bound.
+
+## What follows
+
+**Do not add an enumeration question to rubric20.** The judge receives only the
+D4D record — `rubric20_system_prompt.md` says "Read the provided D4D YAML file" —
+so it cannot see sources and any question it asked would score a count. On these
+numbers a count question would have ranked the API arm *above* the agent arm
+while the 28% of its anchors that are unverified went unexamined.
+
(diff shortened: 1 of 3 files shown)


realmarcin:
## Review pass

Two findings, filed as #335 and #336, both fixed. Between them they produced
almost the whole of the headline gap **and inverted its interpretation** — so the
PR body above is now wrong where it discusses grounding, and the note has been
corrected.

**#335 — minted ids counted as ungrounded.** A record's own `id`
(`CM4AI_creator_release_authors`, `aireadi:variable_ntprobnp`) cannot appear in a
source by construction. Counting those as anchors that were not found measured
identifier *style* and penalised whichever arm gives things stable identifiers.
Now excluded at any depth.

**#336 — canonical URLs read as invention.** Of the 119 `creators` anchors the
agent emitted that were not found, **115 were URLs**, every one an institutional
homepage: `vumc.org`, `emory.edu`, `ufl.edu`, `sickkids.ca`, `wustl.edu`,
`weill.cornell.edu`, `usf.edu`. Correct Bridge2AI VOICE affiliations. The sources
name them in prose — "Vanderbilt University Medical Center" — and the record
supplies the canonical URL. That is enrichment from world knowledge, and
substring matching cannot tell it from invention.

Not fixable, so made visible: URL anchors are counted separately. Resolving
institution names would need an external vocabulary — importing exactly the world
knowledge in question — and dropping URLs entirely would discard 303 of the API
arm's anchors, largely landing pages that *are* in the sources.

### Corrected result

```
agent   identifier anchors 1264  found 1225 (97%)   url 459  found 424 (92%)
API     identifier anchors  948  found  929 (98%)   url 303  found 300 (99%)
```

**The arms are indistinguishable on identifier fidelity: 97% against 98%.** The
72%-against-90% figure in the PR body, and the sentence calling it "consistent
with the agent inventing more", were artifacts of the two faults above. Neither
survives.

One more check worth recording: normalised by text volume, *found* anchor density
is 1.50 per 1,000 characters for the agent against 1.47 for the API. Per unit of
text, verified content is equally dense — so the agent's larger anchor total
follows from writing 1.4× as much, not from packing in more evidence. That
undercuts the "more falsifiable content per entry" claim I made, too.

The item-count correction in the PR title stands unchanged: the API enumerates
more items on all four projects.

`1238 passed, 3 skipped`. 14 mutants across both commits, no survivors.

PR #339, opened 2026-08-05
Stamp v3 as v3 (#337); put the scalar rule in v4, not v3 (#338)

### LINKED ISSUE #337: generic-v3 stamps every record as generic-v2
Found auditing open work against the planned rerun. **Blocking, and only fixable
before the first v3 run.**

`src/download/prompts/d4d_generic_arm_prompt_v3.md`'s HEADER BLOCK is a verbatim
copy of v2's:

```
# Mode: four-phase project agent, generic-v2 prompt
# Prompt: src/download/prompts/d4d_generic_arm_prompt_v2.md (identical for all projects)
```

Those two lines are byte-identical to what the existing
`2026-07-31_claude-opus-5-generic-v2` records already carry. So every record the
v3 arm produces would claim to have been generated under v2.

`api_runner` does not repair it: the only header rewrite is
`"generic prompt"` → `"tuned prompt"` (lines 290-291), with no v2→v3 case.

**Why this matters more than a cosmetic slip.** The comparison v3 exists to make
is v2 against v3. The record header is the human-readable provenance and the
first thing a reader checks — earlier today it was exactly these lines that
established a paired agent/API comparison shared a prompt file. A v3 corpus
stamped v2 makes that check report the wrong answer, and makes the two arms
indistinguishable in the artifact itself.

`CONDITION_AXES['generic_v3']` is correct (`{'base': 'v3', 'tuned': False}`) and
the run label carries the condition, so the damage is confined to the header —
but the header is what a person reads.

**Fix:** change the two lines to name v3, and add a test that each condition's
prompt body stamps its own file. The current arrangement lets any future vN
inherit the wrong name by copy-paste, which is how this one arose.

**Timing.** v3 has never been run — `be5a0935` is titled "staged, not run" — so
editing it now costs nothing. After the first v3 generation the prompt is frozen
and the corpus carries the wrong stamp permanently.

### LINKED ISSUE #338: generic-v3 lacks the scalar-range companion rule, and its existing rule may push toward the defect
Found auditing open work against the planned rerun. Related to #292 and #297.

## What v3 has

One added rule, the companion to v2's hollow-object defect:

> When a slot's declared range is a **class**, populate the fields that class
> declares. Placing the content in a free-text field such as `description` while
> the declared fields stay empty produces an object of the correct shape holding
> none of the structure it exists to carry.

## What it does not have

The rule requested alongside it: **a scalar range takes an identifier, not the
object it identifies.** Nothing in v3 says this.

That matters because #297 established LinkML cannot express a
string-or-inline-object range, so `target_dataset` stays `range: string` and the
prompt is the only available remedy.

## The measured consequence

All three current VOICE replicates still fail validation, each differently:

```
rep1  inline Dataset object in target_dataset   <- needs the missing rule
rep2  relationship_type: related_to             <- no enum equivalent; genuine failure
rep3  relationship_type: IsNewVersionOf         <- fixed: normalise_enum_aliases is
                                                   now on the write path (api_runner:1308)
```

rep3's class is already handled. rep2 has no vocabulary equivalent and is
deliberately left to fail. **rep1 is the one a prompt rule could address, and v3
does not address it.**

## Why this may be worse than an omission

v3's rule tells the model that a class-ranged slot deserves its declared fields
populated. `related_datasets` *is* class-ranged (`DatasetRelationship`), and a
model applying that rule energetically has an obvious next step: populate the
related dataset properly too — into `target_dataset`, which is a string.

That is precisely rep1's failure. So v3 plausibly increases the rate of the one
VOICE failure mode still outstanding, while carrying no rule against it.

This is a hypothesis, not a measurement — v3 has never been run. But it is
cheap to guard against and expensive to discover afterwards.

## Fix

Add the companion sentence to the v3 body, before the first v3 run. Same
timing constraint as #337: `be5a0935` is "staged, not run", so the file is still
editable; after the first generation it is frozen.

Wording that matches the existing rules' register:

> Where a slot's declared range is a scalar, populate it with the identifier of
> the thing it points to, not with the thing itself. An object placed in a
> string-ranged slot fails validation and loses the reference it was meant to
> record.

realmarcin:
Closes #337. Closes #338.

Both were fixable only before the first v3 run. Verified none has happened: no
v3 run directory exists, and the v3 prompt's md5 is attested by nothing.

## #337 — v3 stamped every record as v2

Its HEADER BLOCK was copied from v2 verbatim:

```
# Mode: four-phase project agent, generic-v2 prompt
# Prompt: src/download/prompts/d4d_generic_arm_prompt_v2.md
```

Byte-identical to what the 2026-07-31 generic-v2 records already carry, and
`resolve_prompt` repairs the header only for the tuned arm.

Not cosmetic: the comparison v3 exists to make is v2 against v3, the header is
the human-readable provenance, and it is what a reader checks first — the paired
agent/API comparison in `notes/enumeration_depth_2026-08-04.md` was established
from exactly these two lines.

`test_prompt_self_identification.py` now checks **every** condition, since
copy-paste is how the next version would inherit it too. Its premise tests run
the resolver rather than reading its source, so a header rewrite added later
fails loudly instead of leaving them stale.

## #338 — the scalar rule belongs in v4, not v3

**My first attempt put it in v3, and two existing tests failed. They were right.**

`test_generic_v3_prompt.py` asserts v3 is "v2 plus the marked block and nothing
else", and `notes/generic_v3_analysis_plan.md` is **registered before any
generation run** and names exactly one rule. Adding a second would have
invalidated the pre-registration and made the v2-against-v3 delta unattributable
to either rule.

So v3 keeps only the header fix, and v4 carries the rule:

> Where a slot's declared range is a scalar, populate it with the identifier of
> the thing it refers to, not with the thing itself.

`notes/generic_v4_analysis_plan.md` registers the prediction in advance,
including the guard that matters — a rule about scalars that suppresses
legitimate object population has traded one defect for another, exactly as v2's
rule 1 did — and records that a **null result should retire the rule** rather
than keep it because it sounds sensible.

## Adding v4 exposed a real gap in `comparable_conditions`

It promised a difference "attributable to one change" and returned `True` for v2
against v4, which spans **two** rule additions: `condition_delta` reports the
single axis `base` regardless of how many versions apart. Latent already for v1
against v3; v4 made it load-bearing.

```
generic    vs generic_v2   True
generic_v2 vs generic_v3   True
generic_v3 vs generic_v4   True
generic_v2 vs generic_v4   False   <- two additions
generic    vs generic_v3   False   <- two additions
generic_v2 vs tuned        False   <- two axes
```

## Not wired by hand

`_CONDITIONS = sorted(CONDITION_PROMPTS)`, a lesson already learned when
`generic_v2` was staged, tested and unreachable from every CLI entry point. A
test pins that v4 is launchable.

## What this does not do

It does not decide whether to run v4. That is a study-design call: v3 alone
answers whether the inline-object failure recurs, and if it does not, v4's rule
should be retired unrun. The analysis plan says so explicitly.

Mutation-checked with 7 reintroductions, no survivors. `1261 passed, 3 skipped`.

### DIFF
--- a/notes/generic_v4_analysis_plan.md
+++ b/notes/generic_v4_analysis_plan.md
@@ -1,0 +1,62 @@
+# generic-v4 analysis plan
+
+Registered before any v4 generation run. The prediction is here rather than in
+the prompt because writing it there would instruct the model to produce the
+result the run is meant to test.
+
+## What v4 changes
+
+One rule, added to v3's:
+
+> Where a slot's declared range is a scalar, populate it with the identifier of
+> the thing it refers to, not with the thing itself. An object placed in a
+> string-ranged slot fails validation and loses the reference it was meant to
+> record, even where that thing is richly described elsewhere in the record.
+
+## Why
+
+v3 tells the model to populate the fields a class-ranged slot declares. It says
+nothing about where structure does *not* belong, and `related_datasets` is
+class-ranged while its `target_dataset` is `range: string`.
+
+All three 2026-07-31 VOICE replicates fail validation on `related_datasets`
+(#292), by three different routes:
+
+| replicate | failure | status |
+|---|---|---|
+| rep1 | inline `Dataset` object in `target_dataset` | **this rule** |
+| rep2 | `relationship_type: related_to` | no enum equivalent; a real generation failure |
+| rep3 | `relationship_type: IsNewVersionOf` | already fixed — `normalise_enum_aliases` is on the write path |
+
+#297 established that LinkML cannot express a string-or-inline-object range, so
+the schema cannot refuse an object in `target_dataset`. A prompt rule is the only
+available remedy.
+
+## Prediction
+
+Registered in advance, and falsifiable:
+
+1. **Primary.** The inline-object failure in `target_dataset` does not occur in
+   any v4 replicate. Measured as `linkml-validate` errors of the form
+   `... is not of type 'string' in /related_datasets/*/target_dataset`.
+2. **Guard against over-application.** v4 does not reduce the population of
+   class-ranged slots relative to v3. A rule about scalars that suppresses
+   legitimate object population has traded one defect for another, which is
+   exactly what v2's rule 1 did (`notes/form_defect_split_2026-08-03.md`) and
+   what v3 exists to repair. Measured as items enumerated per record via
+   `data_sheets_schema.enumeration.total_depth`, v4 against v3.
+3. **Null result is informative.** If rep1's failure mode does not recur under
+   v3 either, the rule is unnecessary and should be retired rather than kept
+   because it sounds sensible. v3 must be run to know this.
+
+## What would falsify it
+
+Prediction 1 fails if any v4 replicate places an object in a string-ranged slot.
+Prediction 2 fails if v4's mean enumerated depth is materially below v3's — the
+same trade v2 made, in the opposite direction.
+
+## Comparison
+
+v3 against v4 isolates this rule. v2 against v4 does **not**: it spans two rule
+additions, and `comparable_conditions("generic_v2", "generic_v4")` returns False
+for that reason.
--- a/src/data_sheets_schema/api_runner.py
+++ b/src/data_sheets_schema/api_runner.py
@@ -61,9 +61,15 @@ GENERIC_PROMPT_V2 = PROMPTS / "d4d_generic_arm_prompt_v2.md"
 # was one: v2 produced the 2026-07-31 series and is the baseline v3 is measured
 # against. The comparison that isolates the companion rule is v2 against v3.
 GENERIC_PROMPT_V3 = PROMPTS / "d4d_generic_arm_prompt_v3.md"
+# v4 carries the scalar-range companion to v3's class-range rule. It is a
+# separate version rather than a second rule inside v3 because v3's value is
+# isolating one change, and `notes/generic_v3_analysis_plan.md` was registered
+# before any run naming that one rule (#338).
+GENERIC_PROMPT_V4 = PROMPTS / "d4d_generic_arm_prompt_v4.md"
 CONDITION_PROMPTS = {"generic": GENERIC_PROMPT,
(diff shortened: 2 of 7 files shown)


realmarcin:
## Review pass

One finding, filed as #340 and fixed. It turned out to hide two more.

**#340 — `condition_of()` returned `None` for every v3 and v4 run.** Its
hardcoded chain knew v1, v2 and tuned, and `d4d_generic_arm_prompt_v3.md` does
not contain `d4d_generic_arm_prompt.md` as a substring — the `_v3` sits between
`prompt` and `.md`. So a v3 run fell through every branch and reported nothing.

Silent, and squarely on the rerun path: anything asking which condition produced
a record could not tell a v3 run from an unrecognisable one. Pre-existing for v3;
this PR would have added v4 with the same gap.

Now derived from `CONDITION_PROMPTS` rather than restated — the lesson
`cli/api.py` already carries from when the condition list was written out three
times and left `generic_v2` unreachable.

### Two more, found while testing the fix

**Prompt entries are mappings, not strings.** Provenance records
`{"path": ..., "sha256": ...}`. The previous code matched only because
`str(dict)` happens to contain the path — which also drags every other value in.
The field is now read.

**Registry order matters when several prompts are listed.** `tuned` already
lists two. Dict-order iteration would return whichever matched first, so a run
naming both the v1 base and v4 could be reported as `generic` — silent and wrong
in the same way #340 was. Walked longest-name-first, and now tested: mutation
showed the ordering was otherwise unexercised, since no provenance on disk lists
two generic prompts today.

### Verified against the corpus

```
2026-07-31_claude-opus-5-generic-v2_rep1   -> generic_v2
2026-07-31_claude-opus-5-api-generic_rep1  -> generic
2026-07-28_claude-opus-5-generic_rep1      -> None
```

The last is correct: that series' provenance records `prompts: None`, predating
prompt recording entirely. A property of those records, not of this function —
worth knowing before anyone reads the v1 baseline's condition from provenance.

`1266 passed, 3 skipped`. 11 mutants across both commits, no survivors.

PR #341, opened 2026-08-05
Derive the evaluation count instead of restating it (#315)

### LINKED ISSUE #315: The semantic evaluation count is 20, not 12; pin it to the canonical set rather than restating it
From the Codex review (`notes/codex_rerun_readiness_review_2026-08-04.md`), and it is right.

I told the user **12** semantic evaluations. That figure assumed three canonical projects. If the rerun gives all five projects a canonical record, it is:

```
5 projects × {full, core} × {rubric10_semantic, rubric20_semantic} = 20
```

## This arithmetic has now been wrong three times, in different directions

| stated | assumed | actually |
|---|---|---|
| **8** (#287's table) | 4 projects × 2 rubrics | full records only — unstated |
| **6** (after VOICE failed selection) | 3 projects × 2 rubrics | still full-only |
| **12** | 3 projects × 2 variants × 2 rubrics | correct *only* while VOICE and VOICE_PEDIATRIC have no canonical record |
| **20** | 5 projects × 2 variants × 2 rubrics | correct after a successful rerun |

Every one of those was a number written into a note and then restated, rather than derived. The count depends on how many projects end up with a canonical record, which is not known until after `d4d runs select --execute` — and for VOICE it is genuinely uncertain, because `related_to` may recur (#313).

## Fix

Stop stating the count. Derive it:

```
d4d runs canonical --paths-only | wc -l     # records
× 2 rubrics                                 # evaluations
```

`--paths-only` already emits one line per full and core record, so the evaluation count is `2 × that`, whatever the rerun produces. Any scoping note should quote the command, not a number.

Also worth reconciling while here: `notes/generic_v3_analysis_plan.md` registers a **four-project, three-replicate** v2-vs-v3 comparison, while NEXT_TASKS §9 says `VOICE_PEDIATRIC` is generated once under the winning config because it has no v1/v2 baseline. Those are consistent — pediatric sits outside the comparison — but neither document says so explicitly, and the five-project rerun makes the ambiguity live.

realmarcin:
Closes #315.

## The count has been wrong three times out of four

| stated | assumed | actually |
|---|---|---|
| **8** | 4 projects × 2 rubrics | full records only — unstated |
| **6** | 3 projects × 2 rubrics | still full-only |
| **12** | 3 × 2 variants × 2 rubrics | correct *only* while VOICE and VOICE_PEDIATRIC have no canonical record |
| **20** | 5 × 2 variants × 2 rubrics | correct *only* after a successful rerun |

Every one was typed into a note and then quoted from that note. The count is a
function of how many projects end up with a canonical record — not knowable
until `d4d runs select --execute` has run, and for VOICE genuinely uncertain
because `related_to` may recur (#292).

## The fix

`evaluation_plan.plan()` enumerates the canonical set and pairs each record with
each rubric; `d4d evaluate plan` prints it.

```
$ d4d evaluate plan
AI_READI/full/rubric10_semantic   data/.../AI_READI_d4d.yaml
...
12 evaluations = 3 projects (AI_READI, CHORUS, CM4AI) x 2 variants (core, full) x 2 rubrics
```

Today 12; after a successful rerun, 20 — with no number edited anywhere.
`summarise()` prints the derivation beside the total, because a bare count being
quoted onward is the failure mode itself.

`plan()` **raises** `NothingSelected` rather than returning an empty list.
"Zero evaluations" and "selection has not been run yet" are different states,
and a plan of length zero reads as the first while usually meaning the second.

## The tests do not assert a number

Asserting `len(plan()) == 12` would pass today, recreate this exact defect inside
the suite, and break on the day the rerun succeeds — which is the day it should
quietly become 20. They assert instead that the plan *follows the marks*: mark
another project, get its evaluations.

## Three faults found while writing them

- The first fixture wrote a bare `canonical` key, which `canonical_runs` does
  not recognise, so **every assertion passed against an empty plan**. It now has
  the shape the resolver reads.
- A guard asserting "no test here asserts the live count" listed the forbidden
  patterns in its own source, so it could never pass. Removed rather than
  contorted.
- Mutation showed `plan`'s `sorted()` was **inert** — `canonical_runs` already
  returns a sorted dict. Ordering is now tested against an unsorted mapping, so
  `plan` owns it rather than inheriting an implementation detail.

11 tests, mutation-checked with 6 reintroductions, no survivors.
`1277 passed, 3 skipped`.

### DIFF
--- a/src/data_sheets_schema/cli/evaluate.py
+++ b/src/data_sheets_schema/cli/evaluate.py
@@ -190,10 +190,18 @@ def plan_cmd(config, paths_only):
     """
     from data_sheets_schema.evaluation_plan import (NothingSelected, plan,
                                                     summarise)
+    from data_sheets_schema.runs import AmbiguousCanonical
     try:
         evaluations = plan(config=config)
     except NothingSelected as exc:
         raise click.ClickException(str(exc))
+    except AmbiguousCanonical as exc:
+        # The state the rerun creates: marks under both the old and the new
+        # config until re-selection settles. `plan` is right to propagate rather
+        # than choose one (#308); rendering it is this boundary's job (#342).
+        raise click.ClickException(
+            f"{exc} Pass --config to say which configuration you mean, or "
+            "re-run `d4d runs select --execute` to settle the mark.")
 
     if paths_only:
         for path in dict.fromkeys(str(e.path) for e in evaluations):
--- a/src/data_sheets_schema/cli/evaluate.py
+++ b/src/data_sheets_schema/cli/evaluate.py
@@ -172,3 +172,36 @@ def llm(file, project, method, rubric, output_dir):
         sys.exit(1)
     finally:
         sys.argv = old_argv
+
+
+@evaluate.command("plan")
+@click.option("--config", default=None,
+              help="Restrict to one run config (label prefix).")
+@click.option("--paths-only", is_flag=True,
+              help="One record path per line, for piping into a sweep.")
+def plan_cmd(config, paths_only):
+    """What a semantic evaluation sweep would cover, derived from the canonical set.
+
+    The count has been stated four times and been wrong three of them (#315),
+    because it depends on how many projects end up with a canonical record —
+    which is not known until `d4d runs select --execute` has run. This derives
+    it rather than restating it, and prints the derivation alongside the number
+    so a bare count is harder to quote onward.
+    """
+    from data_sheets_schema.evaluation_plan import (NothingSelected, plan,
+                                                    summarise)
+    try:
+        evaluations = plan(config=config)
+    except NothingSelected as exc:
+        raise click.ClickException(str(exc))
+
+    if paths_only:
+        for path in dict.fromkeys(str(e.path) for e in evaluations):
+            click.echo(path)
+        return
+
+    for evaluation in evaluations:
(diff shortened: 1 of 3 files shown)


realmarcin:
## Review pass

One finding, filed as #342 and fixed.

**`d4d evaluate plan` crashed on the two-mark state the rerun creates.** It
rendered `NothingSelected` as a message and let `AmbiguousCanonical` through as a
bare traceback, with no mention of `--config`:

```
exit code: 1
exception: AmbiguousCanonical
output:
```

A project marked canonical under both the old and the new config is exactly
where the corpus sits between generating v3 and re-selecting.

`plan()` propagating rather than choosing a config is right — choosing is what
#308 stopped it doing. The gap was only at the CLI boundary, where every other
failure in the command was already a message. I checked `d4d runs canonical` for
the same gap; it already handles it.

Three tests, one per outcome, including a resolvable corpus exiting zero so the
guard cannot turn every invocation into an error.

## Noted, not filed

`notes/codex_rerun_readiness_review_2026-08-04.md` still contains "the 12
evaluations". It is a dated record of a review that *raised* the ambiguity rather
than guidance to follow, so correcting it would edit history to remove the
question this PR answers. Flagging so the decision is on the record.

`1280 passed, 3 skipped`. 9 mutants across both commits, no survivors.

PR #343, opened 2026-08-05
Classify VOICE's failure modes (#292); make both coverage options runnable (#287)

realmarcin:
Addresses #292 and closes the open half of #287.

## #292 — three failures, three owners

All three 2026-07-31 VOICE replicates fail `related_datasets`, and
`linkml-validate` reports them as three unrelated errors. Whether the rerun
fixed anything depends on **which mode** recurs, and they have different owners:

| mode | what it means | addressed by |
|---|---|---|
| `inline_target` | object where the schema declares a string. #297: LinkML cannot express string-or-inline, so only validation catches it | **generic-v4**; recurrence retires that rule |
| `aliased_type` | a DataCite spelling the enum declares as an `alias` (#223) | nothing — `normalise_enum_aliases` already rewrites these on the write path, so recurrence means the normaliser did not run |
| `unknown_type` | `related_to`, an alias of nothing | nothing — a real generation failure |

**I checked whether `related_to` genuinely has no equivalent.** The vocabulary is
DataCite's 36 relation types and is deliberately specific — there is no generic
"is related to" — so leaving it to fail is correct rather than lazy, and adding
it would break the DataCite alignment in #223.

`d4d evaluate related-datasets` classifies them. The canonical set reports **0
defects across 6 records**; the three replicates reproduce one mode each.

**rep3 carries three aliased values, not the one the validator stops at.** A
checker that stopped early would have called it fixed as soon as the ordering
changed.

## #287 — both options runnable, and a recommendation

```
d4d evaluate plan                    # 12 today, 20 after a five-project rerun
d4d evaluate plan --all-replicates   # 36 today, 60 after
```

Neither number is written anywhere; both derive from the canonical marks (#315).

`notes/semantic_evaluation_scoping_2026-08-05.md` recommends **canonical-only**,
and the reason is an asymmetry rather than a preference: if the cheap run turns
out to need a variance estimate, the replicates are still on disk and
`--all-replicates` runs them later. Calls spent now cannot be unspent.

Replicate spread does not rescue between-config power. #169 measured the
between-config delta at −2.9 against a replicate sd of 10.9, needing ~110
projects — the limit is the project sample, not the replicate count. Three times
the calls does not move the comparison the study is for.

The note also records what *would* change the recommendation: a claim about
within-config stability makes `--all-replicates` the right run and 36 the number.

## A defect this surfaced

`summarise` printed `36 evaluations = 3 projects x 2 variants x 2 rubrics` — a
derivation multiplying to **12** beside a total of **36**. That is worse than no
derivation, and it is precisely the failure #315 exists to prevent, so it gained
the replicate factor and a test asserting the product equals the total in both
modes.

34 tests, mutation-checked with 10 reintroductions, no survivors.
`1300 passed, 3 skipped`.

### DIFF
--- a/notes/semantic_evaluation_scoping_2026-08-05.md
+++ b/notes/semantic_evaluation_scoping_2026-08-05.md
@@ -1,0 +1,52 @@
+# Semantic evaluation coverage: the decision, and why
+
+Closes the open half of #287. Both preconditions it named are met: canonical
+records are marked for three of four projects, and the prompt config is ready to
+choose (v3 stamps itself correctly since #337; v4 exists since #338).
+
+## Both options are runnable
+
+```
+d4d evaluate plan                    # one record per project
+d4d evaluate plan --all-replicates   # every replicate of the canonical config
+```
+
+Today that is 12 against 36. Neither number is written anywhere — both are
+derived from the canonical marks (#315), and both move on their own as projects
+gain records. After a successful five-project rerun they become 20 and 60.
+
+## Recommendation: canonical-only
+
+**What replicate spread buys.** A within-config variance estimate: how much two
+runs of the same prompt on the same bundle differ. That is a real quantity and
+the fitness axis already reports it.
+
+**What it does not buy.** Any improvement in resolving *between-config*
+differences. #169 measured that directly — the between-config delta was −2.9
+against a replicate standard deviation of 10.9, and ~110 projects would be
+needed. More replicates of the same four projects narrow the standard error of
+each arm's mean, but the limit here is the project sample, not the replicate
+count. Three times the calls does not move the comparison the study is for.
+
+**What it costs.** 3x, and the semantic path is the expensive one — an LLM judge
+per record per rubric.
+
+**The asymmetry that decides it.** If the canonical-only result turns out to
+need a variance estimate, the replicates are still on disk and
+`--all-replicates` runs them later. The reverse is not true: calls spent now
+cannot be unspent. Run the cheap one first.
+
+## What would change the recommendation
+
+If the manuscript claims a *between-config* semantic difference, neither option
+supports it at four projects and the honest answer is to not make the claim.
+Widening the project set would; widening the replicate set would not.
+
+If the claim is about within-config stability — "the same prompt produces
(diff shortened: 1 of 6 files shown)


realmarcin:
## Review pass

One finding, filed as #344 and fixed. It turned out to hide a second, and my
first test for that was wrong.

**#344 — `--all-replicates` enumerated records that do not validate.** Four of
the nine siblings on the 2026-07-31 corpus:

```
AI_READI rep2, rep3    invalid
CM4AI    rep1, rep3    invalid
```

The canonical mark exists *because* that replicate validates — the criterion is
"full and core both validate against the current schema" — and siblings inherit
none of it. Scoring an invalid record and pooling it into a within-config
variance estimate measures validation failure as quality variance, which is the
opposite of what #287 wants the estimate for. Gated on
`runs.validation_status`, so nothing is re-validated; `unverified` is excluded
too, because absence of evidence is not validity.

Exclusions are **named, not counted**: "4 excluded" invites the reader to assume
they were the same kind of thing.

**The second defect.** With exclusions the plan stops being a product, and
`summarise` went on printing `3 projects x 2 variants x 2 rubrics x 3
replicates` beside a total of **20** — a factorisation multiplying to 36. That
is precisely the failure #315 exists to prevent, and worse than no
factorisation because it reads as a check that passed. It now prints the product
only when the product equals the total.

**My first test for that asserted the wrong thing.** It excluded rep2 and rep3
from the fixture's single project — which leaves an *even* grid and is still a
legitimate product. Raggedness needs an **uneven** exclusion, so the test now
builds two projects keeping different numbers of replicates, with a complement
asserting an even exclusion is still reported as a product.

## The recommendation changes with it

The scoping note said the widened run costs 3x. With four replicates excluded it
is **1.7x** — 12 against 20, not 12 against 36. That weakens the cost argument
for canonical-only, though the substantive one is untouched: replicate spread
still does not rescue between-config power (#169), and the asymmetry still holds
— the replicates stay on disk and can be run later.

`1304 passed, 3 skipped`. 15 mutants across both commits, no survivors; the last
needed the summary test tightened from "reports an exclusion" to "names which
one".

PR #405, opened 2026-08-08
Fix two verifiable-matcher defects reporting present values as invented (#404)

### LINKED ISSUE #404: d4d evaluate verifiable: trailing-slash URLs and abbreviated months reported ungrounded (91.4% should be 97.1%)
`d4d evaluate verifiable` reports values as ungrounded that are plainly present in the declared bundle. Two independent matcher defects in `src/data_sheets_schema/verifiable.py` account for **115 of the 175** ungrounded values on the current canonical sweep (`2026-08-07_claude-opus-5-claudecode-generic-v3`, 15 records, 2044 checkable values).

Corrected, the corpus moves from **91.4% → 97.1% grounded**.

This matters beyond the headline number: the bias is not uniform. It penalises records that cite many URLs and that normalise dates to ISO — which is what a *better* record does. Any ranking of replicates or arms read off these rates is distorted.

## Defect 1 — a URL written with a trailing slash can never be grounded (104 values)

`normalise()` strips the trailing `/` from the claim (line 254) but `normalise_bundle()` does not strip it from the bundle (line 370), and `_CONTINUES["url"]` includes `/` (line 326). So the guard `(?!/)` in `grounded_in()` rejects the match whenever the source document wrote the URL with its trailing slash.

```
physionet.org/content/b2ai-voice/3.1.0   grounded=False   present_with_slash=True
healthdatanexus.ai                       grounded=False   present_with_slash=True
```

Note the two sides are asymmetric in exactly the way the comment at line 247 warns against — the same class of bug it was written to fix.

## Defect 2 — abbreviated month names are not generated (11 values)

`renderings("iso_date", ...)` emits only full month names. PhysioNet, Dataverse and most other sources abbreviate:

```
renderings('iso_date','2025-08-18') -> ['2025-08-18','august 18, 2025', ...]
bundle contains                     -> 'Aug. 18, 2025'   (3 occurrences)
```

## The guard test encodes the bug

`tests/test_evaluation/test_verifiable.py::TestAgainstTheCorpus::test_the_known_voice_finding_is_still_detected` asserts that VOICE rep2 states three release dates "absent from its bundle in any format", and describes this as "the check's first real positive".

All three are in the bundle:

| asserted fabricated | actually in `VOICE_preprocessed.txt` | occurrences |
|---|---|---|
| `2025-08-18` | `Aug. 18, 2025` | 3 |
| `2025-12-16` | `Dec. 16, 2025` | 4 |
| `2025-12-17` | `Dec. 17, 2025` | 1 |

So the check has no confirmed true positive on that record, and the test protects the defect rather than the behaviour. It needs replacing with a fixture whose absence is constructed rather than assumed.

## Proposed fix

1. In `grounded_in()`, allow an optional trailing `/` for `url` before the continuation guard: `(?<!before)rendering/?(?!after)`. Regex backtracking preserves the "not inside a longer token" guarantee — `example.com/a` still must not match inside `example.com/a/b`.
2. In `renderings()`, generate abbreviated month forms (`aug.`, `aug`, `sept.`, `sep.`, …) alongside the full names.
3. Replace the false-positive corpus test with a constructed-absence fixture, and raise the corpus floor in `test_no_record_falls_below_the_measured_floor`.

## Scope

Every run measured with `d4d evaluate verifiable` to date is affected. Rates published from it should be treated as lower bounds until re-run.

realmarcin:
Closes #404.

`d4d evaluate verifiable` reported **175 of 2044** values ungrounded on the current canonical sweep (`2026-08-07_claude-opus-5-claudecode-generic-v3`, 15 records). **115 of those are present in the declared bundle** and were missed by the matcher. Corrected, the corpus moves **91.4% → 96.4%**.

The bias was not uniform, which is why this matters beyond the headline: it penalised records that cite many URLs and normalise dates to ISO — what a *better* record does. Any ranking of replicates or arms read off these rates was distorted.

## Defect 1 — trailing-slash URLs (104 values)

`normalise()` strips the trailing `/` from the claim; `normalise_bundle()` normalises the bundle wholesale and keeps it; `_CONTINUES["url"]` includes `/`. So `(?!/)` rejected every URL a source wrote with its trailing slash — the same claim/bundle asymmetry the comment at line 247 was written to fix.

`grounded_in` now allows an optional trailing `/`, for `url` only.

**This does not weaken the boundary guarantee.** Against `example.org/a/b`, the branch that consumes the slash is rejected by the lookahead (`b`), and the branch that backtracks past it is rejected by the slash. Both fail, which is the required answer. Both directions are tested.

## Defect 2 — abbreviated months (11 values)

`renderings()` emitted only full month names; PhysioNet writes `Aug. 18, 2025`, Dataverse `Aug 18, 2025`. It now emits abbreviated forms too, with `sept.` alongside `sep.` and no spurious `may.`.

Widening the month form cannot equate two different dates — a rendering still has to carry the same day and year — and there is a test asserting no rendering names a different day or year.

## The guard test encoded the bug

`test_the_known_voice_finding_is_still_detected` asserted VOICE rep2 states three release dates "absent from its bundle in any format", describing them as the check's first real positive. All three are in the bundle:

| asserted fabricated | actually present as | occurrences |
|---|---|---|
| `2025-08-18` | `Aug. 18, 2025` | 3 |
| `2025-12-16` | `Dec. 16, 2025` | 4 |
| `2025-12-17` | `Dec. 17, 2025` | 1 |

The check therefore had **no confirmed true positive** on that record. The test is inverted into a regression guard, and `test_a_date_absent_by_construction_is_still_reported` adds the true-positive case with absence *constructed* rather than assumed of a real bundle — so it cannot quietly become a statement about the matcher's blind spots again.

Corpus floor raised 0.70 → 0.80 (lowest record on that label is now CHORUS rep3 at 0.864).

## Verification

- `tests/test_evaluation/test_verifiable.py`: **57 passed** (7 new).
- Full suite: 11 failures **before and after**, byte-identical sets — confirmed by stashing this diff and re-running. All 11 are pre-existing and unrelated (`test_canonical_resolver`, `test_d4d_pair_consistency`, `test_rocrate/test_provenance`, `test_semantic_exchange`). This change causes none of them and fixes none of them.
- Blast radius: `MONTHS`, `renderings` and `grounded_in` have no importers outside `verifiable.py` and its tests.

## Not in this PR

Re-running the check surfaced **74 remaining ungrounded values** that need triage, including two further matcher leads I deliberately left out to keep this change matched to the filed issue:

- **DOIs with a trailing colon** (4 values, e.g. `10.60775/fairhub.1):`) — `normalise` strips `.,;)]` but stops at `:`, so the `)` is never reached. Same punctuation-asymmetry class as defect 1, one character to fix.
- **Ordinal day forms** (2 values) — the bundle writes `January 17th, 2023`; `renderings` generates no ordinal variant.
- **38 URLs and 30 counts** that need inspection to tell genuine overstatement from further matcher gaps.

Happy to fold the first two in here if preferred; I kept them out so the diff matches what #404 describes and was reviewed against.

## Scope

Every run measured with `d4d evaluate verifiable` to date is affected. Published rates should be treated as lower bounds until re-run.

### DIFF
--- a/src/data_sheets_schema/verifiable.py
+++ b/src/data_sheets_schema/verifiable.py
@@ -372,20 +372,28 @@ _PRECEDES = {
 def grounded_in(kind: str, rendering: str, haystack: str) -> bool:
     """Is this rendering present as a whole token, not inside a longer one?
 
-    URLs may carry a trailing slash on either side. `normalise` strips it from
-    the claim but the bundle is normalised wholesale and keeps it, and `/`
-    continues a URL, so `(?!/)` rejected every URL a source wrote with its
-    trailing slash -- 104 values on the 2026-08-07 sweep, all of them present
-    (#404). The optional `/?` restores the symmetry that `normalise` intends.
+    A locator may carry a trailing slash on either side. `normalise` strips it
+    from the claim but the bundle is normalised wholesale and keeps it, and `/`
+    continues both a URL and a DOI, so `(?!/)` rejected every one a source
+    wrote with its trailing slash -- 104 values on the 2026-08-07 sweep, all of
+    them present (#404). The optional `/?` restores the symmetry that
+    `normalise` intends.
+
+    Applied to `doi` as well as `url`, because `normalise` ends in
+    `rstrip("/")` for every kind, so the asymmetry it creates is not specific
+    to one. No DOI in the corpus currently trips it -- 0 of 245 measured on the
+    2026-08-07 sweep -- so this half is latent, fixed because the defect is the
+    same one and finding it twice is worse than fixing it once.
 
     It does not weaken the boundary guarantee, because the guard still applies
     after it: against `example.com/a/b`, `/?` first consumes the slash and the
     lookahead rejects `b`, then backtracks to empty and the lookahead rejects
-    the slash. Both branches fail, which is the required answer.
+    the slash. Both branches fail, which is the required answer. A rendering
+    that already ends in `/` is unaffected, since `/?` may match empty.
     """
     after = _CONTINUES.get(kind, r"\w")
     before = _PRECEDES.get(kind, r"\w")
-    optional = "/?" if kind == "url" else ""
+    optional = "/?" if kind in ("url", "doi") else ""
     pat = re.compile(
         rf"(?<!{before}){re.escape(rendering)}{optional}(?!{after})")
     return bool(pat.search(haystack))
--- a/src/data_sheets_schema/verifiable.py
+++ b/src/data_sheets_schema/verifiable.py
@@ -197,6 +197,27 @@ MONTHS = ("January", "February", "March", "April", "May", "June", "July",
           "August", "September", "October", "November", "December")
 
 
+# Sources abbreviate, and they do not agree on how. PhysioNet writes "Aug. 18,
+# 2025" and "Sept. 3, 2025"; Dataverse writes "Aug 18, 2025"; AP style leaves
+# March, April, May, June and July unabbreviated. Generating only the full name
+# reported 11 dates as invented on the 2026-08-07 sweep that the bundle plainly
+# carried -- including all three of the dates the corpus test called this
+# check's "first real positive" (#404).
+#
+# Widening this cannot equate two different dates: a rendering still has to
+# carry the same day and year, so the month form is the only thing varying and
+# every form here names the same month.
+def _month_forms(m: int) -> list[str]:
+    full = MONTHS[m - 1].lower()
+    forms = [full]
+    if len(full) > 3:                      # "may" is already its abbreviation
+        stem = full[:3]
+        forms += [f"{stem}.", stem]
+        if full == "september":            # "sept." is commoner than "sep."
+            forms += ["sept.", "sept"]
+    return list(dict.fromkeys(forms))
+
+
 def renderings(kind: str, value: str) -> list[str]:
     """Every spelling a source document might plausibly use for one token.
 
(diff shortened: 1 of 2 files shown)


realmarcin:
## Rebased on main, and adversarially reviewed

Rebased onto current `main` (clean, no conflicts) now that #411 has landed and fixed the three schema-drift failures that were making this PR's CI red. Verified the corpus these tests depend on — `2026-07-28_claude-opus-5-generic_rep{1,2,3}` — survived #411's archiving untouched.

### One real finding, fixed in `5ac7bf20` — filed as #413

**The same trailing-slash asymmetry exists for `doi`, which the original fix did not cover.** `normalise()` ends in `rstrip("/")` for *every* kind, not just URLs; `normalise_bundle()` strips none from the bundle; and `_CONTINUES["doi"]` treats `/` as a continuation. So:

```
claim  https://doi.org/10.1234/x  ->  10.1234/x
bundle "see https://doi.org/10.1234/x/ here"    grounded=False
```

**Latent, not active**: 0 of 245 DOIs across the 2026-08-07 sweep trip it, and the corpus figure is unchanged at **1970/2044** with the fix applied. Fixed anyway — identical defect, identical function, one tuple element — because a half-fixed asymmetry is how this comes back. Three tests added, including that the boundary guarantee still holds for DOIs.

### Two things I suspected and measured instead of assuming — both non-findings

- **Performance.** Widening the month forms takes a date from 8 renderings to 10–21, each a regex search over an ~80k-token bundle. A microbenchmark looked alarming (12.9s for 200 repeated checks), but that repeats one date; end-to-end `d4d evaluate verifiable` over all 15 records is **8.3s**. Not a concern.
- **A latent trap in `grounded_in`.** I expected a rendering already ending in `/` to produce a pattern demanding `//`. It does not — `/?` may match empty, so `grounded_in("url", "ex.org/a/", …)` grounds correctly. Added a test pinning it rather than leaving the reasoning in a comment.

### Re-verified after the rebase

- Floor still valid: lowest record is CHORUS rep3 at **0.864** against the asserted 0.80.
- The docstring occurrence counts in the inverted corpus test still hold: `Aug. 18, 2025` ×3, `Dec. 16, 2025` ×4, `Dec. 17, 2025` ×1.
- Full suite: **1351 passed, 4 skipped, 0 failed.**

Issue #410, opened 2026-08-08
Composition-path mapping rows collide with the class's own slots (d4d:Dataset/id → d4d:dataAnomalies)


realmarcin:
Found while fixing #407. Not fixed there, because the fix changes rows that predate that work.

`_map_composition_paths` names its subject after the **last segment** of the composition path:

```python
d4d_slot=path.split('.')[-1],
```

So walking `Dataset.anomalies → DataAnomaly` and reaching `DataAnomaly.id` emits a row whose subject is `d4d:Dataset/id` — which is *also* the identifier of `Dataset`'s own `id` slot. The row then reads:

> `d4d:Dataset/id` closeMatch `d4d:dataAnomalies`

which is false. The Dataset's own `id` does not closely match an anomaly. The information that distinguishes them — that this row is about `anomalies.id`, not `id` — survives only in the free-text `structural_notes` column, which nothing consumes.

All 16 composition rows in the committed mapping have this shape:

```
d4d:Dataset/anomalies         --closeMatch-> d4d:dataAnomalies | Composition path: anomalies
d4d:Dataset/anomaly_details   --closeMatch-> d4d:dataAnomalies | Composition path: anomalies.anomaly_details
d4d:Dataset/id                --closeMatch-> d4d:dataAnomalies | Composition path: anomalies.id
d4d:Dataset/name              --closeMatch-> d4d:dataAnomalies | Composition path: anomalies.name
d4d:Dataset/description       --closeMatch-> d4d:dataAnomalies | Composition path: anomalies.description
d4d:Dataset/notes             --closeMatch-> d4d:dataAnomalies | Composition path: anomalies.notes
d4d:Dataset/source_caveats    --closeMatch-> d4d:dataAnomalies | Composition path: anomalies.source_caveats
d4d:Dataset/used_software     --closeMatch-> d4d:dataAnomalies | Composition path: anomalies.used_software
```

and the same eight for `DataSubset`.

Only the first of each group (`/anomalies`) is a correct subject. The other seven are the generic slots every `DatasetProperty` inherits, re-attributed to the parent class.

## Why it surfaced now

It did not change behaviour; it changed *volume*. #385 added `notes` and `source_caveats` to the base class, so the walk produced two more rows per class. That is what made `test_regeneration_invents_nothing` fail, and #407 resolved it by accepting the new rows as consistent with the twelve already committed. Consistent, but consistently wrong.

[...]

Issue #409, opened 2026-08-08
Twelve unexplained BOM-prefixed D4D_*.yaml schema subsets at the repository root


realmarcin:
Twelve files sit untracked at the repository root:

```
D4D_Collection.yaml       D4D_Maintenance.yaml     D4D_Motivation.yaml
D4D_Composition.yaml      D4D_Metadata.yaml        D4D_Preprocessing.yaml
D4D_Data_Governance.yaml  D4D_Minimal.yaml         D4D_Uses.yaml
D4D_Distribution.yaml     D4D_Ethics.yaml          D4D_Human.yaml
```

They are **not** copies of the schema modules in `src/data_sheets_schema/schema/`, despite sharing their names:

```
D4D_Collection.yaml   root=6306 bytes  schema=9017 bytes  identical=NO
D4D_Ethics.yaml       root=4213        schema=5985        identical=NO
D4D_Minimal.yaml      root=4644        schema=1468        identical=NO
```

Each begins with a UTF-8 BOM and declares itself a subset:

```yaml
id: https://w3id.org/bridge2ai/data-sheets-schema/collection
name: d4d-collection
title: "D4D Collection Subset"
description: >
  A LinkML schema subset for capturing how data was collected, ...
```

The BOM suggests they arrived from an external editor or download rather than from `gen-project`. `CLAUDE.md` states module files live in `src/data_sheets_schema/schema/` and not elsewhere, so root is not their home under the documented layout.

They have been in the tree at least since 2026-08-07 and possibly longer. Nothing in `src/` or the Makefile references a root-level `D4D_*.yaml`.

Left alone rather than committed or deleted while landing `provenance-prompt-hashes-agentic`, because their provenance is unknown and deleting someone's unsaved work is not a reviewer's call.

Needs: whoever put them there to say what they are. Then either home them (a `subsets/` directory?), delete them, or add them to `.gitignore` if they are scratch. Until then every `git status` in this repository has twelve lines of noise, which is how a real untracked file gets missed.

realmarcin:
## Investigated — these are stale pre-reorganisation copies, not a parallel design

Four things establish it:

**1. Never in git.** `git log --all --diff-filter=AD -- D4D_Collection.yaml` is empty. No history, so no recovery if deleted.

[...]

Issue #385, opened 2026-08-07
notes escape hatch misused 4:1 — prose defaults to notes over description; structured slots bypassed; provenance commentary needs a channel


realmarcin:
Full audit of all 155 `notes` values in the schema-2 CHORUS + AI-READI rep1 records (49 + 106):

| class | CHORUS | AI-READI |
|---|---|---|
| (a) well-placed | 6 | 16 |
| (b) misplaced — a fitting slot exists | 27 | 43 |
| (c) redundant — restates a sibling | 11 | 33 |
| (d) evidence/provenance commentary | 5 | 14 |

Three distinct failure modes:
1. **`notes` beat `description` as the default prose home.** All ten `DistributionFormat` objects and five `DatasetRelationship` objects consist of *nothing but notes* while inherited `name`/`description` sit null. 34 notes are ordinary descriptive prose.
2. **Structured slots bypassed**: `Creator.affiliations` empty in every Creator of both records while 26 notes carry affiliation prose (often with ROR + ORCID — exactly what the slot and #378's enrichment want); `FundingMechanism.grants` unused with grant numbers/amounts in notes; CHORUS `Creator.name` empty (name only in notes) and `principal_investigator: 'true'` despite the slot description saying "never a yes/no flag".
3. **Evidence commentary is genuinely homeless** (19 notes: source conflicts, transcription provenance, unanswered healthsheet questions) — valuable, but mixed indistinguishably with content notes; a consumer cannot separate "extra fact" from "trust caveat about the sibling field".

Proposed fixes (pre-VOICE/CM4AI decision — changes instructions, so sweep-uniformity applies):
- Instruction: `description` is the default home for prose; `notes` only for what `description` cannot hold; populate structured slots (affiliations, grants, name) before reaching for notes.
- Decide whether evidence commentary gets its own slot (e.g. `evidence_notes`/`source_caveats`) — mode 3 is arguably a feature worth first-classing for the manuscript's evidence-boundary story.
- Repair/enrichment could mechanically lift mode-2 content (affiliation/ROR/ORCID, grant numbers) into their slots.

Full audit with verbatim examples in the session transcript; summary findings in the sweep findings file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

realmarcin:
[...]

PR #411, opened 2026-08-08
Fix the five failing tests: stale schema artifacts (#407) and superseded run directories (#408)

### LINKED ISSUE #407: Three tests fail on main: pair-consistency slot count (76 vs 78) and SSSOM mapping no longer regenerates
Three tests fail on a clean checkout of `main`, independent of any branch. Confirmed by stashing all working-tree changes and running them against HEAD.

## 1. `test_d4d_pair_consistency::test_schema_derives_identity_and_projected_slots`

```
AssertionError: 76 != 78
```

The test pins the number of schema-identical shared slots between `Dataset` and `CoreDataset` at 76. `SchemaView` now derives 78. Either two slots were legitimately added to both classes and the constant is stale, or two slots drifted into alignment unintentionally. The test cannot tell the difference, which is the deeper problem — a bare count does not say *which* slots, so a regression that removes one slot and adds another would pass.

Worth replacing the literal with a derived assertion plus an explicit list of the slots, so a change to the set is visible in the diff.

## 2 & 3. `test_structural_mapping_drift` (both tests)

The committed SSSOM mapping no longer regenerates from its inputs:

```
10 row(s) in the committed file that regeneration does not produce:
    d4d:CoreDataset            --skos:exactMatch->  schema:Dataset
    d4d:CoreDatasetCollection  --skos:exactMatch->  schema:Dataset
    d4d:CoreDistribution       --skos:exactMatch->  schema:DataDownload
    d4d:DataSubset             --skos:exactMatch->  schema:Dataset
    d4d:DatasetCollection/resources   --skos:exactMatch->  schema:hasPart
    d4d:File/file_type         --skos:exactMatch->  d4d:fileType
    d4d:FileCollection/collection_type --skos:exactMatch-> d4d:collectionType
    d4d:FileCollection/file_count      --skos:exactMatch-> d4d:fileCount
    d4d:FileCollection/resources       --skos:exactMatch-> schema:hasPart
    d4d:FileCollection/total_bytes     --skos:exactMatch-> dcat:byteSize

4 row(s) regeneration produces that the committed file lacks:
    d4d:DataSubset/notes           --skos:closeMatch->  d4d:dataAnomalies
    d4d:DataSubset/source_caveats  --skos:closeMatch->  d4d:dataAnomalies
    d4d:Dataset/notes              --skos:closeMatch->  d4d:dataAnomalies
    d4d:Dataset/source_caveats     --skos:closeMatch->  d4d:dataAnomalies
```

The test's own message states the stake: "A mapping nobody can rebuild is a mapping nobody can safely change (#234)."

The 10 missing rows look hand-added (the Core* classes and the FileCollection/File property mappings), so regeneration would drop real curation. The 4 extra rows look like a generator change that now maps `notes` and `source_caveats` to `d4d:dataAnomalies` — which reads wrong on its face: `notes` is not an anomaly.

Needs a decision on which side is authoritative before either is regenerated.

## Why this is filed rather than fixed

Found while reviewing `provenance-prompt-hashes-agentic`, but present at `main` and unrelated to it: `src/` and `project/` are unmodified on that branch, so neither the schema nor the mapping inputs changed. Fixing them means deciding what the mapping should say, which is its own piece of work.

### LINKED ISSUE #408: Untracked superseded/failed run artifacts break the end-of-run gate locally and use two undocumented conventions
The working tree accumulates run directories and files that are neither current runs nor archived ones, using at least two undocumented suffix conventions. They are not committed, but they sit in `data/d4d_concatenated/` where the tooling and the tests read from, and one of them makes the repository's own end-of-run gate fail locally.

## The corrupt one

```
data/d4d_concatenated/claudecode_agent{,_core}/
  2026-08-06_claude-opus-5-1m-generic-v3-schema2_rep1.superseded-pre-stragglers/
```

`tests/test_rocrate/test_provenance.py::TestTheEndOfRunGate::test_the_whole_corpus_passes_the_gate` fails on it:

```
CHORUS/...superseded-pre-stragglers: artifacts changed after provenance was recorded: f
```

`TestHashUnification::test_the_corpus_carries_no_stale_verdicts_after_migration` fails on the same directory. Moving just these aside takes `tests/test_rocrate/test_provenance.py` from 2 failed to **43 passed**, so this single directory is the whole of that failure.

CI is unaffected — the directory is untracked, so a fresh checkout never sees it — but every local `make test` run reports two failures that have nothing to do with the change under test, which is how real failures get ignored.

## The full inventory

Five `.superseded-*` directories:

- `..._rep1.superseded-pre-notes` (agent, agent_core)
- `..._rep1.superseded-pre-stragglers` (agent, agent_core) ← the corrupt one
- `..._rep1.superseded-dual-writer` (agent_core)
- `..._rep2.superseded-dual-writer` (agent, agent_core)

Three `.failed-core-degeneration` files under `2026-08-05_claude-opus-5-generic-v3_rep1`:

- `AI_READI_d4d.yaml.failed-core-degeneration`
- `AI_READI_api_progress.json.failed-core-degeneration`
- `AI_READI_reasoning.jsonl.failed-core-degeneration`

One scratch run: `data/d4d_concatenated/claudecode_agent_core/2026-08-07_test_rep1/TESTPROJ_provenance.yaml`

## The point

`d4d runs archive` already exists — "Move runs out of analysis into data/ATTIC" — and `data/ATTIC/` already holds `superseded_2026-07-23` directories, so the convention exists and these did not use it. Two ad-hoc dotted suffixes grew up beside it with nothing describing what they mean, whether anything reads them, or when they can be deleted.

Suggested:

1. Archive or delete the corrupt `.superseded-pre-stragglers` pair, so `make test` is honest locally.
2. Put the rest through `d4d runs archive` if they are worth keeping, or delete them.
3. Either document the two suffix conventions where someone will find them, or drop them in favour of `runs archive`.

realmarcin:
Closes #407. Closes #408.

Takes `make test` from **5 failures to 0** — 1338 passed, 4 skipped. First fully green run in a while.

## #407 — three failures, one root cause

All three traced to the same change: `notes` and `source_caveats` were added to the base class in `f192c34f` (#385), and two committed artifacts were never updated to match.

**Pair consistency, 76 → 78.** Diffing the induced slot sets against the schema as it stood when 76 was written: `notes` and `source_caveats` added, nothing removed, `resources` still the only projection. The count was stale, not the schema wrong.

Replaced the bare count with the explicit 78-slot tuple, as #407 suggested. A count says nothing about *which* slots — a change that removed one and added another would have passed, and when it did fail it gave no clue what had happened. Added a companion test pinning `notes`, `source_caveats` and `description` as shared, so removing them from one class only fails with the cause in the test name.

**SSSOM — stale, not wrong.** The four new rows are *not* a name-similarity misfire; `_semantic_similarity("notes", "d4d:dataAnomalies")` is `0.0`. They come from `_map_composition_paths` walking `Dataset.anomalies → DataAnomaly` and emitting one row per reachable slot. The committed file already carried twelve rows of exactly that shape, including `d4d:Dataset/id`, `/name` and `/description` all mapped to `d4d:dataAnomalies`. Two new base-class slots produced two more per class.

So the file was stale — which is the diagnosis `test_regeneration_invents_nothing` already gives in its own docstring.

**I did not run `make gen-sssom-structural`.** A blind regeneration writes only the generator's 153 rows and would have silently dropped the 10 hand-added ones that `KNOWN_UNDERIVABLE` exists to protect. Instead: regenerated to scratch, merged with the hand-added rows, asserted the kept set is *exactly* `KNOWN_UNDERIVABLE` and that the column set is unchanged. 159 → 163 rows. Summary regenerated alongside.

`--check` still exits non-zero on those 10 rows. That is the documented state of #234/#295, not a regression, and the tests assert it deliberately.

## #408 — two failures, one directory

Both came from `2026-08-06_..._rep1.superseded-pre-stragglers`, whose artifacts changed after its provenance was recorded. `discover()` walks `data/d4d_concatenated`, so a run kept in place under a dotted suffix is still a run to every analysis path. Moving just that pair aside took `test_provenance.py` from 2 failed to **43 passed**, which confirmed the diagnosis before anything was moved.

It could not simply be committed either — committing a state the repo's own test rejects is not an option — so it sat untracked indefinitely, along with everything else using the same convention.

Moved through **`d4d runs archive`**, which already existed for this and which `data/ATTIC` already held nine directories from. Archive moves rather than renames, records a reason, and is reversed exactly by `d4d runs restore`. Four labels, CHORUS only, named explicitly rather than with `allow_partial_labels` — the guard was right to refuse, since one label holds a placeable record and archiving it is a deliberate choice, not collateral.

Also archived: the three `.failed-core-degeneration` files, under their original method/label path.

**One deletion:** `2026-08-07_test_rep1/TESTPROJ_provenance.yaml` — a smoke-test record for a fake project, written three minutes before `d4d provenance record --prompt` was committed. It describes no real generation and is regenerable in one command. Everything else was moved, not deleted.

The ATTIC README now documents both retired suffix conventions, what went wrong with them, and what to use instead — the gap that let a second one grow up beside the first.

## Result

- `make test`: **1338 passed, 4 skipped, 0 failed**
- `scripts/check_large_files.py`: passes (largest changed file 38.86 KB)
- Untracked entries in the tree: 23 → 12, and the 12 left are the root `D4D_*.yaml` of unknown provenance, which are #409 and not mine to move

## Filed, not fixed

**#410** — the composition-row subject convention. `_map_composition_paths` names its subject after the last path segment, so `anomalies.id` becomes subject `d4d:Dataset/id` and the row asserts that the Dataset's own `id` closely matches an anomaly. All 16 composition rows have this shape; only the 4 roots are correct subjects. The distinguishing information survives only in a free-text column nothing consumes, and every future base-class slot adds two more such rows.

I accepted the 4 new rows in #407 because they are consistent with the 12 already committed. Consistent, but consistently wrong — and fixing it changes rows that predate this work, so it belongs with #234/#295 and the question of whether this mapping is generated or curated.


realmarcin:
## Adversarial review — findings and what changed

Reviewed as a separate pass over the diff. Three findings; one was mine, one was a pre-existing bug I triggered, one is a disclosure.

### 1. `d4d runs archive` destroyed documentation — filed as #412, fixed in `6ae8000d`

The worst finding, and I caused it. `_write_archive_note` `write_text`'d the whole README from the current invocation's `moved` list. Archiving four labels under #408 therefore left a README describing only the fourth, and **deleted the rationale for the 2026-04 and 2026-07-23 series** — the only place that explanation lived. The function's own docstring says that reason is "exactly what a later reader needs and cannot reconstruct."

Two further defects in the same function: the heading counted files but said "run directories" (9 files read as 9 directories, against 25 actually present), and entries were `<label>/<file>`, which does not identify anything given `claudecode_agent` and `claudecode_agent_core` share labels.

Fixed: appends a dated section per invocation, counts files and directories separately, paths relative to the archive root. The function had **no direct test coverage**, which is how a clobbering writer survived — three regression tests added. The damaged README is rebuilt with the prior rationale restored, and its inventory verified against disk in both directions (131 files, 25 directories, listed set == on-disk set).

### 2. SSSOM integrity — verified clean, but the commit message under-described it

Diffed the merged mapping against `2e3e2c46`:

- **0 rows lost**, exactly the 4 expected rows added, column set identical.
- One row I did not mention in `54047e04`: `d4d:LabelingStrategy/data_annotation_platform` had `subject_multivalued` corrected `True` → `False`. Checked against the schema — `data_annotation_platform` is `range=string, multivalued=False`, so the committed `True` was stale and the new value is right. A correction, not a regression, but the commit message described only additions and should have said so.

### 3. Archive integrity — verified

All archived YAMLs parse, `d4d runs restore --label …` resolves the moved paths, and `d4d runs list` no longer sees any superseded label.

### Result

`make test`: **1341 passed, 4 skipped, 0 failed** (up from 1338; the 3 new are the archive-note regression tests).

PR #423, opened 2026-08-10
Carry a still-true validation verdict across a re-record (#396)

### LINKED ISSUE #396: d4d provenance record silently drops the validation: block, breaking runs check --strict
Found while re-recording provenance for `2026-08-07_claude-opus-5-claudecode-generic-v3_rep1` (CHORUS).

`d4d provenance record` rewrites `{PROJECT}_provenance.yaml` from scratch. The `validation:` block is written separately, by `d4d runs validate`. So any re-record removes it, and the run immediately fails the gate:

```
❌ CHORUS  2026-08-07_...  nothing to verify: (no validation block).
   A run that cannot be checked is not a run that passed
1 run(s) checked, 1 subject to the requirement, 1 failing
```

Nothing warns that this happened. The command prints `✓ <path>` and exits 0.

This is easy to hit, because re-recording is the *correct* response to several situations — a header field was added, an artifact was corrected in Phase 3, provenance was recorded before the last edit. The playbook (`.claude/commands/d4d-full-core.md`) tells you to re-run `d4d runs validate` "if anything changed", but the coupling is not obvious from either command's output, and a run that skips it looks finished while failing `--strict`.

Options, roughly in order of preference:

1. Preserve an existing `validation:` block across a re-record when the artifact hashes it names still match the artifacts on disk. If they don't match, drop it — that's a genuine staleness.
2. Have `record` warn on stdout when it overwrites a record that had a `validation:` block, naming the follow-up command.
3. At minimum, document the ordering in the `record` docstring so `--help` shows it.

(1) is the real fix: the verdict is still true if the bytes it describes are unchanged, and discarding it forces a redundant re-validation sweep.

realmarcin:
Closes #396.

`d4d provenance record` rewrote the record from scratch while `d4d runs validate` writes the `validation:` block separately, so **any re-record deleted the verdict** — the run then failed `d4d runs check --strict` with "nothing to verify", while `record` printed a tick and exited 0.

This was easy to hit, because re-recording is the *correct* response to several situations (a header field added, Phase 3 corrected an artifact, provenance recorded before the last edit). The playbook worked around it with an order-critical sequence executed **by hand on every run** — one of the six manual interventions the genericity audit (#418) catalogued.

## The fix

A verdict is a claim about specific bytes, and it is still true if those bytes have not changed. `preservable_validation()` carries the block forward when every artifact it names still hashes to what it recorded, and returns `None` otherwise — the same staleness rule `validation_status` already applies, so the two cannot disagree about what a verdict describes.

**Dropping is now loud**, which was the actual defect:

```
  validation: carried forward — the artifacts it names still hash to what it recorded

  ⚠️  validation: dropped — the artifacts changed since it was recorded, so the
      verdict is about bytes that no longer exist.
      Re-run `d4d runs validate --label <L> --project <P>` before `d4d runs check`.
```

Deleting the verdict is *correct* when the bytes changed. Doing it silently is what made a broken run look finished.

## Refused rather than carried

Four cases, each tested: a caller that supplied its own block; a verdict whose artifacts no longer match; one naming a **deleted** file (`verify_entry` returns `None` there, not `False` — the distinction an early version of this got wrong); and one with no artifacts at all, which cannot be shown still true.

## Verified end to end on a real record

- Re-recorded `2026-08-07_…_rep2` VOICE with artifacts unchanged → block preserved, `--strict` passed. **Before this change that sequence failed the gate.**
- Edited the full record, re-recorded → warning fired, block dropped.
- Restored the artifact byte-identically, re-validated → passing again, `data/` clean.

## Also

The playbook's completion criteria no longer prescribe the ordering, since it is no longer load-bearing — the manual step only really goes away if the instructions stop asking for it.

5 tests. **1370 passed, 4 skipped, 0 failed.**

### DIFF
--- a/.claude/commands/d4d-full-core.md
+++ b/.claude/commands/d4d-full-core.md
@@ -477,7 +477,17 @@ and it marks what it cannot recover rather than filling it in.
   closing rows were appended, and a whole series was hashed before its headers
   were edited. The API path cannot do this, because it writes provenance
   in-process after every phase; running the check gives this path the same
-  property. Re-run `d4d runs validate` and check again if anything changed.
+  property.
+
+  **Re-recording no longer discards the verdict** (#396). `provenance record`
+  used to rewrite the file from scratch and delete the `validation:` block that
+  `d4d runs validate` had written, so a re-record silently failed the gate while
+  printing a tick. It now carries the block forward when every artifact it names
+  still hashes to what it recorded, and drops it with a warning naming the
+  follow-up command when one does not. So the sequence is no longer
+  order-critical: record and validate in either order, and re-record freely.
+  Still run `d4d runs validate` once after the artifacts are final, because a
+  run that has never been validated has no verdict to carry.
 - Final summary to the user: per project, report full/core line counts as
   informational metadata, never as a quality gate, plus validation status.
 
--- a/src/data_sheets_schema/cli/provenance.py
+++ b/src/data_sheets_schema/cli/provenance.py
@@ -41,6 +41,17 @@ def record(project, method, label, input_bundle, prompts):
                        prompt_paths=[Path(p) for p in prompts] or None)
     out = rec.write(record_path_for(project, method, label))
     click.echo(f"✓ {out}")
+    # Say what happened to any prior verdict. Re-recording used to delete it
+    # silently and the run then failed `--strict` with "nothing to verify",
+    # while this command printed a tick and exited 0 (#396).
+    if rec.validation_carried is True:
+        click.echo("  validation: carried forward — the artifacts it names "
+                   "still hash to what it recorded")
+    elif rec.validation_carried is False:
+        click.echo("  ⚠️  validation: dropped — the artifacts changed since it "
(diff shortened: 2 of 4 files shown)


realmarcin:

**A carried verdict could outlive its schema.** `validation.artifacts` pins only the record, not the schema it was validated against. Carrying the block forward on unchanged artifacts alone would let a verdict survive a schema change — a gap the *previous* behaviour bounded by accident, since deleting the block on every re-record meant a verdict rarely outlived much.

Fixed in `4c9d0f1`: `preservable_validation()` now also compares the record's `schema` block (`full_sha256`/`core_sha256`, already present on every record) and drops the verdict if either moved. Three tests.

This closes only the half this PR opened. A verdict written by `d4d runs validate` and never re-recorded still has no schema binding, and `validation_status()` still doesn't check one — filed as **#426**.

Re-verified on the real 2026-08-07 VOICE record: unchanged schema still carries forward, `--strict` still passes, `data/` restored clean. **1373 passed, 4 skipped.**

Issue #422, opened 2026-08-10
Per-GC scope boundaries belong in the manifest, not in the launch prompt


realmarcin:
From the genericity audit (#418).

The VOICE launch prompt for the 2026-08-07 sweep carried this, and it is the single clearest per-GC intervention in the pipeline:

> CRITICAL SCOPE BOUNDARY: this run covers the adult/main Bridge2AI-Voice dataset ONLY. `VOICE_PEDIATRIC` is a separate project with its own DOI, protocol and ethics approval … Do not read `data/preprocessed/concatenated/VOICE_PEDIATRIC_preprocessed.txt` … doing so is what previously caused every VOICE replicate to fail validation (issue #292).

Three things are wrong with solving it this way.

**It is the wrong layer.** The frozen prompt body already says *"DECLARED INPUT BUNDLE — your only source of dataset facts: {BUNDLE}"*. The bundle **is** the scope. A prompt paragraph restating the scope is redundant when the bundle is right, and cannot save it when the bundle is wrong.

**It is a quality warning.** Naming #292 tells the model what went wrong last time, which the priming taxonomy in `.claude/commands/d4d-full-core.md` marks ⚠️ even for the tuned arm and excludes from generic. It steers behaviour rather than describing evidence.

**It does not generalise.** Any new dataset with a companion cohort needs someone to notice and write a bespoke paragraph. That is the definition of per-GC adaptation, and it is invisible to the prompt-condition tests because it lives in the launch text, not the file.

## What the bundle already does

`VOICE_PEDIATRIC` has its own manifest selection of six sources and its own 204 KB bundle (NEXT_TASKS §9). `VOICE`'s bundle is byte-identical to the one 49 runs attest. So the scoping is already expressed correctly in the manifest — the prompt paragraph is belt-and-braces that became load-bearing prose.

## Suggested fix

1. Delete the scope paragraph from the launch template. If a run needs a scope constraint the prompt cannot express, that is a manifest bug, not a prompting problem.
2. If the concern is that VOICE's bundle *discusses* the pediatric dataset (it does — the pediatric PhysioNet record is one of VOICE's eleven sources), decide that at the manifest level: either remove it from VOICE's selection, or accept that VOICE may reference it and let the schema's relationship slot carry it, which is what happened and what validated.
3. Add a test that the launch template contains no project name — the mirror of the existing test that the prompt *file* contains none. Related to #419 and #420, which are about the same blind spot from the provenance side.

Issue #419, opened 2026-08-10
Agentic provenance hashes the base prompt, not the instruction that was sent


realmarcin:
From the genericity audit (#418, `notes/PIPELINE_GENERICITY_AUDIT.md`).

`d4d provenance record --prompt <file>` hashes the file it is handed and the record then asserts that file was the instruction:

```yaml
prompts:
  files:
    - path: src/download/prompts/d4d_generic_arm_prompt.md
      sha256: 0fbc626b…
      bytes: 5754
```

and the generated record repeats it in its header: `# Prompt: …d4d_generic_arm_prompt.md (identical for all projects)`.

**On the agentic path that is not what the model received.** Those runs are launched from a hand-composed task prompt. The VOICE run under `2026-08-07_claude-opus-5-claudecode-generic-v3_rep2` was sent, in addition to the file's content:

- a `CRITICAL SCOPE BOUNDARY` paragraph naming VOICE, VOICE_PEDIATRIC and issue #292;
- a five-step `FINALIZATION SEQUENCE` working around #396;
- `REASONING EFFORT` instructions working around #397;
- exact output paths and a concurrency warning.

None of that is in the hashed file. `VOICE_PEDIATRIC_reconciliation.md` from the same label shows the same shape, so it is not a one-off.

Separately, **no `.claude/` file is hashed in any provenance record**, and `.claude/commands/d4d-agent.md` is where the v2/v3 decision rules actually reach the agent (see #420).

## Why it matters

`provenance.py`'s own docstring for `prompt_facts` says *"A condition whose prompt cannot be identified cannot be replicated."* Today an agentic record identifies a prompt that was only part of its condition, which is worse than identifying none — it reads as complete.

It also silently breaks the prompt-condition study: the generic-vs-tuned comparison assumes generic runs received identical text, and these did not.

## Suggested fix

Record the instruction actually sent, not the file it was based on:

- add `--prompt-text <file>` (or have the launcher write the composed prompt to the run directory) so the verbatim launch text is hashed;
- hash the `.claude/` playbooks a run followed, since they carry decision rules;
- keep the base-file hash as well — knowing both what it was derived from and what was sent is the useful pair.

A cheaper interim: have `d4d runs check` warn when an agentic record's `prompts.files` contains only a base prompt, so the gap is visible rather than implied.

PR #425, opened 2026-08-10
Render the agentic instruction instead of typing it (#419, #422)

realmarcin:
Addresses #419 and #422. Makes the agentic path standardized the way the API path already is.

## Why the API path is more standardized — the specific reason

It never types an instruction. `resolve_prompt()` builds the text from a `RunSpec`, so the condition, every substitution and any per-project content are **functions of declared inputs**. Per-GC content can enter through exactly one door — `--condition tuned` → `components/{PROJECT}.md` — and the record names it.

The agentic path had no way to obtain that text, so launch prompts were hand-composed. The VOICE run of 2026-08-07 was sent a project-specific scope paragraph naming the pediatric dataset and #292 — in no prompt file — while its provenance hashed the base file and the header said *"identical for all projects"*.

## The fix is mostly connecting parts that already existed

**`d4d api render-prompt`** emits exactly what a run should receive, for any runtime:

```
$ d4d api render-prompt --project VOICE --label 2026-08-10_x_rep1 \
    --condition generic_v3 --out /tmp/sent.txt
✓ /tmp/sent.txt
# rendered generic_v3 for VOICE / baseline / runtime=Claude Code
# sha256 32619074e100ea1c…  (5244 bytes)
```

The launcher now renders rather than writes, so "generic" versus "generic with a paragraph added" stops depending on who composed the prompt. Commentary goes to stderr so stdout pipes cleanly.

Two things had to become parameters. `RUNTIME` was a module constant, so every rendered header said "Claude API (direct)". And the provider came from `provider_identity()` — the endpoint *this process* is configured against, which is right for a run this process makes and wrong for rendering one another runtime executes: it put `LBL CBORG (proxy to Anthropic)` into a Claude Code header. Both default to the previous values, so **the API path is unchanged**.

## The record now carries what was sent

`prompt_facts` gains a `request` block (sha256 + length of the resolved text), `build_record` takes `prompt_request`, and both paths pass it — the API path from `resolve_prompt(spec)`, the agentic path from `d4d provenance record --prompt-text`. Only the hash is stored; the point is to prove sameness, and a record is not an archive.

**This was a shared defect, not an API-versus-agentic one.** Neither path recorded the resolved request. `prompt_request_hash()` was written for exactly this, with a docstring arguing for itself — *"substitution is exactly what makes the file and the request different objects, so the request needs its own hash"* — and had **no caller**. It does now.

## What this does not do

Nothing yet *checks* that an agentic record's request hash matches what its spec would render. That gate is the point of the exercise, and it needs this field to exist first — it's the natural next PR, and it's what turns "don't intervene" from a rule into something detectable.

## Verification

- 13 tests, including: the runtime and provider appear in the rendered header; no `{PLACEHOLDER}` survives in any of the four generic conditions; a generic render names no GC it wasn't given; stdout is the instruction alone; the digest is of the resolved text, not the file; the two hashes differ; absent stays absent when `--prompt-text` is omitted.
- Full suite: **1378 passed, 4 skipped, 0 failed.**

### DIFF
--- a/src/data_sheets_schema/api_runner.py
+++ b/src/data_sheets_schema/api_runner.py
@@ -206,6 +206,18 @@ class RunSpec:
     # The GitHub assistant writes flat into data/sheets_d4dassistant. Rather than
     # two runners, the layout is a parameter — everything else is identical.
     out_dir: Path | None = None
+    # Which runtime the rendered instruction should declare. Defaults to this
+    # module's own, so the API path is unchanged. It is a parameter because the
+    # agentic path needs the same instruction rendered for `Claude Code`: the
+    # whole point of rendering is that nobody types the header by hand, and a
+    # hardcoded runtime would force exactly that (#419).
+    runtime: str = RUNTIME
+    # Likewise the provider. `provider_identity()` reports the endpoint *this
+    # process* is configured against, which is the right answer for a run this
+    # process is about to make and the wrong one for rendering an instruction
+    # another runtime will execute — it rendered "LBL CBORG (proxy to
+    # Anthropic)" into a Claude Code header, a provider that run never touches.
+    provider: str | None = None
 
     @property
     def full_path(self) -> Path:
@@ -325,8 +337,8 @@ def resolve_prompt(spec: RunSpec) -> str:
         # headed "Agent runtime: Claude Code" on "claude-opus-5[1m]" because
         # this prompt was written for the agent path and reused verbatim — the
         # artifact asserted a runtime and model it never touched.
-        "{RUNTIME}": RUNTIME,
-        "{PROVIDER}": ident["provider"] or PROVIDER,
+        "{RUNTIME}": spec.runtime,
+        "{PROVIDER}": spec.provider or ident["provider"] or PROVIDER,
         "{MODEL}": settings["name"],
         # v2 introduced `{DATE}` but nothing substituted it, so the literal
         # string reached the model. Its records carry the right date only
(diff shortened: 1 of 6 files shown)


realmarcin:

**This PR added a third `resolve_prompt(spec)` call, in `build_record`, after the last phase.** `resolve_prompt` reads `provider_identity()` and `_model_settings()`, so anything moving between send and record — an edited deterministic config, a changed endpoint — would have the record attest a prompt that was never sent.

That is *precisely* the failure `run_date` is frozen to avoid, and the comment there already spells it out: recomputing per call "made the provenance digest, which is computed after the last phase, attest a prompt that was never sent". Adding a hash of the request while leaving the request re-rendered would have reintroduced it in the one field whose entire purpose is to be exact.

Fixed: `RunSpec.instruction` is a `cached_property` and all three call sites use it — the phase message, the digest, and the provenance record. Three tests, including one that swaps `_model_settings` between two reads and asserts the instruction does not move.

## Also checked

- **Tuned still works through the new command.** `--condition tuned` renders the `PROJECT-SPECIFIC EVIDENCE` block from `components/{PROJECT}.md`; `--condition generic_v3` does not. So the one declared door for per-GC content still opens, and only there.
- **No `{PLACEHOLDER}` survives** in any of the four generic conditions.
- **API defaults unchanged** — `runtime` defaults to the module constant, `provider` to `provider_identity()`.

## Filed, not fixed

**#428** — the command lives under `d4d api`, which is the group for the path that does *not* need it. An agentic run has no reason to look there. Real discoverability defect, and the audience most likely to miss it is the one it was built for.

**1390 passed, 4 skipped, 0 failed.**

PR #429, opened 2026-08-10
Detect label/prompt mismatch, and hash the playbooks that carry the rules (#420)

### LINKED ISSUE #420: Runs labelled generic-v3 hash the v1 prompt, and the playbooks carrying v3's rules are hashed nowhere
From the genericity audit (#418).

The 2026-08-07 canonical sweep is labelled `…_claude-opus-5-claudecode-generic-v3_rep{1,2,3}`. Its provenance hashes:

```
src/download/prompts/d4d_generic_arm_prompt.md   sha256 0fbc626b…   bytes 5754
```

5754 bytes is **v1**. `d4d_generic_arm_prompt_v3.md` is 7204 bytes and its body is 4903 characters against v1's 3567.

v3 adds **seven substantive decision rules** over v1 — verified by diffing the two bodies through `prompt_body()`:

- one object per distinct entity when the range is multivalued
- populate the field with the information asked for, not a pointer or a "pending" note
- read the slot description; put evidence in the field it answers
- when a range is a class, populate the fields that class declares rather than cramming prose into `description`

So a run labelled v3 attests a v1 hash, and the label is the only place the v3 claim exists.

## Where the v3 behaviour actually came from

Not nowhere — `.claude/commands/d4d-agent.md` carries the slot-filling contract ("structured slots first … then `description` … then `notes`"), which is PR #394's playbook parity work. That file reaches the agent and shapes the output.

**But no `.claude/` file is hashed in any provenance record.** So the instructions that distinguish v1 behaviour from v3 behaviour are unversioned in the record, and two runs with different playbook content are indistinguishable afterwards.

## Why it matters

The v2-vs-v3 comparison that NEXT_TASKS §6 defers the promotion decision to cannot be run on these records: nothing in them establishes which rule set was in force. The label is an assertion by the person who typed it.

## Suggested fix

1. `d4d runs check` should fail, or at least warn, when a label names a prompt version whose file is not among `prompts.files`.
2. Hash the `.claude/` playbooks alongside the base prompt (see #419).
3. Decide whether the agentic path should read `d4d_generic_arm_prompt_v3.md` directly rather than having its rules mirrored into a playbook — mirroring is what let the two drift.

realmarcin:
Closes #420. Two halves of the same gap: a run labelled `generic-v3` hashed **v1**, and the files that actually gave it v3 behaviour were hashed **nowhere**.

## The label was the only place the claim lived

`condition_from_label()` reads what a label asserts; `condition_of()` already read what its hashed prompt implies. When they disagree, `d4d runs check` says so:

```
⚠️  5 run(s) whose label and hashed prompt name different conditions (#420):
   VOICE  2026-08-07_..._rep2  label claims 'generic_v3' but the hashed prompt
   is 'generic' (d4d_generic_arm_prompt.md); the two are different conditions
```

**Reported, never fatal.** It's a defect in records that already exist, and failing them retroactively would block every gate on history nobody can change. Visible is the point — the sweep said v3 and hashed v1 for three days with nothing to say so.

Two details that matter:

- **Longest-condition-first matching is load-bearing.** `generic` is a prefix of `generic-v3`, so a shortest-first scan answers `generic` for every versioned label — reporting the exact mismatch this exists to detect *as agreement*.
- **Silence when it can't tell.** Labels predating the convention name no condition; a run with no record isn't thereby inconsistent. Absence of evidence must not read as a mismatch.

## The rules arrived through unhashed files

`.claude/commands/d4d-agent.md` carries the slot-filling contract that gives the agentic path its v2/v3 behaviour — mirrored there by #394 rather than read from the versioned prompt. The agent opens it because the prompt's first line tells it to, so it's a generation input the record never named.

`playbooks:` now hashes three: the full/core playbook, the agent playbook, the provenance guard. Recorded on **every** run, not just agentic ones — an API run doesn't read them, and a record showing them unchanged is how you can later tell that it didn't. A missing file is recorded as missing rather than omitted, so the set is always the same three.

10 tests. **1400 passed, 4 skipped, 0 failed.**

### DIFF
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -310,9 +310,12 @@ def check_cmd(method, label, project, strict):
     Use `--strict` after a generation run so a missing record fails the step
     rather than being noticed later.
     """
-    from data_sheets_schema.runs import check_provenance, discover, is_complete
+    from data_sheets_schema.runs import (check_provenance, discover,
+                                         is_complete,
+                                         prompt_condition_mismatch)
 
     rows = []
+    mismatches = []
     for run in discover():
         if run.is_core or run.deterministic:
             continue
@@ -326,6 +329,10 @@ def check_cmd(method, label, project, strict):
             if not is_complete(run.method, run.label, proj):
                 continue
             rows.append(check_provenance(run.method, run.label, proj))
+            m = prompt_condition_mismatch(run.method, run.label, proj)
+            if m:
+                mismatches.append({"project": proj, "label": run.label,
+                                   "reason": m})
 
     failed = [r for r in rows if not r["ok"]]
     required = [r for r in rows if r["required"]]
@@ -335,6 +342,19 @@ def check_cmd(method, label, project, strict):
                f"requirement, {len(failed)} failing")
     if not failed:
         click.echo("All runs subject to the live-provenance requirement satisfy it.")
+
+    # Reported separately from the provenance verdict, and never fatal. A
+    # label naming a condition its prompt does not match is a real defect
+    # (#420) — but it is a defect in records that already exist, and failing
+    # them retroactively would block every gate on history nobody can change.
+    # Visible is the point: the 2026-08-07 sweep says `generic-v3` and hashes
+    # v1, and nothing said so for three days.
+    if mismatches:
+        click.echo(f"\n⚠️  {len(mismatches)} run(s) whose label and hashed "
+                   "prompt name different conditions (#420):")
+        for m in mismatches:
+            click.echo(f"   {m['project']:9} {m['label']:44} {m['reason']}")
+
     if strict and failed:
         raise SystemExit(1)
 
--- a/src/data_sheets_schema/provenance.py
+++ b/src/data_sheets_schema/provenance.py
@@ -648,6 +648,23 @@ def build_record(project: str, method: str, label: str, *, mode: str,
             "reason": "manifest has been edited since this run",
         })
 
+    # ---- agent playbooks -------------------------------------------------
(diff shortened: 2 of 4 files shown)


realmarcin:

**`playbook_facts()` was fabricating provenance for reconstructed records.** It was called unconditionally in `build_record`, and `d4d provenance backfill` builds `mode="reconstructed"` records — so backfilling a historical run would have hashed *today's* `.claude/` content and recorded it as what that run followed.

That is precisely what this module's docstring forbids, in the same shape as the example it gives:

> "hashing them today and recording that against an April run would be a fabricated provenance claim. Such fields are listed under `unrecoverable` with the reason, never silently filled."

Adding a field to close a provenance gap while opening a fabrication would have been a poor trade. Now gated on `mode == "live"`; a reconstructed record carries `playbooks: null` plus an `unrecoverable` entry, because an absent field with no reason reads as "this run had none".

## Checked and clean

- **No false positives on the label check.** 15 runs flagged across the corpus — all three replicates of the 2026-08-07 sweep and nothing else. The 2026-08-06 schema2 runs correctly hash v3 and stay silent, which is the useful control: the mismatch is specific to the agentic sweep.
- **`tuned` does not false-positive**, despite sharing v1's prompt file — 0 of the 15 involve it.
- **Label parsing** handles `tuned`, `generic-v3-schema2`, plain `generic`, and returns `None` for `bare` (a label naming no condition is not a defect).

## Filed, not fixed

**#431** — `AGENT_PLAYBOOKS` is a hand-maintained list of which files carry decision rules. A renamed playbook silently becomes `exists: false`; a new one is invisible. That reproduces the original problem — instructions reaching the model through untracked files — one level up.

**1403 passed, 4 skipped, 0 failed.**

PR #430, opened 2026-08-10
Build the render gate: check the instruction against the spec that produced it

realmarcin:
Stacked on #429 (#420). Merge that first.

Rendering the instruction (#425) made hand-editing **avoidable**. This makes it **detectable** — the difference between a convention and a control.

## It catches the real thing

Demonstrated against the actual intervention that started this thread, a `CRITICAL SCOPE BOUNDARY` paragraph appended to a VOICE launch prompt naming the project, the pediatric dataset and #292:

```
rendered, unedited      -> match
a scope paragraph added -> mismatch
    recorded ac65085a5807… but the spec renders f959a43f5aea…
```

One added space is caught too.

## The record now carries what it was rendered from

`prompts.request.spec` holds everything `resolve_prompt` reads. Two of those fields — `manifest_line` and `run_date` — were recorded **nowhere**, and both change the rendered bytes, so reconstructing a spec from the record's other blocks would have been guesswork that failed intermittently. Storing it makes the record able to re-derive its own claim.

## Four outcomes, and the negative ones are distinct

| status | meaning | under `--strict` |
|---|---|---|
| `match` | the run received what its spec says it should have | pass |
| `mismatch` | it did not | **fatal** |
| `unverifiable` | a hash with no spec, or a spec that will not render | reported |
| `absent` | no request hash — every record before #425 | silent |

`mismatch` is fatal because every record carrying a request hash was written *after* the field existed, so it is a claim about a run made now. `absent` is silent because failing history for a field that postdates it is the error the live-provenance cutoff exists to avoid. And `unverifiable` is not `mismatch`: "cannot check" and "checked and wrong" say different things.

## The remaining hole, stated

The one way to make a tampered instruction pass is to **edit the recorded spec to match it**. That changes the condition the record asserts, and the label check in #429 catches it. Tested — but it is a two-step consistency argument, not a single lock, and worth knowing.

8 tests. **1408 passed, 4 skipped, 0 failed.**

### DIFF
--- a/src/data_sheets_schema/api_runner.py
+++ b/src/data_sheets_schema/api_runner.py
@@ -219,6 +219,21 @@ class RunSpec:
     # Anthropic)" into a Claude Code header, a provider that run never touches.
     provider: str | None = None
 
+    def render_spec(self) -> dict[str, Any]:
+        """Everything `resolve_prompt` reads, for the record to keep.
+
+        A record that stores only the resolved hash can say the instruction
+        changed but not what it should have been. Storing the spec beside it
+        lets `verify_request()` re-render and compare, which is what turns
+        "do not intervene" from a rule into something detectable (#420).
+        """
+        return {"condition": self.condition, "arm": self.arm,
+                "manifest_line": self.manifest_line, "run_date": self.run_date,
+                "runtime": self.runtime,
+                "provider": self.provider or provider_identity()["provider"]
+                or PROVIDER,
+                "bundle": str(self.bundle)}
+
     @cached_property
     def instruction(self) -> str:
         """The resolved instruction, rendered once per spec.
@@ -1689,6 +1704,7 @@ def execute(spec: RunSpec, *, dry_run: bool = False, resume: bool = True,
         # record exactly what it sent rather than only what it was built from
         # (#419). `prompt_request_hash` was written for this and had no caller.
         prompt_request=spec.instruction,
+        prompt_request_spec=spec.render_spec(),
         schema_digest_md5=schema_digest.fingerprint(schema_digest.digest_text("Dataset")),
         extra_notes=[
             (f"Generated via {RUNTIME}; temperature {settings['temperature']} "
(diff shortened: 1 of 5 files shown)


realmarcin:

**A prompt file changing after a run reported `mismatch` — fatal under `--strict`.** But a differing hash has two causes, and they are not the same finding:

- the instruction was edited (the intervention the gate exists for)
- the prompt file moved since the run, so re-rendering no longer reproduces what was sent

The second is ordinary evolution — v4 exists, v5 will — and as written, **the first legitimate prompt edit would have made `--strict` fail every historical record carrying a request hash.** A gate that fires on maintenance is a gate people turn off.

Demonstrated before fixing:

```
before any edit                 -> match
after the PROMPT FILE changed   -> mismatch     <- wrong
```

Fixed with data the record already carries. `prompts.files` holds each prompt's sha256 as of the run, so comparing to disk settles it:

| situation | verdict |
|---|---|
| file unchanged, hashes differ | `mismatch` — message now says "from an unchanged prompt file" |
| file changed since the run | `unverifiable` — the recorded hash still stands, it just cannot be re-derived |
| file not pinned in the record | `unverifiable` — will not guess between the two |

Four tests, including that **an edited instruction is still a mismatch** — the discrimination must not be bought by weakening the case the gate is for.

## Left open, deliberately: #432

An instruction edited *before* rendering — by editing the prompt file itself — renders and verifies consistently, so it passes. The gate proves the instruction matches the spec **as the files stand now**, not that the condition's prompt is canonical.

Closing that needs a per-condition canonical hash registry. That is real maintenance (every legitimate prompt edit needs the pin updated) and a judgement about whether prompt-file integrity is actually a threat model here, which is not mine to make.

Also worth stating: the two-step argument I flagged when opening this PR is now weaker than I claimed. Editing the recorded spec to match a tampered instruction still changes the condition the record asserts, and #429's label check catches it — but that only holds when the label names a condition, which not every label does.

**1415 passed, 4 skipped, 0 failed.**

PR #435, opened 2026-08-11
Pin each condition's prompt text, so an edit made before rendering is a finding (#432)

### LINKED ISSUE #432: The render gate cannot detect an instruction edited before rendering
Raised reviewing #430, and left open there deliberately.

The render gate verifies that a record's `prompts.request.sha256` equals what re-rendering `prompts.request.spec` produces. That catches an instruction edited **after** rendering — the `CRITICAL SCOPE BOUNDARY` paragraph that motivated the whole thread.

It does not catch an instruction edited **before** rendering, by editing the prompt file itself:

1. add the project-specific paragraph to `d4d_generic_arm_prompt_v3.md`
2. render — the paragraph is now part of what the spec produces
3. run, record

`verify_request()` re-renders from the same file and gets the same bytes. **`match`.**

## What does stand in the way

Not nothing, but nothing decisive:

- `prompts.files` records the prompt file's sha256, so the edit is *visible* in the record — but it is compared against the file on disk, not against a known-good value for that condition. An edited-then-committed prompt is self-consistent.
- The prompt file's own editing rule says a change re-baselines the arm for every project at once, and the generic-arm tests assert the file names no project. So a project-specific paragraph added to a generic prompt would fail those tests — **if** they were run, and if it were added to a *generic* file rather than to a new condition.

So the honest statement is: the gate proves the instruction matches the spec **as the files stand now**, not that the condition's prompt is the canonical one.

## What would close it

Pin the expected hash per condition, so a record can be checked against what `generic_v3` is *supposed* to be rather than only against what the file currently says:

```yaml
# a registry, versioned in the repo
generic_v3: sha256 32619074e100ea1c…
```

`d4d runs check` then compares three things instead of two — the recorded request, a fresh render, and the condition's canonical prompt — and can distinguish "instruction edited", "prompt file changed since the run" (already handled, see below) and "prompt file is not the canonical one for its condition".

That is a real registry with real maintenance cost: every legitimate prompt edit needs the pin updated, and the pin has to live somewhere that is itself trustworthy. Worth doing only if prompt-file integrity is actually a threat model here, which is a judgement about who edits this repo.

## Already handled, for the record

A prompt file changing *after* a run no longer reports `mismatch`. #430 distinguishes it: the recorded file hash is compared to disk, and a difference yields `unverifiable` with the reason, because "the file moved" and "the text was edited" are different findings and only one is fatal under `--strict`.

realmarcin:
Closes #432.

## The hole

The render gate (#425, #429, #430) verifies that a record's `prompts.request.sha256` equals what re-rendering `prompts.request.spec` produces. That catches an instruction edited **after** rendering — the `CRITICAL SCOPE BOUNDARY` paragraph that started this thread.

It cannot catch one edited **before**:

1. add the paragraph to `d4d_generic_arm_prompt_v3.md`
2. render — the paragraph is now part of what the spec produces
3. run, record

`verify_request()` re-renders from the same file, gets the same bytes, and reports `match`. Correctly. The record is self-consistent and the condition label is a lie.

`tests/test_prompt_registry.py::TestTheRecordGate::test_the_pre_render_edit_the_render_gate_cannot_see` is that reproduction, asserting `match` from the render gate and `uncanonical` from the new one side by side. `test_the_same_run_under_the_published_prompt_passes_both` is the negative control.

## What is added

`src/download/prompts/canonical_hashes.yaml` — a pin per prompt file: sha256, byte count, `pinned_on`, `pinned_at_commit`, and a **required** reason. Rotation keeps the previous hash rather than dropping it, so runs made under the prompt of the day stay `superseded` rather than turning `uncanonical` the moment a condition moves.

Four verdicts for a hash: `canonical` / `superseded` / `uncanonical` (never pinned — the finding) / `unpinned`.

Three gates, at the three places a claim is made:

| where | asks | on failure |
|---|---|---|
| `d4d api prompts check --strict` + `tests/test_prompt_registry.py` | is the working tree at its pins? | fails, `unpinned` included |
| `d4d api run` / `batch` | is this run's condition declared? | refuses before the cost confirmation, no override |
| `d4d runs check --strict` | did this record hash a published prompt? | fatal on `uncanonical` only |

`render-prompt` and `provenance record` warn and proceed. Recording an uncanonical prompt is the honest act — it puts the evidence in the record for the check to fail on; refusing would reward not recording at all.

## What it does not do

**Not tamper-proofing, and the module docstring says so.** Whoever can edit a prompt can rotate its pin. What it buys: the two are separate deliberate acts, and a rotation leaves a dated line with a stated reason in a small file a reviewer reads, rather than a paragraph inside a 200-line prompt they do not. `pinned_at_commit` makes each pin re-derivable with `git show <commit>:<path>`.

## Blast radius

- Pinned from the files as they stand. **All 77 recorded prompt hashes in the corpus match** — nothing fails retroactively. `d4d runs check --strict`: 158 runs, 0 failing, before and after. `d4d api prompts check --strict`: 9 files, 0 off their pin.
- **Behaviour change:** `--condition tuned` for a project with no `components/{PROJECT}.md` (VOICE_PEDIATRIC) is now refused. Previously it silently produced a *generic* prompt under a tuned label, which is the same class of defect this thread is about — but it is a change, and it is the one thing here that could block a run someone expected to work.
- The playbook never told the agentic path to pass `--prompt` / `--prompt-text`, so a run following it recorded no prompt evidence at all — and as of today that fails `--strict` under #419. `.claude/commands/d4d-full-core.md` now passes both and says where to get the rendered text. That file is hashed into live records, so records made after this merge will carry a different playbook hash.

1450 passed, 4 skipped.

### DIFF
--- a/.claude/commands/d4d-full-core.md
+++ b/.claude/commands/d4d-full-core.md
@@ -442,7 +442,9 @@ After Phase 4 validates, emit a machine-readable provenance record:
 ```bash
 poetry run d4d provenance record \
   --project {PROJECT} --method {METHOD} --label {VERSION} \
-  --input-bundle {EXACT INPUT BUNDLE PATH}
+  --input-bundle {EXACT INPUT BUNDLE PATH} \
+  --prompt {EACH PROMPT FILE THE RUN CONSUMED} \
+  --prompt-text {THE INSTRUCTION AS SENT}
 ```
 
 Writes `{METHOD}_core/{VERSION}/{PROJECT}_provenance.yaml` capturing schema
@@ -454,6 +456,24 @@ if the input bundle is unreadable, because a run that cannot identify its own
 input has not produced reproducible output. Do not substitute a reconstructed
 record — `d4d provenance backfill` exists only for runs that predate this step,
 and it marks what it cannot recover rather than filling it in.
+
+**The prompt is an input like the bundle.** `--prompt` may repeat; pass every
+file the condition is built from. `--prompt-text` takes the instruction as
+actually sent — render it with `d4d api render-prompt --condition {CONDITION}
+--out <file>` rather than retyping it, so the text and its hash come from the
+same place. From 2026-08-10 a run without a recorded instruction fails
+`d4d runs check --strict` (#419): the gate was otherwise opt-in by omission,
+since a launcher that simply passes neither flag records nothing and nothing
+says so.
+
+Recording the file is not the same as vouching for it. A paragraph edited into
+a prompt file *before* rendering re-renders to itself, so the render gate
+reports `match` about an instruction nobody published (#432). What catches that
+is the canonical pin: `d4d api prompts check` compares each condition's prompt
+against the hash this repo declared for it, and `d4d runs check` compares the
+hash in the record. If either reports `uncanonical`, the run was made under
+text that is not a published version of its condition — say so in the
+reconciliation report rather than pinning the edit to make the check pass.
 
 ## Completion criteria (per project)
 
@@ -469,7 +489,8 @@ and it marks what it cannot recover rather than filling it in.
 - The provenance audit confirms that no older full/core YAML was used.
 - The core header contains `Phase 4 reconciliation: completed`.
 - The Phase 3/4 reconciliation report is present.
-- The live provenance record is present and its `record_mode` is `live`.
+- The live provenance record is present and its `record_mode` is `live`, and it
+  names both the prompt files and the instruction as sent.
 - `d4d runs check --strict` passes for the run. **Recording provenance is not the
   same as recording it correctly.** This path writes provenance as a step
   separate from writing the artifacts, so a record can pin a state the files
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -336,6 +336,31 @@ make gen-d4d-html
 make d4d-pipeline-{individual|concatenated|full}-gpt5
 ```
 
+## Canonical Prompt Registry
+
+Each condition's prompt files are pinned by hash in
+`src/download/prompts/canonical_hashes.yaml`.
+
+```bash
+d4d api prompts check --strict          # working tree vs the pins (CI gate)
+d4d api prompts pin --file <path> --reason '<why this is the text>'
+```
+
+**Editing a prompt file without rotating its pin fails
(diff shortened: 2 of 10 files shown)


realmarcin:
## Adversarial review, and what it found

Four findings, three fixed here (`abed21e2`), one filed and left.

**#436 — the pin is keyed on path, so `cp` was the whole bypass.** Measured on a scratch tree:

```
copied-path record -> ('unpinned', 'no canonical hash pinned for my_prompt_copy.md')
condition_of       -> None
label mismatch     -> None
```

Three checks and all three silent, while the label still claims `generic-v3`. The gate I had just added was avoidable with one `cp`. Closed with **coverage**: if the label claims a condition, the record must hash *some* pinned condition prompt. Deliberately not the claimed condition's exact file — 16 corpus records are labelled `generic-v3` and hash the pinned v1, which is #420, already reported, and non-fatal by an explicit earlier decision. Requiring the exact file would have reversed that decision through a side door.

**#437 — `unpinned` was doing two jobs.** A pinned file that has been deleted, and a record naming a pinned path it hashed nothing for, are both now `missing` and fatal wherever `uncanonical` is. The reason strings already said so; the status did not, and the status is what fatality is decided on.

**#438 — `pinned_at_commit` could name a commit that does not contain the pinned bytes.** The docstring advertises `git show <commit>:<path>` as the audit route, and pinning an uncommitted edit made it produce the wrong answer for exactly the case the registry exists to catch. `pin()` now refuses a dirty file. The initial pins were verified against history:

```
$ git show 225f2245:src/download/prompts/d4d_generic_arm_prompt_v3.md | shasum -a 256
2126da06d083…   # equals the pin
```

**#439 — the registry is re-parsed on every lookup.** Filed, not fixed: a few hundred reads of a 4 KB file, and a cache without mtime invalidation would make the rotation tests pass for the wrong reason.

### What the review could not settle

An **unlabelled** run is still free to use any prompt it likes — coverage keys off the condition the label claims, and a label naming none is left alone for the same reason `prompt_condition_mismatch` leaves it alone. Noted in #436 rather than closed by making labels mandatory in the same change.

1458 passed, 4 skipped. `d4d runs check --strict` 158 runs, 0 failing; `d4d api prompts check --strict` 9 files, 0 off their pin.

Issue #402, opened 2026-08-07
uriorcurie slots accept arbitrary bare tokens: 42% of nested ids in a clean-validating label are neither URI nor CURIE


realmarcin:
Found analysing label `2026-08-07_claude-opus-5-claudecode-generic-v3_rep1` (5 projects, all passing `linkml-validate` and `linkml-term-validator`).

## The trap

`uriorcurie` declares no `pattern`, so it renders in JSON Schema as:

```json
"id": {"type": ["string", "null"]}
```

Any string passes. A bare token like `purpose_ai_ready_critical_care_dataset` or `funder_nih` is neither a URI nor a CURIE on a declared prefix, and validates cleanly. The declared range promises identifier syntax; nothing enforces it.

This is a trap slot in the #360 sense that yields **zero** trap-inventory rows, because the failure is invisible to the validator. The trap inventory mines validation failures, so a slot that silently accepts wrong values is exactly the kind it cannot find.

## Measured

600 nested-object `id` values across the five records; **251 (42%) are neither a URI nor a CURIE on a prefix declared in the schema.**

| project | nested ids | not URI-or-known-CURIE | convention used |
|---|---|---|---|
| AI_READI | 21 | 0 | absolute URL |
| CM4AI | 218 | 0 | absolute URL |
| VOICE | 108 | 0 | `d4d:` CURIE |
| CHORUS | 115 | **114** | bare token |
| VOICE_PEDIATRIC | 138 | **137** | bare token |

Four mutually incompatible conventions, from one prompt, one model, one schema, one label. The prompt cannot be the variable.

## Why it matters

The core record is a **semantic exchange layer**. Entity resolution across records is the point, and it is currently impossible. The same real-world organization — NIH — appears as four different things:

- `funder_nih` (CHORUS, VOICE_PEDIATRIC)
- `https://cm4ai.org/data-releases/#funder-nih` (CM4AI)
- `d4d:VOICE-funding-nih-common-fund` (VOICE)
- no `id` at all, name only: `NIH Common Fund Bridge2AI program award` (AI_READI)

Nothing can join those. A bare token is worse than an absent id, because it looks like an identifier and is scoped to nothing — two records both using `funder_nih` would collide by accident rather than agree by reference.

## Suggested fix

1. Give `uriorcurie` a pattern, or add one to the identifier slots, so a bare token fails validation. This is the load-bearing change: without a validator signal, guidance in the prompt will keep drifting.
[...]

PR #440, opened 2026-08-11
Declare dataset scope in the manifest, where the next dataset inherits it (#422)

### LINKED ISSUE #422: Per-GC scope boundaries belong in the manifest, not in the launch prompt

realmarcin:
Closes #422.

## The finding

The VOICE launch prompt for the 2026-08-07 sweep carried this:

> CRITICAL SCOPE BOUNDARY: this run covers the adult/main Bridge2AI-Voice dataset ONLY. `VOICE_PEDIATRIC` is a separate project … Do not read `data/preprocessed/concatenated/VOICE_PEDIATRIC_preprocessed.txt` … doing so is what previously caused every VOICE replicate to fail validation (issue #292).

It worked, and it was per-GC adaptation at the wrong layer, of a kind the priming taxonomy excludes from the generic arm, that no future dataset inherits and no prompt test can see.

## The replacement

`source_manifest.yaml` gains a `scope:` block per project — `referent_id`, and any `related_but_distinct` dataset with `express_as` (the slot that carries the relation) and `in_bundle` (the source id, where its documentation is legitimately in the bundle):

```yaml
VOICE:
  referent_id: https://doi.org/10.13026/37yb-1t42
  related_but_distinct:
    - id: https://doi.org/10.13026/mf9s-5r03
      also_known_as: [https://doi.org/10.13026/h995-bt35, …]
      manifest_key: VOICE_PEDIATRIC
      in_bundle: physionet_pediatric_1_1_0
      express_as: related_datasets
```

```bash
d4d download scope --project VOICE      # show it
d4d download scope --check --strict     # check every record against it
```

**Checkable:** a record that identifies itself as a dataset its own project declares distinct is the #292 failure, now caught by a rule rather than prevented by a warning. 171 records checked, none out of scope. `check_manifest` refuses a declaration naming a project or a source that does not exist — a declaration that reads as a control and enforces nothing is worse than none.

**Inherited:** a new dataset declares two lines and gets the same check. An empty `related_but_distinct` means the question was asked and the answer was none.

## The issue's three asks

1. **Delete the paragraph from the launch template** — nothing in the repo held it; it lived in the launch message (`grep` finds it only in notes and tests). Since #429 an instruction that does not match its rendered spec fails `d4d runs check --strict`, so it can no longer arrive unrecorded. What was missing was somewhere legitimate for the information to live, which is this.
2. **Decide at the manifest level** — the pediatric PhysioNet record **stays** in VOICE's sources. Current VOICE documentation advertises the two releases together; dropping the source would make the bundle a worse description of the evidence in order to make a rule easier to state. The manifest now says the source is there and what it means.
3. **A test that the launch template names no project** — moved to the artifact actually sent: no rendered instruction, under any condition, names another project or another project's bundle. Banning directive phrases outright would be wrong (the prompt's uniform rules *are* directives, applied identically to every project); what must not appear is a second dataset. Plus a test that no `.claude/` playbook line carries a scope directive against a named project.

## Facts, and where they came from

Checked against the source documents, not against prior records: `10.13026/37yb-1t42` is the adult project-level DOI and `10.13026/mf9s-5r03` the pediatric one, both recorded in the bundle as "DOI (latest version)"; `h995-bt35` is pediatric 1.1.0; `8xbn-nq66` / `k81f-qr68` / `249v-w155` are adult 3.1.0 / 3.0.0 / 1.1. Version DOIs go under `also_known_as` — a check that knew only the project DOI would pass a record naming the version, which is the form the records actually use.

## Blast radius

- No bundle changed; `d4d download audit-manifest` still reports 41 sources.
- The manifest md5 that provenance records under `inputs.source_manifest` changes for future records. Nothing re-verifies that field, and the preprocessor writes no top-level manifest key into a bundle.
- `.claude/commands/d4d-full-core.md` gains a Scope section and a completion criterion, so its playbook hash changes again.
- The scope check reads the record's `id` only. A record can stay in scope by its identifier and still discuss the related dataset in prose — legitimate, and what `related_datasets` is for. Stated in the command's output rather than implied.

1478 passed, 4 skipped.

### DIFF
--- a/.claude/commands/d4d-full-core.md
+++ b/.claude/commands/d4d-full-core.md
@@ -64,6 +64,28 @@ neither, because a prompt written for one path was reused verbatim by another.
 - Full schema: `src/data_sheets_schema/schema/data_sheets_schema_all.yaml` (class `Dataset`)
 - Core schema: `src/data_sheets_schema/schema/data_sheets_schema_core_all.yaml` (class `CoreDataset`)
 
+### Scope: what the record is about
+
+The bundle is the evidence. What the record is *about* is declared in the
+`scope:` block of `data/preprocessed/source_manifest.yaml` — the referent, and
+any dataset that is related to it but distinct from it, with the slot that
+carries the relation. Read it with `d4d download scope --project {PROJECT}`.
+
+**A scope constraint goes in the manifest, never in the launch text.** The VOICE
+run of 2026-08-07 was sent a paragraph naming the project, the companion
+pediatric dataset, a file not to read, and the issue number of the last time it
+went wrong (#422). It worked, and it was per-GC adaptation invisible to every
+prompt test, because it lived in the message rather than in a file. If a run
+seems to need a constraint the bundle and the manifest cannot express, that is a
+manifest bug — fix it there, where the next dataset inherits it.
+
+A bundle may legitimately contain sources *about* a related dataset: VOICE's
+does, and the manifest says so (`in_bundle: physionet_pediatric_1_1_0`).
+Represent the relation through the declared slot — `related_datasets` — rather
+than merging the two, and never as a nested object standing in for the other
+dataset's own record. `d4d download scope --check` verifies afterwards that no
+record identifies itself as a dataset its project declares distinct.
+
 ## Outputs (per project)
 
 - Full: `data/d4d_concatenated/claudecode_agent/{VERSION}/{PROJECT}_d4d.yaml`
@@ -487,6 +509,9 @@ reconciliation report rather than pinning the edit to make the check pass.
 - The core file header names both its source-document bundle and full YAML input.
 - Both headers state that prior D4D factual reuse is prohibited.
 - The provenance audit confirms that no older full/core YAML was used.
+- `d4d download scope --check --project {PROJECT}` reports the record in
+  scope: it does not identify itself as a dataset the manifest declares
+  distinct from this project.
 - The core header contains `Phase 4 reconciliation: completed`.
 - The Phase 3/4 reconciliation report is present.
 - The live provenance record is present and its `record_mode` is `live`, and it
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -336,6 +336,33 @@ make gen-d4d-html
 make d4d-pipeline-{individual|concatenated|full}-gpt5
 ```
 
+## Dataset Scope (what a record is about)
+
+Declared per project in the `scope:` block of
+`data/preprocessed/source_manifest.yaml`, not in any prompt or launch message.
+
+```bash
+d4d download scope --project VOICE        # show the declaration
+d4d download scope --check --strict       # check every record against it
+```
+
+Each entry names the `referent_id` and any `related_but_distinct` dataset, with
+the slot that carries the relation (`express_as: related_datasets`) and, where
+the related dataset's documentation is legitimately in the bundle, the source id
+that carries it (`in_bundle`).
+
+**A scope constraint belongs in the manifest, never in the launch text** (#422).
+The VOICE run of 2026-08-07 was sent a paragraph naming the project, the
+companion pediatric dataset and a file not to read; it worked, and it was per-GC
+adaptation that no future dataset inherits and no prompt test can see. The
+manifest declaration is checkable (`check_record` catches a record that
+identifies itself as a dataset its project declares distinct) and inherited by
+any project that declares one.
+
+The check reads the record's `id` only. A record can stay in scope by its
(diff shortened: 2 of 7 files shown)


realmarcin:
## Adversarial review, and what it found

Four findings, all fixed here. The first is a gap in the fix this PR shipped.

**#441 — the check read only `id`, and reported 329 of 329 in scope.** Walking each record's whole tree for the identifiers VOICE declares distinct:

```
32 record(s), 143 value(s) place a related-but-distinct dataset outside its declared slot

2026-07-27_claude-opus-5_rep1/VOICE_d4d.yaml
    resources[1].doi = https://doi.org/10.13026/h995-bt35
    resources[1].download_url = https://physionet.org/content/b2ai-voice-pediatric/1.1.0/
    resources[1].version_access.latest_version_doi = https://doi.org/10.13026/mf9s-5r03
2026-07-28_claude-opus-5-deprimed_rep2/VOICE_d4d.yaml
    resources[1].id = https://doi.org/10.13026/h995-bt35
    file_collections[3].download_url = …
```

`distribution_formats[].access_urls` and `file_collections[].download_url` pointing at the pediatric release say *you obtain this dataset there*. That is the #292 shape one level down — not a record about the wrong dataset, but a record absorbing the other dataset into its own distribution. The `id` check, on its own, reassures without covering it.

`foreign_references()` now reports these, and **never as a verdict**: 32 existing records carry them, and not every placement is wrong — a VOICE record may legitimately cite the pediatric page as a resource, since that page is in its bundle.

**#442 — two identifier spellings slipped through.** `10.13026/h995-bt35` and the `dx.doi.org` forms both answered `ok`. Bare DOIs are not hypothetical; #402 measured 42% of nested ids in one clean-validating record as neither URI nor CURIE.

**#443 — paired declarations were not required to be symmetric.** VOICE and VOICE_PEDIATRIC declare each other only because both were written by hand at the same time. Now required when the other side has a scope block; a related dataset outside this corpus legitimately has none.

**#444 — one unparseable record aborted the sweep.** Reported as `unreadable` with the file named.

Also found in passing: the sweep globbed `*_d4d.yaml` only, reporting 16 records where the corpus holds 32. Core records are records; both halves now checked (329, not 171).

### What the review could not settle

Whether any given `resources[]` placement is a citation or a conflation. The check lists them; deciding is a judgement, and a rule that decided it would be the same overreach as the paragraph this PR replaces.

1487 passed, 4 skipped.

Issue #421, opened 2026-08-10
Bundles carry curator prose that instructs and asserts facts, unequally per project


realmarcin:
From the genericity audit (#418).

The BASELINE arm is described as "input documents only". The concatenated bundles also carry **curator prose**, injected by the preprocessing step into each source's header block:

```
SOURCE METADATA
Project: VOICE
Source ID: physionet_3_1_0
...
Curation note: Current PhysioNet release, published 2026-05-01, 833 participants.
A minor update over 3.0.0 with no new participants: additional data for some
participants, repaired parquet files, ... prefer this over physionet_3_0_0 where
the two disagree.
```

Counted across the five bundles:

| project | sources | curation notes in bundle | manual-fetch |
|---|---|---|---|
| AI_READI | 10 | **6** | 0 |
| CM4AI | 10 | **5** | 0 |
| VOICE | 11 | 3 | 0 |
| CHORUS | 4 | 1 | **1** |
| VOICE_PEDIATRIC | 6 | 1 | 0 |

## Two distinct problems

**They instruct.** *"prefer this over `dataset_documentation` where the two disagree"*, *"retain it only as evidence about the v2.0.0 release"*, *"treat this source as corroborating identity, size, license and keywords rather than as a substantive description"*. These are conflict-resolution rules addressed to the reader — exactly the kind of guidance the prompt condition is supposed to control, arriving through the input instead.

**They assert dataset facts.** The VOICE note above states *"published 2026-05-01, 833 participants"*. A record can take 833 from the curator rather than from PhysioNet, and `d4d evaluate verifiable` will score it grounded, because the string is in the bundle.

And the volume is unequal: AI_READI gets six, CHORUS one. So the arm that is meant to isolate "what the documents alone support" gives different editorial support per project.

## Why this is a decision, not a bug

The notes are genuinely useful — they are how the manifest records why a superseded source was retained. The question is whether they belong in the *bundle* (which the model reads) or only in the *manifest* (which the curator reads).

Options:

1. **Strip them at concatenation.** The bundle carries source identity and bytes; the manifest keeps the reasoning. Cleanest, and makes "documents only" true. Changes every bundle's md5, so it re-baselines every arm — this is the real cost.
2. **Keep them and rename the arm.** Honest, and free, but concedes that baseline includes curation.
3. **Keep them and equalise.** Worst of both: still editorial, and now editorial by policy.

[...]

Issue #397, opened 2026-08-07
Reasoning effort is absent from the generic arm prompt's header block, so every generic run loses it from provenance


realmarcin:
`d4d provenance record` builds its entire `model:` block by parsing `# Key: value` header lines out of the generated YAML (`provenance.py`, `---- model identity ----`). A field missing from the header is missing from provenance, permanently and silently — only a missing `Model` is flagged into `unrecoverable`.

`src/download/prompts/d4d_generic_arm_prompt.md` lists `{MODEL}` in its substitution table and header block but has no `Reasoning effort` line. `.claude/commands/d4d-full-core.md` does have it (`# Reasoning effort: {EFFORT}`). So a run launched from the generic prompt file, followed exactly, produces a record with no reasoning effort at all — and it will still validate, reconcile, and pass `d4d runs check --strict`.

This matters beyond one field: `Reasoning effort` is in `PROCEDURE_FIELDS` (`runs.py`), one of the fields that decides whether two runs are replicates or different procedures.

Caught on the CHORUS canary for `2026-08-07_claude-opus-5-claudecode-generic-v3_rep1` and patched for that label by pinning the field in the launch prompt, which avoided editing the shared template mid-series. The durable fix is to add `{EFFORT}` to the template's substitution table and header block.

Two things to decide when doing it:

- **It is a mechanical field, not tuning.** Adding it does not create a third condition, because it carries nothing project-specific. But the file's own editing rule says a change re-baselines the arm for every project, and it changes the prompt's sha256 against earlier generic-v3 runs. Worth deciding deliberately whether that's a new prompt version.
- **Pin the value's provenance.** `CLAUDE_EFFORT` is exposed in the Claude Code environment, so the agent can *observe* the effort rather than assert it. That is a stronger claim than `Temperature: 0.0`, which `provenance.py` already flags as `asserted by the generating agent, not observed`. The template should say to read the env var, not to write a literal.
- **Guard the wording.** `runs.py` records that past runs wrote `"default"` and `"not applicable"` for this field on identical configurations, which is why `PLACEHOLDER_VALUES` discards them. The template should forbid inventing a value when the variable is unset.

realmarcin:
Fixed in #445, merged as `4fc4d6b2`. All four CI checks green (3.10/3.11/3.12 + large-files).

PR #445, opened 2026-08-11
Take the last curator text out of the bundles; establish reasoning effort in the recorder (#427, #397)

### LINKED ISSUE #427: verification_url is still injected into bundles after #424 stripped curation notes
Raised reviewing #424 before merge. Filed rather than fixed there, with a measurement.

#424 strips `curation_note` out of the bundles because it is manifest metadata addressed to a curator. The same header block still carries `verification_url` from the same manifest, by the same mechanism:

```
SOURCE METADATA
Project: VOICE
Source ID: physionet_pediatric_1_1_0
...
Verification URL: https://physionet.org/content/b2ai-voice-pediatric/
```

Seven occurrences across the five bundles — AI_READI 3, CM4AI 2, VOICE 1, VOICE_PEDIATRIC 1, CHORUS 0. Same unequal-per-project shape as the notes had.

## Why it was not stripped in #424

Measured: **0 values in the canonical sweep are grounded only by a `Verification URL:` line.** Removing every one of them from the bundles and re-scoring changes nothing, because each of those URLs is independently present in the source documents — they are DOIs and landing pages the documents themselves carry.

And unlike a curation note, a bare URL neither instructs ("prefer this over X") nor asserts a dataset fact (VOICE's note stated "833 participants"). Those two properties are what made the notes load-bearing, and this line has neither.

So the exposure is **cosmetic, not material**, and stripping it would change five bundle hashes a second time for no measurable effect — while #424 already changes them once and 55 provenance records reference the old ones.

## The question worth deciding

It is still manifest metadata inside a file described as "input documents only", and leaving it makes the rule "no manifest metadata in the bundle, except this one kind". Either:

1. **Strip it too**, batched with some other change that re-baselines bundles anyway, so the hashes move once more rather than twice.
2. **Keep it and say why** — a source's canonical location arguably *is* part of identifying the source, in a way its curator's opinion is not.

Recommend (2) with a line in the manifest's `selection_policy`, since the measurement says it costs nothing and the identity argument is real. But it should be a decision rather than an omission.

realmarcin:
Closes #427 and #397 — the two things you flagged as needing to happen before a new sweep, in that order because the first changes the inputs.

## #427 — the last curator text in the bundles

#421 removed `curation_note` and left `verification_url` beside it. 7 lines across four of the five document bundles, unequally — three for AI_READI, none for CHORUS. The URL records where a curator checked a capture against upstream; it appears in no source document, so a record could cite it and still be scored grounded. That is how `10.60775/fairhub.4` reached four values through the note (#424): the channel is the defect, not the field.

**Canary first.** Re-running the preprocessor *unchanged* reproduced all 35 files byte-identically, so the post-change diff is attributable to the change. It removes exactly one line from each of six files; the bundles differ only by those lines and the `SIZE:` figures the concatenator computes.

**Rebuilding the crate bundles found the larger half of the problem.** `_preprocessed_with_crate.txt` embeds the document bundle verbatim and had not been rebuilt since #421 — so the de novo arm was still being handed 9 curation notes that the baseline arm stopped seeing on 2026-08-10. The strip had been **arm-dependent for a day** and nothing said so.

| bundle | removed |
|---|---|
| CHORUS with_crate | 1 curation note |
| CM4AI with_crate | 5 curation notes, 2 verification URLs |
| VOICE with_crate | 3 curation notes, 1 verification URL |
| AI_READI, CM4AI, VOICE, VOICE_PEDIATRIC docs | 7 verification URLs total |

No bundle under `concatenated/` now contains either string.

### Hashes

`bundle_hash_history` becomes a list of events — #421's mapping intact, #427's added with the crate bundles alongside. Where existing records point:

```
55  pre-#421 document bundles
 9  pre-#427 crate bundles          (the de novo arm runs)
12  crate_only / healthsheet_only   unchanged, still verify on disk
89  no bundle_md5 recorded at all
```

No record cites a hash that maps to nothing.

**Comparability:** any new de novo run consumes different bytes from those 9, exactly as #421/#424 did for the baseline arm. `check_replicate` raises `ReplicateMismatch` on differing input fingerprints, which is the designed behaviour, not a surprise to work around.

## #397 — reasoning effort

`{EFFORT}` appears in no generic prompt, v1–v4. Three sources now, in order:

1. **the model route** — the provider expresses effort as a name suffix (`google/claude-opus-5-high`), so the route is evidence; recorded as observed
2. **`--reasoning-effort`** — recorded as asserted by the launcher, and still listed under `unverified`
3. **nothing** — the field stays absent and the gap is named

Fixed in the recorder rather than the prompt because a header line added to v3 would change its bytes, re-baseline the condition for every project, and now require a pin rotation (#432).

The rule underneath: **an effort nobody chose is not a value.** A test asserts no record ever contains "default", "unspecified" or "n/a". A run that did not choose an effort is a different claim from a run whose effort is unknown, and neither is a run at high.

Effort is not an attesting field, so naming the gap downgrades nothing — tested explicitly.

## Guard tests

Two failed and were re-pinned rather than loosened: the VOICE bundle hash (both superseded values now kept, and cross-checked against the manifest so the two records cannot disagree), and the preprocessor test, which asserted the URL *was* written and now asserts it is not.

1500 passed, 4 skipped. `d4d runs check --strict`, `d4d download scope --check --strict`, `d4d api prompts check --strict` all pass; `audit-manifest` still reports 35/41.


realmarcin:
## Adversarial review

Three findings. Two fixed here; one filed and deliberately left.

**#446 — nothing detects a stale derived bundle**, which is the root cause of what this PR fixed by hand. The crate bundles embed the document bundle verbatim, were not rebuilt after #421, and so fed the de novo arm 9 curation notes the baseline arm had stopped seeing — for a day, with both described as the same corpus.

`d4d download audit-bundles` rebuilds each into a temp file and compares. Canaried against the checker itself:

```
$ printf '\n# an edit nothing rebuilt\n' >> data/preprocessed/concatenated/CM4AI_preprocessed.txt
$ d4d download audit-bundles --strict
📦 8 derived bundle(s) rebuilt and compared
   ❌ stale  data/preprocessed/concatenated/CM4AI_preprocessed.txt
      rebuild: d4d download concatenate --project CM4AI
   ❌ stale  data/preprocessed/concatenated/CM4AI_preprocessed_with_crate.txt
      rebuild: d4d rocrate bundle --project CM4AI
exit 1
```

The second line is the propagation failure itself — editing the document bundle makes the crate bundle stale. Restoring the bytes restores the verdict, hash-verified.

Rebuild-and-compare rather than mtime, because `crate_only` and `healthsheet_only` are legitimately older than the document bundles and an mtime rule would cry wolf on three files today. `build_crate_bundle` gained an `out_path` so the check never writes over what it is checking; a test asserts the real bundle is untouched, and another asserts the builder is deterministic, without which rebuild-and-compare would report everything stale forever.

**#447 — the `unverified` list is written by every record and read by no command.** 77 records carry one, and #397 was about to add a second field to a list nobody sees. `d4d runs check` now prints it:

```
ⓘ  77 value(s) across these runs were recorded but not observed:
     model.temperature            77
```

Never fatal. The test for this says what it means rather than what would be convenient: the fixture *does* fail `--strict`, for the unrelated and correct reason that it has never been validated (#396), so the assertion is that no failure line is about an unobserved value — a blanket exit-0 assertion would have passed for the wrong reason and then broken later.

**#448 — 49 records could carry an observed reasoning effort** recoverable from `google/claude-opus-5-high`. Filed, not fixed: populating them rewrites 49 files and belongs in its own reviewable commit, not as a side effect of this one.

### Checked and clean

The crate builder is deterministic (built twice, identical md5), so the "after" hashes recorded in `bundle_hash_history` are reproducible by anyone.

1509 passed, 4 skipped. `audit-bundles --strict`, `runs check --strict`, `scope --check --strict`, `api prompts check --strict` all pass.

realmarcin:
## Adversarial review

Read-only pass over the diff, separate from writing it. Four findings, all filed; three fixed here in `98518207`.

| # | finding | status |
|---|---|---|
| #449 | Claude Code effort marked "asserted, not observed" although the runtime exposes `CLAUDE_EFFORT` | **fixed** |
| #450 | `--reasoning-effort` accepted the placeholder values its own help forbade | **fixed** |
| #451 | a route silently discarded a disagreeing `--reasoning-effort` | **fixed** |
| #452 | 64 records reference bundles whose bytes have since changed, undetected | filed, not fixed |

**#449** is the substantive one. The code asserted that the Claude Code runtime exposes nothing to read effort from; it exposes `CLAUDE_EFFORT`, and the 2026-08-07 sweep's agents obtained their header value by reading it. The fix corroborates rather than substitutes — `d4d provenance record` may run in a different session than the generator, so the environment can confirm a header value but never supply or overwrite one, and a disagreement is reported instead of resolved.

**#452 is deliberately left open.** This PR adds `audit-bundles`, which detects a bundle gone stale against its sources. One layer up the same staleness exists and nothing detects it: a record pins `inputs.bundle_md5` and asserts *"verified identical to the bytes consumed"*, and nothing re-checks that against the file. Measured today — 12 records still match, **64 no longer do**, 86 never recorded a hash. Most of that is #421's doing and this PR adds `verification_url` to it; both strips are correct, and the defect is that a corpus-wide input change lands unreported. `check_replicate` already compares record-to-record and will correctly refuse to call a post-#445 run a replicate of a pre-#445 one, which protects new work and says nothing about the 64.

**On scope.** The title names #427 and #397, but the branch carries a second commit covering #446 and #447. Worth reading as a four-issue PR.

**Tests:** 672 pass across `test_rocrate/`, `test_cli/`, `test_download/`, `test_provenance_reasoning_effort.py`, `test_prompt_registry.py`. The effort suite goes 14 → 21. One pre-existing test asserted the old #449 behaviour and was silently environment-dependent — it passed here only because `CLAUDE_EFFORT=high` happened to match its fixture; it now controls the variable explicitly.

realmarcin:
## Second review round — a CI failure that was telling the truth

`audit-bundles` passed locally and reported `CM4AI_preprocessed_with_crate.txt` **stale** on all three CI Pythons. The bundle was current; the checkout was not.

```
.gitignore:152   data/ro-crate_packages/*/crate/
```

The de novo bundle includes `ai_ready_score.json` from inside that ignored directory, so a clean clone rebuilds from fewer artifacts. Reproduced locally by moving the directory aside:

```
· unchecked CM4AI_preprocessed_with_crate.txt: this checkout is missing crate
  artifacts the bundle was built from: ai_ready_score.json
```

The audit now compares **inputs before bytes** — the bundle's header lists what it was built from, which is the only thing that separates "this file is out of date" from "this machine has fewer inputs than the one that built it". The crate-staleness test skips where the inputs are absent: 8 passed 1 skipped with the directory hidden, 9 passed restored.

**#453 is filed and left open**, because the reporting fix does not touch the finding: the de novo arm's input bundle cannot be reproduced from a clean clone, and nine provenance records name its md5. Three options are laid out there; only committing the ~6 KB artifact makes the arm reproducible, and that is a call about what belongs in the repo.

## Correction to this PR's earlier claim

My first commit recorded every Claude Code reasoning effort as *"asserted by the generating agent, not observed"*, reasoning that the runtime exposes nothing to read. **That was wrong** — `CLAUDE_EFFORT` exists and is exactly where the 2026-08-07 agents were told to read it. Corrected in `98518207` (#449): the environment can now corroborate a header value but never supply or overwrite one, since the recorder may run in a different session than the generator, and a disagreement is reported rather than resolved. The test that encoded my wrong claim was replaced by three that name the environment explicitly.

Also corrected there: `--reasoning-effort` accepted the placeholder values its own help forbids (#450), and a flag disagreeing with the route was silently discarded (#451).

1518 passed, 4 skipped; CI green on 3.10, 3.11 and 3.12.

PR #455, opened 2026-08-11
Audit identifier syntax: the trap that passes validation (#402)

### LINKED ISSUE #402: uriorcurie slots accept arbitrary bare tokens: 42% of nested ids in a clean-validating label are neither URI nor CURIE

realmarcin:
Closes #402.

## What

`d4d runs identifiers` — a corpus audit of identifier syntax, sibling to `trap-inventory`.

```
🔍 341 record(s), 23378 identifier(s), 33 declared prefix(es)
     8168  absolute IRI
     8086  CURIE on a declared prefix
     3510  CURIE on an undeclared prefix
     3614  bare token — neither IRI nor CURIE

⚠️  7124 identifier(s) (30%) across 124 record(s) resolve to nothing
```

## Why it needed a new command

`trap-inventory` mines the corpus for validation **failures**. It cannot find this one, because these values pass. That is the point: `uriorcurie` declares no `pattern`, so LinkML renders it `{"type": ["string","null"]}` and `funder_nih` validates exactly as cleanly as `https://ror.org/01cwqze88`.

Nothing was looking, so four mutually incompatible conventions coexisted inside a single label — `2026-08-07..._rep1` is **42% unresolvable** — with no warning from any gate. The corpus now holds NIH as four different things:

- `funder_nih` (CHORUS, VOICE_PEDIATRIC)
- `https://cm4ai.org/data-releases/#funder-nih` (CM4AI)
- `d4d:VOICE-funding-nih-common-fund` (VOICE)
- no id at all, name only (AI-READI)

Nothing joins those, which defeats the purpose of a semantic exchange layer. A bare token is worse than an absent id: it looks like an identifier and is scoped to nothing, so two records both emitting `funder_nih` collide by accident rather than agree by reference.

## Design notes

**Four shapes, not two.** A CURIE on an *undeclared* prefix is well-formed and still unresolvable, and it is counted apart from a bare token because the remedy differs — declare the prefix, versus rewrite the value. 3,510 of the corpus's unresolvable ids are this kind, so collapsing the two would have pointed the work in the wrong direction.

**Prefixes come from the schema.** Read from the merged schema's own `prefixes:` block rather than a hardcoded list, so the audit cannot drift from what the schema would actually expand.

**Reports, never repairs.** Adding the pattern to `uriorcurie` is the correct end state and would invalidate values in records already committed — a migration, not a flag flip. Naming what is out there comes first. `--strict` is opt-in for anyone who wants the gate today.

**`summarize()` is separate from `audit()`** so a filtered caller recomputes its headline. The first version printed a label-filtered record list beneath corpus-wide totals: "10 records, 23378 identifiers", two scopes in one breath, reading as one. Truncated offender lists say how many they dropped, for the same reason.

## Not in scope

The `pattern` itself, and the migration of the 7,124 existing values. #402 stays the reference for that decision; this PR makes the number visible and stable enough to plan against.

## Testing

`tests/test_identifiers.py` — 17 tests: the four classifications, nested/`[]`-normalised traversal, non-string ids ignored, ATTIC excluded by default, unparseable records reported rather than skipped, dominant-convention detection, and that `summarize` totals describe exactly the records handed to it.

250 pass across `test_rocrate/test_runs.py`, `test_cli/`, `test_identifiers.py`, `test_provenance_reasoning_effort.py`. Verified `--strict` exits 1 on a label with unresolvable ids.

### DIFF
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -98,15 +98,22 @@ def identifiers_cmd(label, method, output, show, strict):
     report = ident.audit(root=root)
     if label:
         # Recompute rather than filter in place: the headline must describe
-        # exactly the records it is printed above.
+        # exactly the records it is printed above. `slots_audited` describes
+        # the schema, not the selection, so it is carried across rather than
+        # recomputed — dropping it made the header claim only `id` was checked
+        # while the by-slot line below it named others.
+        audited = report.get("slots_audited")
         report = ident.summarize(
             [r for r in report["records"] if f"/{label}/" in r["path"]],
             report["prefixes_declared"], report.get("unreadable"))
+        report["slots_audited"] = audited
 
     rows = [r for r in report["records"] if r["offenders"]]
+    slots = report.get("slots_audited") or ["id"]
     click.echo(f"🔍 {len(report['records'])} record(s), "
                f"{report['identifiers']} identifier(s), "
                f"{report['prefixes_declared']} declared prefix(es)")
+    click.echo(f"   slots with range uriorcurie: {', '.join(slots)}")
     c = report["counts"]
     click.echo(f"   {c[ident.URI]:>6}  absolute IRI")
     click.echo(f"   {c[ident.CURIE_DECLARED]:>6}  CURIE on a declared prefix")
@@ -117,6 +124,10 @@ def identifiers_cmd(label, method, output, show, strict):
         click.echo(f"\n⚠️  {report['unresolvable']} identifier(s) "
                    f"({report['unresolvable_share']:.0%}) across {len(rows)} "
                    "record(s) resolve to nothing:")
+        by_slot = report.get("unresolvable_by_slot") or {}
+        if by_slot:
+            click.echo("   by slot: " + ", ".join(f"{s} {n}"
+                                                  for s, n in by_slot.items()))
         for r in rows:
             name = Path(r["path"]).name
             click.echo(f"   {name}  ({len(r['offenders'])} of {r['total']}, "
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -65,6 +65,86 @@ def telemetry_cmd(label_prefix, method, output, findings_path, do_validate):
         click.echo("✓ report validates against d4d_run_telemetry.yaml")
 
 
+@runs.command("identifiers")
+@click.option("--label", default=None, help="limit to one run label")
+@click.option("--method", default=None, help="limit to one method directory")
+@click.option("-o", "--output", type=click.Path(), default=None,
+              help="write the full per-record report here as YAML")
+@click.option("--show", default=8, type=int, show_default=True,
+              help="list up to N offending values per affected record")
+@click.option("--strict", is_flag=True,
+              help="exit non-zero if any identifier cannot be resolved")
+def identifiers_cmd(label, method, output, show, strict):
+    """Audit identifier syntax across records — the trap the validator misses (#402).
+
+    `trap-inventory` mines validation failures. This is its complement, and the
+    defect it finds is the more dangerous kind because it passes: `uriorcurie`
+    declares no pattern, so LinkML renders it `{"type": ["string","null"]}` and
+    a bare token like `funder_nih` validates exactly as cleanly as a ROR IRI.
+
+    Nothing is repaired here. Adding the pattern would invalidate values in
+    records already committed — the right end state, and a migration rather
+    than a flag flip. Naming what is out there comes first.
+    """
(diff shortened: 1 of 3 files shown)


realmarcin:
## Adversarial review (self)

Same standard as the pass on #445. Three findings, all in code I wrote, all fixed in `ced5b44d`.

**1. The PR closed #402 while auditing 1 of 6 `uriorcurie` slots.** The schema ranges six slots that way; I walked only `id`. The five I missed hold the worse values:

| slot | unresolvable | example |
|---|---|---|
| `unit` | **148 of 148** | `%`, `years` |
| `publisher` | 68 | `PhysioNet` |
| `latest_version_doi` | 31 | `10.60775/fairhub.3` |
| `data_substrate` | 22 | *"Multimodal per-participant records comprising…"* |
| `data_topic` | 22 | *"Type 2 diabetes mellitus and associated health outcomes"* |

Prose sentences in an identifier slot, and a `unit` slot that has never once held an identifier — a modelling error the audit reported as clean. The slot set is now derived from the schema by **induced** range, so a `slot_usage` narrowing some other slot to `uriorcurie` is picked up without anyone remembering to add it.

Corpus figure moves from 7,124 / 23,378 across 124 records to **7,415 / 24,477 across 157**, now broken out per slot — because the remedies are different sizes. A wholly-unresolvable slot is a modelling error; a mostly-clean slot with strays is data entry.

**2. `d4d:` classified as a valid CURIE.** The reference was `\S*`; a bare prefix expands to the namespace itself and identifies nothing. Now `\S+`.

**3. The `--label` path dropped `slots_audited`.** The header claimed only `id` was checked while the by-slot line directly beneath it named others — the same mixed-scope defect I'd already fixed once in this PR, in a second place. `slots_audited` describes the schema, not the selection, so it is carried across rather than recomputed.

**On the fixture.** The tests passed throughout, because the fixture was a prefixes-only stub that could not exercise any of this. It is now a loadable LinkML schema declaring a non-`id` `uriorcurie` slot, so the coverage is tested rather than assumed. 17 → 25 tests; 258 pass across the wider suite.

Scope unchanged: still reports, never repairs. The `pattern` on `uriorcurie` and the migration of the 7,415 values stay with #402 as the follow-on decision.

Issue #454, opened 2026-08-11
The 2026-08-07 agentic sweep is labelled generic-v3 but ran the generic prompt: 15 runs, 30 records


realmarcin:
Filed so this stops living only in a conversation. Surfaced by the #420 check, which reports it on every one of the fifteen runs:

> label claims `generic_v3` but the hashed prompt is `generic` (`d4d_generic_arm_prompt.md`); the two are different conditions

## What happened

The label `2026-08-07_claude-opus-5-claudecode-generic-v3_rep{1,2,3}` was chosen to sit alongside the earlier `2026-08-05`/`2026-08-06` generic-v3 series. The runs actually consumed `src/download/prompts/d4d_generic_arm_prompt.md` (`0fbc626b…`), which is the plain **generic** condition — `d4d_generic_arm_prompt_v3.md` is a separate, separately-pinned file. `d4d-full-core.md` names the unsuffixed path as the generic prompt, so following the playbook and copying the older series' naming produced a label that contradicts the prompt.

Scope: 15 runs × (full + core) = **30 records**, across three label directories and two method directories.

Not fatal. `d4d runs check --strict` reports `5 subject, 0 failing, exit 0` for each label; the live-provenance requirement holds and every record validates. But anything grouping runs by config name places these with a series they did not share a prompt with.

## Worth noting about how it was caught

Nothing could have caught it before #395. The label was only checkable against the prompt once the records began hashing the prompt they consumed — the check exists because the data does.

## Two defects a rename would not fix

**1. No recorded instruction.** These predate #419, so their provenance carries no `instruction` / `request` / `render` block. The launch text cannot be audited against them.

**2. The label is not one uniform condition.** VOICE and VOICE_PEDIATRIC received a scope paragraph in their launch text that AI_READI, CHORUS and CM4AI did not — the practice #422 removed, and which CLAUDE.md now cites this exact run for. Renaming all fifteen to `generic` would assert a uniformity that does not hold for two of five projects: a visible wrong label traded for an invisible one.

## What the runs are still good for

The non-uniformity is *between projects*, not between replicates — all three replicates of a given project got byte-identical instructions. So the replicate-variance result stands on its own: within-project sd **3.47** slots, mean Jaccard **0.858**, between-project sd **10.91**. They are also clean on the new scope check: no `out_of_scope` verdict, and none of the 32 records that place the pediatric release outside its declared slot.

## Options

[...]

PR #459, opened 2026-08-11
Register the v1-vs-v3 comparison, and correct what §6 claimed had run (#458)

### LINKED ISSUE #458: The v2-vs-v3 comparison is not derivable: v3 never ran on 3 of 5 projects, and every v3 label is confounded with schema or route
Found while scoping the §6 promotion decision. **The v2-vs-v3 comparison is not merely unwritten — it is not derivable from any run on disk**, and NEXT_TASKS §6 currently overstates what exists.

§6 says v3 "has 20 runs across four labels". Fifteen of those twenty are the 2026-08-07 sweep, which #454 establishes ran the *generic* prompt, not v3. Here is every run that actually hashed `d4d_generic_arm_prompt_v3.md`, against the v2 series it would be compared with:

| label | prompt | schema digest | route | coverage |
|---|---|---|---|---|
| `2026-07-31_…generic-v2_rep{1,2,3}` | v2 | `34d24ff3` (1.0.0) | `google/claude-opus-5-high` | 4 projects × 3 |
| `2026-08-05_…generic-v3_rep1` | v3 | `b065e1cd` (1.0.0) | `google/claude-opus-5-high` | AI_READI, CHORUS × 1 |
| `2026-08-05_…1m-generic-v3_rep{1,2,3}` | v3 | `b065e1cd` (1.0.0) | `claude-opus-5` (1m) | AI_READI × 3, CHORUS × 1 |
| `2026-08-06_…1m-generic-v3-schema2_rep{1,2,3}` | v3 | `583d79c1` (2.0.0) | `claude-opus-5` (1m) | AI_READI, CHORUS × 2 |
| `2026-08-07_…claudecode-generic-v3_rep{1,2,3}` | **generic (v1)** | 2.0.0 | `claude-opus-5` | 5 projects × 3 |

Three independent defects, any one of which sinks the comparison:

**1. Coverage.** CM4AI, VOICE and VOICE_PEDIATRIC have **never run under v3**. Real v3 coverage is AI_READI at three replicates and CHORUS at one.

**2. Schema confound.** Every v3 label sits on a different schema digest than the v2 series (`b065e1cd` or `583d79c1` vs v2's `34d24ff3`). The schema delta between them includes the trap-slot work (#376 scalarized the narrative family, #382 made `Organization.id` optional) — which mechanically moves the *exact* defect classes `form_defects` counts. A hollow-object count cannot be compared across it.

**3. Route confound.** Three of the four v3 labels ran `claude-opus-5` (1m) where v2 ran `google/claude-opus-5-high`.

`notes/generic_v3_analysis_plan.md` pre-registered the requirement: *"Same four projects, three replicates, same model and settings as the v1 and v2 series."* **No label meets it.**

## The instrument already enforced this

All 1441 cached fitness judgements are keyed `(34d24ff3, google/claude-opus-5-high)` — v2's exact configuration:

```
('34d24ff3', 'google/claude-opus-5-high') 1441
```

No v3 record shares that key, so scoring one was never a cache hit and never silently mixed arms. The guard §6 describes worked exactly as designed. What it could not do is announce that the arm it was guarding had never been generated — absence of a cache entry is indistinguishable from absence of a run.

## Consequence

§6's promotion decision needs generation, not a writeup. Filing separately from #454 because the two have different remedies: #454's records are mislabelled but internally honest (their headers name the generic prompt correctly), whereas this is a missing arm.

Also worth correcting in NEXT_TASKS §6: the "20 runs across four labels" line, and the description of the 2026-08-07 set as the v3 "canonical set".

realmarcin:
Closes #458. Corrects what NEXT_TASKS §6 and §9 claim, and registers the comparison that replaces the one that cannot be run.

## What was wrong

§6 said v3 had "20 runs across four labels". Fifteen of those are the 2026-08-07 sweep, which hashed the **generic** prompt, not v3 (#454). Every run that did hash v3 is confounded against the v2 series it would be compared with:

| label | prompt | schema digest | route | coverage |
|---|---|---|---|---|
| `2026-07-31_…generic-v2_rep{1,2,3}` | v2 | `34d24ff3` (1.0.0) | `google/claude-opus-5-high` | 4 projects × 3 |
| `2026-08-05_…generic-v3_rep1` | v3 | `b065e1cd` (1.0.0) | `google/claude-opus-5-high` | AI_READI, CHORUS × 1 |
| `2026-08-05_…1m-generic-v3_rep{1,2,3}` | v3 | `b065e1cd` (1.0.0) | `claude-opus-5` (1m) | AI_READI × 3, CHORUS × 1 |
| `2026-08-06_…1m-generic-v3-schema2_rep{1,2,3}` | v3 | `583d79c1` (2.0.0) | `claude-opus-5` (1m) | AI_READI, CHORUS × 2 |
| `2026-08-07_…claudecode-generic-v3_rep{1,2,3}` | **generic (v1)** | *none recorded* (2.0.0) | `claude-opus-5` | 5 projects × 3 |

Three independent defects, any one sufficient: CM4AI/VOICE/VOICE_PEDIATRIC never ran under v3; every v3 label sits on a different schema digest than v2, and the delta includes the trap-slot work (#376, #382) that moves the exact defect classes `form_defects` counts; three of four v3 labels ran a different route.

The pre-registered plan asked for "same four projects, three replicates, same model and settings as the v1 and v2 series." No label meets it.

## Why nothing was corrupted

All 1441 cached fitness judgements are keyed `(34d24ff3, google/claude-opus-5-high)` — v2's exact configuration. No v3 record was ever a cache hit, so no arms were silently mixed. The guard worked as designed. What it could not do is announce that the arm it protected had never been generated: an absent cache entry and an absent run look identical from the instrument's side.

## What this PR does

- **Registers `notes/generic_v1_v3_analysis_plan.md`** before any generation, keeping the prediction outside the prompt for the same reason `generic_v3_analysis_plan.md` does.
- **Corrects §6** with the disk facts and records the decision: both arms fresh, 5 projects × 3 replicates = 30 runs, at today's digest `e802cdc3` — which no run on disk carries.
- **Adds the #454 caveat to §9**, whose canonical set is that label. The records are internally honest — every header names `d4d_generic_arm_prompt.md` and says "generic prompt" — so only the directory label misstates the condition.

## Why v1-vs-v3, not v2-vs-v3

v1 is the playbook default, so v1-vs-v3 is the comparison the *promotion* decision rests on. It measures all four rules at once, which is the right granularity for a decision to adopt all four. The mechanism question — does rule 4 alone do what it says? — stays open and needs a v2 arm at the same digest; this plan does not produce one and says so.

## Not in this PR

The 30 generation runs themselves. Docs-only change; the plan is the gate that has to land first.

### DIFF
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -207,23 +207,33 @@ This section said v3 was "drafted, wired and tested but unrun", then said it had
 every run that actually hashed `d4d_generic_arm_prompt_v3.md`, against the v2
 series it would be compared with:
 
-| label | prompt | schema digest | route | coverage |
+Replicate counts are per project, measured from the provenance records.
+
+| label | prompt | schema digest | route | replicates |
 |---|---|---|---|---|
-| `2026-07-31_…generic-v2_rep{1,2,3}` | v2 | `34d24ff3` (1.0.0) | `google/claude-opus-5-high` | 4 projects × 3 |
-| `2026-08-05_…generic-v3_rep1` | v3 | `b065e1cd` (1.0.0) | `google/claude-opus-5-high` | AI_READI, CHORUS × 1 |
-| `2026-08-05_…1m-generic-v3_rep{1,2,3}` | v3 | `b065e1cd` (1.0.0) | `claude-opus-5` (1m) | AI_READI × 3, CHORUS × 1 |
-| `2026-08-06_…1m-generic-v3-schema2_rep{1,2,3}` | v3 | `583d79c1` (2.0.0) | `claude-opus-5` (1m) | AI_READI, CHORUS × 2 |
-| `2026-08-07_…claudecode-generic-v3_rep{1,2,3}` | **generic (v1)** | *none recorded* (2.0.0) | `claude-opus-5` | 5 projects × 3 |
+| `2026-07-31_…generic-v2` | v2 | `34d24ff3` (1.0.0) | `google/claude-opus-5-high` | AI_READI 3, CHORUS 3, CM4AI 3, VOICE 3 |
+| `2026-08-05_…generic-v3` | v3 | `b065e1cd` (1.0.0) | `google/claude-opus-5-high` | AI_READI 1, CHORUS 1 |
+| `2026-08-05_…1m-generic-v3` | v3 | `b065e1cd` (1.0.0) | `claude-opus-5` (1m) | AI_READI 3, CHORUS 1 |
+| `2026-08-06_…1m-generic-v3-schema2` | v3 | `583d79c1` (2.0.0) | `claude-opus-5` (1m) | AI_READI 3, CHORUS 3 |
+| `2026-08-07_…claudecode-generic-v3` | **generic (v1)** | *none recorded* (2.0.0) | `claude-opus-5` | all five, 3 each |
+
+v3 totals: **AI_READI 7, CHORUS 5, CM4AI 0, VOICE 0, VOICE_PEDIATRIC 0.**
 
 **The v2-vs-v3 comparison is not merely unwritten — it is not derivable.** Three
 independent defects, any one sufficient:
 
-1. **Coverage.** CM4AI, VOICE and VOICE_PEDIATRIC have never run under v3.
+1. **Coverage.** CM4AI, VOICE and VOICE_PEDIATRIC have never run under v3, so no
+   comparison covering the corpus is possible.
 2. **Schema confound.** Every v3 label sits on a different digest than v2's
    `34d24ff3`, and the delta includes the trap-slot work (#376, #382) that moves
    the exact defect classes `form_defects` counts.
 3. **Route confound.** Three of four v3 labels ran `claude-opus-5` (1m) where v2
    ran `google/claude-opus-5-high`.
+
+Note the shape of 2 and 3 together: the only v3 series with three replicates on
+two projects is `2026-08-06_…1m-generic-v3-schema2`, which is confounded against
+v2 on **both** axes at once. The best-powered v3 data is also the most
+confounded (#460).
 
 `notes/generic_v3_analysis_plan.md` pre-registered "same four projects, three
 replicates, same model and settings as the v1 and v2 series." No label meets it.
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -273,7 +273,9 @@ Guards in place:
   measured against.
 - Tests assert v2 differs from v1 *only* within the marked block, that the block
   names no project, dataset identifier or quantity, and that every project
-  receives byte-identical text after mechanical substitution.
+  receives byte-identical text after mechanical substitution — **for the four
+  projects the test hardcodes.** VOICE_PEDIATRIC is not among them, so the
+  byte-identity guarantee is unasserted for the fifth project (#467).
 - The prediction is deliberately absent from the prompt. Written there it would
   instruct the model to produce the result the run is meant to test, which the
   priming taxonomy excludes from both arms.
@@ -356,7 +358,18 @@ which is why no VOICE replicate validated (#292).
 
 **Done:** the registry, the shared-corpus declaration, and the schema correction
 that made the nested form wrong rather than ambiguous. Adding the fifth project
-broke nothing; the four-project list in 26 files is prose, not logic.
+broke nothing *visible*, and the four-project list is prose in most of the 26
+files that carry it.
+
+⚠️ **Not in all of them.** `form_defects._value_index` hardcoded the four as a
+runtime list, so any VOICE_PEDIATRIC failure would have attributed as
+`unattributed` — silently, since nothing distinguishes "a project we did not
+open" from "a value that matched nothing". Fixed to read `PROJECTS` (#463).
+Two four-project literals are still live and are *not* prose:
+`agreement.DEFAULT_PROJECTS`, and the byte-identity assertion in
+`tests/test_download/test_generic_v2_prompt.py` (#467). "Broke nothing" held
+because nothing from the fifth project had reached those code paths yet, not
+because they were generic.
 
 **Not done, and each needs a decision:**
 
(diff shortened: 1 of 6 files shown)


realmarcin:
## Review rounds 1–3, and what they turned up

Three rounds of adversarial review on this PR, each finding something real. All fixed in this branch.

**Round 1 — #460, an error in my own table.** The `× N` coverage notation conflated projects with replicates, and the 2026-08-06 row was wrong outright. Measured per project:

```
2026-08-05_…generic-v3             AI_READI 1, CHORUS 1
2026-08-05_…1m-generic-v3          AI_READI 3, CHORUS 1
2026-08-06_…1m-generic-v3-schema2  AI_READI 3, CHORUS 3

v3 totals: AI_READI 7, CHORUS 5, CM4AI 0, VOICE 0, VOICE_PEDIATRIC 0
```

So "CHORUS at one replicate" was wrong — it has three. The correction *sharpens* the finding: the only v3 series with three replicates on two projects is the schema-2 series, confounded against v2 on **both** schema and route at once. The best-powered v3 data is also the most confounded.

**Round 2 — #461, the repo carries two different numbers for the same defect.** NEXT_TASKS §6 quoted collapsed cardinality as 27 → 0 and hollow objects as 2 → 33. Those are the *manual read* that `notes/form_defect_split_2026-08-03.md` explicitly superseded with a classifier. Worse, the note's headline figures (42 → 7, 8 → 50) appear in neither of its own tables, because they fold the `both` bucket in — a convention written down nowhere and implemented nowhere. An arm counted from the raw table compares against a baseline of 34 rather than 42, wrong by 8 values on v1 and 5 on v2, which is larger than several of the effects this comparison is meant to see. `folded()` now implements it, printed alongside the raw table rather than replacing it.

**Round 3 — #462, the measurement instrument was quietly broken.** `form_defects --offline` raised `OfflineCacheMiss` while its cache held all 106 labels, because the classifier resolved the *live config pin* (`claude-opus-5`) instead of the instrument the cache records (`google/claude-opus-5-high`). 106 of 106 entries out of scope. `recorded_model()` was written for this exact failure, names it in its docstring, and was called by nothing.

The offline break was the mild symptom. **Without** `--offline` the command succeeds — 106 paid calls, a second instrument appended to the cache, after which `recorded_model` refuses the file and the v1→v2 baseline can no longer be reproduced. This directly threatened the plan this PR registers, which names `form_defects` as its instrument: the v3 arm would have been classified by a different model than its baseline.

Now verified end to end rather than by exit code:

```
instrument: google/claude-opus-5-high
0 call(s), 106 cached
collapsed_cardinality  34   2      folded:  42   7
hollow_object           0  45               8  50
```

That reproduces the published table and both headline figures exactly, at zero cost.

Full suite: **1553 passed, 4 skipped, 0 failed.**

## Scope note

Rounds 2 and 3 grew this beyond a docs-only change into `src/data_sheets_schema/form_defects.py` and its tests. That is deliberate: both defects sit directly under the measurement the registered plan depends on, and shipping the plan while its instrument silently compared across instruments would have been the more disruptive choice.

realmarcin:
## Round 6 — Codex adversarial review

Five hand rounds each returned a real defect, which is the threshold for escalating. Codex traced the data flow rather than reading docstrings and found **three real defects in code this branch had just added**, plus four documentation claims that do not survive checking.

### Defects in new code

**#464 — the pooled-cache fallback, which I had reasoned my way into.** `_default_model` swallowed both of `recorded_model`'s `ValueError` cases. They are opposite situations: *nothing recorded* may fall back to the live pin; *more than one recorded* must refuse. My comment on the line said "already pooled — fall through" and my test asserted construction succeeded, on the reasoning that `recorded_model` would surface the error where it is documented.

Nothing on the production path calls `recorded_model`. So a pooled cache loaded only the entries matching whichever model the pin happened to name and printed their counts as the table — a partial result presented as whole, with my test encoding it as expected behaviour. This is the same shape as the two guard tests §11 found asserting the bug they guarded.

**#466 — the plan's own instrument could not measure the plan's arms.** `attribute()` was wired to the historical v1/v2 labels with no override. Every fresh failure would have read as `unattributed`, or attributed to an old arm wherever an identical slot/value occurs in both. Demonstrated on the current corpus:

```
subtype               unattributed     v1
collapsed_cardinality      2     34
hollow_object             45      0
total                     56     50
```

**`--model` could pool silently.** `main()` now refuses before doing any work, with `--allow-pooled-cache` as the stated act.

### Test quality

Codex was right that my three PROJECTS tests restated the implementation — they compared `_value_index`'s *default argument* against `PROJECTS`, which passes unchanged if the function ignores the argument entirely. Replaced with fixtures on disk that index a record per project and assert each attributes, plus the negative case so the guard is known to be able to fire.

### Documentation claims that did not hold

- **The plan said the form total is the sum of the two folded counts.** It is neither: the total also counts `other`, and the folded pair overlaps on `both`. The total can rise while both components fall — predictions 1–3 can come apart without any being wrong.
- **The plan said cached labels stay valid "for their own key".** The sub-type key is `(slot, value)` — no schema, unlike the fitness cache. A label survives a schema change that redefines it, and `Organization.id` becoming optional (#382) is exactly such a change. Filed #465; the plan now works around it with a separate cache file.
- **§9 said the four-project list is "prose, not logic".** Two live literals are logic: `agreement.DEFAULT_PROJECTS`, and the byte-identity test asserting "every project" over a hardcoded four — a guard whose stated scope is wider than its actual scope, in precisely the area #419/#422 were about. Filed #467.
- **§9 said the 2026-08-07 sweep ran "under v3"; §12 said its records were already superseded.** Neither holds. It hashed the generic prompt, and the fresh arm is decided but not run — `e802cdc3` appears in no record and no judgement. That second one was my own overclaim, added in this PR.

Full suite: **1563 passed, 4 skipped, 0 failed.**

Issues filed across all six rounds: #460, #461, #462, #463, #464, #465, #466, #467. Fixed here: #460–#464, #466. Left open with the reasoning recorded: #465 (cache key migration), #467 (touches published §7 figures).

Issue #447, opened 2026-08-11
The `unverified` list is written by every record and read by no command


realmarcin:
From the #445 review.

`build_record` writes an `unverified` list — fields recorded but not observed, each with a reason. 77 records carry one today (all `model.temperature`), and #445 adds `model.reasoning_effort` to it.

No CLI command reads it. `grep -rn '"unverified"' src/data_sheets_schema/cli/` returns nothing; `check_provenance` does not look at it, and `attestation` deliberately does not gate on it. So the honest gap-naming that #397 introduces lands in a field nobody sees — which is the "a check nobody runs is documentation" problem in a different costume.

Suggest `d4d runs check` print a count, and the per-field breakdown under a flag:

```
158 run(s) checked … 
   ⓘ 89 record(s) carry values that were recorded but not observed:
        model.temperature       77
        model.reasoning_effort  12
```

Never fatal. The point of the list is that these values are usable *with* a caveat, and a gate would collapse that back into a binary.

realmarcin:
Fixed in #445.

realmarcin:
Fixed in #445, merged as `4fc4d6b2`. All four CI checks green (3.10/3.11/3.12 + large-files).

Issue #446, opened 2026-08-11
Nothing detects a stale derived bundle; the crate bundles were a day behind the strip


realmarcin:
Found reviewing #445. This is the root cause of what that PR fixed by hand.

`{PROJECT}_preprocessed_with_crate.txt` embeds the document bundle verbatim. #421 stripped `curation_note` out of the document bundles on 2026-08-10 and nobody rebuilt the crate bundles, so for a day the de novo arm was still being handed 9 curation notes that the baseline arm had stopped seeing:

```
CHORUS with_crate   1 curation note
CM4AI  with_crate   5 curation notes, 2 verification URLs
VOICE  with_crate   3 curation notes, 1 verification URL
```

Two arms, two different input treatments, described everywhere as the same corpus. Nothing detected it. It surfaced only because #445 rebuilt the crate bundles for an unrelated reason and diffed the result.

## What would catch it

A staleness check over derived bundles: rebuild each into a temp path, compare to what is on disk, report any that differ. Every builder is deterministic — verified for `d4d rocrate bundle` by building twice and getting identical md5s — so a difference means the file is stale, never noise.

Candidates and their inputs:

| bundle | derived from |
|---|---|
| `{P}_preprocessed.txt` | `individual/{P}/*` per the manifest |
| `{P}_preprocessed_with_crate.txt` | the document bundle + normalized crate |
| `{P}_crate_only.txt`, `{P}_healthsheet_only.txt` | crate / healthsheet inputs only |

Mtime is not sufficient: `crate_only` and `healthsheet_only` are legitimately older than the document bundles because they do not derive from them, so an mtime rule would cry wolf on three files today.

Worth wiring into `d4d download audit-manifest`, which is already the "does disk match what is declared" command, or a sibling `audit-bundles`.

realmarcin:
Fixed in #445.

realmarcin:
Fixed in #445, merged as `4fc4d6b2`. All four CI checks green (3.10/3.11/3.12 + large-files).

Issue #439, opened 2026-08-11
The pin registry is re-parsed on every lookup


realmarcin:
From the #435 review. Cosmetic, filed so it is not rediscovered.

`pins()` reads and parses `canonical_hashes.yaml` on every lookup, and `canonical_prompt_status` calls it once per prompt file per record. A full `d4d runs check` over 158 runs re-reads the same small file several hundred times.

It is not slow enough to notice today (the check runs in a few seconds) and the file is under 4 KB. Worth an `lru_cache` keyed on path plus mtime rather than a plain cache, because `d4d api prompts pin` writes the file inside the same process as the tests that then read it — a cache without invalidation would make `test_rotation_keeps_the_previous_hash` pass for the wrong reason.

Issue #398, opened 2026-08-07
Prompt paths in provenance are recorded verbatim, so the same prompt can appear under two different path strings


realmarcin:
Review finding on #395, which adds `d4d provenance record --prompt`.

`prompt_facts()` records `{"path": str(p), ...}` using whatever path it was handed. Nothing normalizes it. Every other path-like field in the record is repo-relative by construction (`inputs.bundle_path`, `schema.full_path`, `outputs.*.path`) because they come from module constants or are built from them.

So a record written with `--prompt src/download/prompts/d4d_generic_arm_prompt.md` and one written with `--prompt /Users/.../data-sheets-schema/src/download/prompts/d4d_generic_arm_prompt.md` describe the same prompt with two different strings. The `sha256` still matches, so integrity checking is unaffected — but any analysis that groups or compares runs *by prompt path* sees two conditions where there is one, and the absolute form leaks a local filesystem layout into a checked-in-adjacent artifact.

The same applies to `api_runner.py`'s `spec.prompt_files`, which is not normalized either, so this is not specific to the new CLI flag.

Suggested fix: resolve prompt paths relative to the repo root in `prompt_facts()`, falling back to the absolute path with a note when the file is outside the repo. Doing it in `prompt_facts()` rather than at each call site fixes both paths at once.

Low severity — cosmetic and analysis-hygiene, not a correctness or integrity problem.

PR #468, opened 2026-08-11
Report records whose declared input bundle has since changed (#452)

### LINKED ISSUE #452: 64 records reference input bundles whose bytes have since changed, and nothing detects it
Review finding on #445, and the mirror image of #446.

#445 adds `d4d download audit-bundles`, which rebuilds each derived bundle and compares it to disk — closing the gap where a bundle derived from changed sources went stale undetected. That is the right fix at that layer. One layer up, the same staleness exists and is not detected.

A provenance record pins `inputs.bundle_md5` and asserts `hash_basis: verified identical to the bytes consumed`. Nothing ever re-checks that hash against the file at `inputs.bundle_path`. Measured across the corpus today:

| | records |
|---|---|
| recorded `bundle_md5` still matches the file | **12** |
| recorded `bundle_md5` no longer matches | **64** |
| no bundle hash recorded at all | 86 |

By bundle: CHORUS 18, AI-READI 16, CM4AI 9, VOICE 9, the three `_with_crate` variants 3 each, VOICE_PEDIATRIC 3.

The records are not wrong — each correctly states the bytes *it* consumed. What is missing is that the path no longer resolves to those bytes, so anyone re-reading a record's declared input reads something else. #421 caused most of this by stripping curator notes; #445 adds to it by stripping `verification_url`. Both strips are correct and neither is the defect. The defect is that the corpus absorbed a corpus-wide input change with no report.

`check_replicate` already does the analogous comparison **record-to-record** and will correctly refuse to call a post-#445 run a replicate of a pre-#445 one. That protects new work and says nothing about the 64.

Suggested fix: a `d4d runs check` line in the same shape as the new unobserved-values counter — count records whose `bundle_md5` no longer matches the file at their `bundle_path`, report and never fail. Non-fatal is right for the same reason it was right there: these records remain usable, they just cannot be re-derived from the path they name.

Worth deciding alongside: whether a bundle change should require re-pinning, the way a prompt change now requires a pin rotation under #432.

realmarcin:
Closes #452.

A provenance record pins `inputs.bundle_md5` and asserts `hash_basis: verified identical to the bytes consumed`. Nothing ever re-checked that against the file at `inputs.bundle_path`. `d4d runs check` now does:

```
ⓘ  64 record(s) name an input bundle whose bytes have since changed
   (12 still match, 82 record no hash):
     CHORUS_preprocessed.txt                      18
     AI_READI_preprocessed.txt                    16
     CM4AI_preprocessed.txt                        9
     VOICE_preprocessed.txt                        9
     VOICE_PEDIATRIC_preprocessed.txt             3
     CHORUS_preprocessed_with_crate.txt           3
     CM4AI_preprocessed_with_crate.txt            3
     VOICE_preprocessed_with_crate.txt            3
   Each record correctly states the bytes it consumed; the path no longer resolves to them.
```

That reproduces both the counts and the per-bundle breakdown the issue was filed on.

## The shape of it

The mirror of `audit-bundles` (#446) one layer up. That command asks whether a derived **bundle** still matches what its inputs produce; this asks whether a **record's declared input** still matches what that record consumed.

**Reported, never fatal** — verified `--strict` still exits 0. A drifted record is not wrong: it correctly states the bytes it read, and has only lost the ability to be re-derived from the path it names. That is a caveat, not a defect, and a gate would collapse the distinction — the same reasoning as the unobserved-values counter (#447).

#421 caused most of the drift by stripping curator notes; #445 added to it by stripping `verification_url`. **Both strips were correct.** The defect is that the corpus absorbed a corpus-wide input change with no report.

## Four outcomes, kept distinct

`current`, `drifted`, `absent` (the path is gone, so it cannot even be diffed), `unrecorded`. `unrecorded` is never counted as `current` — folding 82 hashless records into the matching count would report the corpus far healthier than it is, the same error as treating an absent cache entry as a hit.

## On 82 vs the issue's 86

Scope, not disagreement, and documented on the function. The four extra are the `guarded-union` derived merges: not runs, and they declare `bundle_md5` not-applicable by design (§3), because a derived record consumes replicates rather than a bundle and so has no input that can drift. All four are hashless, so the drifted and current counts are unaffected.

## Tests

Ten, including one pinning 64/12 against the real corpus — so the next corpus-wide input change is a visible test failure rather than a silent absorption, which is the entire defect. Also asserts that a check resolving the bundle from live config rather than from the record would miss the drift it exists to find.

Full suite: **1573 passed, 4 skipped, 0 failed.**

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -354,6 +354,25 @@ bundles embed the document bundle verbatim, and after #421 stripped curator
 prose they were not rebuilt — so the de novo arm read 9 curation notes the
 baseline arm no longer saw, for a day, with nothing to detect it.
 
+**One layer up, `d4d runs check` reports bundle drift** (#452): does the file at
+a record's `inputs.bundle_path` still hash to the `bundle_md5` that record
+pinned? Currently **64 records drifted, 12 current, 82 with no hash recorded**.
+
+The two checks are mirror images. `audit-bundles` asks whether a *bundle* still
+matches what its inputs produce; this asks whether a *record's declared input*
+still matches what that record consumed. A drifted record is not wrong — it
+correctly states the bytes it read — but the path it names no longer resolves
+to them, so anyone re-reading its declared input reads something else.
+
+Reported and never fatal, for the same reason as the unobserved-values counter
+(#447): these records stay usable, they just cannot be re-derived from the path
+they name. #421 caused most of the drift by stripping curator notes and #445
+added to it by stripping `verification_url`; **both strips were correct**. The
+defect was that the corpus absorbed a corpus-wide input change with no report.
+
+Derived records (`record_mode: derived`) are out of scope by construction: they
+consume replicates rather than a bundle and declare `bundle_md5` not-applicable.
+
 Each entry names the `referent_id` and any `related_but_distinct` dataset, with
 the slot that carries the relation (`express_as: related_datasets`) and, where
 the related dataset's documentation is legitimately in the bundle, the source id
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -384,8 +384,13 @@ because they were generic.
   including them would import adult facts into a pediatric datasheet, which is
   the failure the fitness axis calls `target`.
 
-  The VOICE bundle is byte-identical (md5 `e637eb75`, what 49 runs attest), and
-  a test pins it.
+  The VOICE bundle was byte-identical across the fork at md5 `e637eb75`, which
+  49 runs attest. **It is now `dcd71717`** — #421 stripped curator prose from
+  every bundle after those runs. `tests/test_voice_pediatric_bundle.py` pins
+  both, current and `_PRE_421`, so the transition is recorded rather than
+  overwritten. The sentence here previously named only `e637eb75` as the
+  bundle's hash, which stopped being true when the strip landed and was found
+  by the drift check below (#452).
 - **A generation run — done.** Both projects were generated in the 2026-08-07
   five-project sweep (`2026-08-07_claude-opus-5-claudecode-generic-v3_rep{1,2,3}`,
   PR #395). `VOICE_PEDIATRIC` was generated from its own 204 KB bundle as its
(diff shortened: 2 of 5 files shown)


realmarcin:
## Review rounds

**Round 1 — cost, and a deliberate "no" (#469).** `bundle_drift` hashes once per *record*, not once per *bundle*: 158 records against 12 distinct bundles, ~20× redundant, 0.77s for a full sweep. Same shape as #439, and the obvious fix is a path-keyed md5 cache.

Not doing it. A cache here is a correctness hazard in exactly the check it would speed up — this function exists to detect that a file's bytes changed, and a path-keyed cache answers `current` for a file that drifted after its first read. `(path, size, mtime_ns)` narrows the window without closing it: a same-size edit within one filesystem tick is plausible for generated bundles and silent, and the failure mode is a corpus-wide input change going unreported, which is the whole of #452. Recorded in a comment so it is not rediscovered as an oversight, and guarded by a test that calls twice either side of a same-length edit.

Two other theoretical concerns checked and dismissed against the corpus: **0** records carry an absolute `bundle_path` (the #398 hazard), and **no** bundle basename maps to more than one path, so the per-bundle grouping cannot collide.

**Round 2 — the check immediately found something.** Filtering to the canonical label reports **5 of 5 drifted, 0 current**:

| project | pinned | now |
|---|---|---|
| AI_READI | `8aadffca` | `20150c10` |
| CHORUS | `1ce7d891` | `9b2ef4b6` |
| CM4AI | `3694e188` | `1dfd34e5` |
| VOICE | `e637eb75` | `dcd71717` |
| VOICE_PEDIATRIC | `327c2920` | `008212ac` |

So the canonical set carries two independent caveats — a label naming the wrong condition (#454), and declared inputs that no longer re-fetch. §6's fresh arm clears both.

Chasing the VOICE row turned up a **stale claim in NEXT_TASKS §9**: it said the VOICE bundle is *"byte-identical (md5 `e637eb75`, what 49 runs attest), and a test pins it"*. The test is fine — `test_voice_pediatric_bundle.py` pins `dcd71717` as current and `e637eb75` as `_PRE_421`, recording the transition rather than overwriting it. The prose named only the old hash and stopped being true when #421 landed.

That is the check doing its job one layer above the records: the same corpus-wide input change had been absorbed into the documentation as well as the data, with nothing reporting either.

CLI filters exercised — per-project, per-label, and the empty-result case, which reports cleanly and exits 0.

Full suite: **1573 passed, 4 skipped.** All four CI checks green.

Issue #470, opened 2026-08-11
Two records carry reasoning_effort: default, the value CLAUDE.md explicitly forbids, on a route with no effort ladder


realmarcin:
Found while implementing #448's backfill pass, which surveys every record's `model.reasoning_effort`.

Two records carry `reasoning_effort: default`:

```
claudecode_agent_core/2026-07-27_claude-opus-5_rep1/VOICE_provenance.yaml
   route: claude-opus-5[1m] | basis: None
claudecode_agent_crate_core/2026-07-27_claude-opus-5_rep3/VOICE_provenance.yaml
   route: claude-opus-5[1m] | basis: None
```

CLAUDE.md forbids exactly this value, in the section that #397 wrote:

> Never write "default", "unspecified" or a guess. A run that did not choose an effort is a different claim from a run whose effort is unknown, and neither is a run at high.

Both also carry no `reasoning_effort_basis`, so there is nothing recording where the value came from — which is itself the tell, since every honestly-derived effort now carries one.

Their route is `claude-opus-5[1m]`, which offers no effort ladder. Under the current rules that means the field should be **absent**, and the gap named. The other 54 records on that route have exactly that shape.

## Why it is worth fixing rather than leaving

`default` is not a null. Anything grouping runs by effort reads these two as a third condition alongside `high` and absent — and #400, which wants to compare the agentic and API arms on reasoning spend, is precisely such a consumer.

The fix is a two-record edit, but it should be deliberate and should say what it is doing: removing a value that was never observed, not correcting one that was wrong.

Related: #448 (the observed-effort backfill), #397 (why the recorder owns effort at all).

realmarcin:
**Broader than filed.** A test written for #448 checked every recorded effort and found **17 records carrying an effort with no `reasoning_effort_basis`**, not 2:

```
  49  ('google/claude-opus-5-high', 'high', 'read from the model route; ...')
  15  ('claude-opus-5',             'high', 'NO BASIS')
   2  ('claude-opus-5[1m]',      'default', 'NO BASIS')
```

Two distinct defects, and they need different fixes:

**The 15** are the 2026-08-07 canonical sweep (5 projects × 3 replicates). Their value is almost certainly *right* — the header says `# Reasoning effort: high`, the runtime exposes `CLAUDE_EFFORT`, and #449 established that as genuinely observed rather than asserted. What is missing is only the basis, which postdates them. The fix is to record where the value came from, not to change it.

[...]

Issue #449, opened 2026-08-11
Claude Code effort is recorded as 'asserted, not observed' although the runtime does expose CLAUDE_EFFORT


realmarcin:
Review finding on #445.

`build_record` gives every Claude Code run's reasoning effort this basis:

```python
model.setdefault("reasoning_effort_basis",
                 "asserted by the generating agent, not observed")
```

with the comment *"written because a human told the agent to write it, not read from anything the runtime exposes."*

The second half is not correct. The Claude Code runtime **does** expose reasoning effort, as the environment variable `CLAUDE_EFFORT`. Verified directly: it reads `high` in this runtime, and the 2026-08-07 agentic sweep obtained its header value by running `echo "$CLAUDE_EFFORT"` and using the result verbatim — it was told *where to read the value*, not *what value to write*.

That matters because the basis string is the whole point of the field. It exists to separate a measured parameter from a restated one, and here it demotes a genuinely observed value to an assertion. Temperature sits beside it and is a true assertion — the runtime exposes no temperature knob — so lumping the two together loses the distinction the field was added to make. 30 records across the three 2026-08-07 labels are affected.

**Care needed in the fix.** Reading `os.environ["CLAUDE_EFFORT"]` unconditionally in the recorder would be wrong: `d4d provenance record` can be run from a *different* session than the one that generated the record, and would then assert the recording session's effort as the generating run's. The safe form is corroboration rather than substitution — when the header already carries an effort, the runtime is Claude Code, and `CLAUDE_EFFORT` agrees with it, upgrade the basis to observed; when they disagree, say so rather than silently preferring either.

Ranked below #402 in severity: this mislabels the strength of a provenance claim, it does not corrupt a value.

realmarcin:
Fixed in #445, merged as `4fc4d6b2`. All four CI checks green (3.10/3.11/3.12 + large-files).

Issue #444, opened 2026-08-11
One unparseable record aborts the whole scope sweep


realmarcin:
From the #440 review. Robustness, no measured failure.

`d4d download scope --check` sweeps every `*_d4d.yaml` and calls `check_record`, which does `yaml.safe_load(path.read_text())` with no guard. One unparseable record aborts the whole sweep with a traceback instead of reporting the other 170 and naming the bad file. The corpus parses today, so this is latent.

Also, in the same area as #439: `scope_of` re-reads and re-parses the ~40 KB manifest on every call, so a 171-record sweep parses it several hundred times. Same fix shape, same caveat about invalidating a cache the tests write to.

realmarcin:
Fixed in #440.

realmarcin:
Closed by #440.

PR #471, opened 2026-08-11
Record the reasoning effort 49 records' own routes already named (#448)

### LINKED ISSUE #448: 49 records could carry an observed reasoning effort recoverable from their model route
From the #445 review.

#445 derives reasoning effort from the model route, because the provider expresses effort as a name suffix. Across the corpus:

```
81  claude-opus-5[1m]           no ladder — correctly yields nothing
49  google/claude-opus-5-high   effort recoverable: high
25  claude-opus-5               no ladder
 7  None
 3  google/claude-opus-5        no ladder
```

**49 records could carry an observed reasoning effort and do not.** The information was in the route all along; nothing read it. They are not wrong today — the field is absent, which is honest — but they are less placeable than they need to be, and #400 (comparing agentic against API runs on reasoning spend) wants exactly this field.

The derivation only runs when a record is built, so nothing changes retroactively on its own. Populating them means a re-record pass over the 49, which is safe now that re-recording carries the validation verdict forward (#396) — but it rewrites 49 files, so it should be a deliberate, reviewable commit rather than a side effect of some other command.

realmarcin:
Closes #448.

`_effort_from_route` runs only when a record is *built*, so records written before it existed name `google/claude-opus-5-high` and carry no `reasoning_effort`. The information was in the route all along; nothing read it.

Adds `d4d provenance backfill-effort` — reports by default, writes only under `--execute`, because rewriting 49 records should be a deliberate reviewable commit rather than a side effect of some other command.

```
162 record(s) examined, 49 would update:
   google/claude-opus-5-high -> high            49
```

## Why a bulk pass is defensible here

The value is **observed, not asserted**, so it does not enter `unverified`. The route is already in the record and the effort is read off it — this adds no claim the record was not already making, it makes an existing one legible. The same pass would not be defensible for temperature, which no record carries evidence for.

## Canaried before the batch

One record first, per the standing rule. Diff inspected: **+3 lines**, no reformatting, comment preamble intact, validation block carried forward, `unverified` still naming only `model.temperature`, and `d4d runs check --strict` still exit 0. Then the remaining 48.

The full 49-file diff is **147 insertions, 0 deletions**, containing only:

```
+  reasoning_effort: high
+  reasoning_effort_basis: read from the model route; the provider exposes effort as
+    a model-name suffix, not a parameter
```

No record was reformatted on the way past.

## What is deliberately left alone

Routes with no ladder — `claude-opus-5`, `claude-opus-5[1m]`, `google/claude-opus-5` — stay absent. A run that was never offered an effort is a different claim from one at high, and filling it would invent exactly the distinction CLAUDE.md forbids.

## What the tests found

Fourteen tests. One pins that no recoverable effort is left unrecorded, so #448's state cannot return.

Another asserts that **every recorded effort says where it came from**, and that turned up more than was filed. Its first premise — "the route must name it" — was wrong: the agentic path legitimately records `high` on a ladderless route because the runtime exposes `CLAUDE_EFFORT` (#449). The real invariant is the *basis*, and 17 records lack one:

```
  49  ('google/claude-opus-5-high', 'high', 'read from the model route; ...')
  15  ('claude-opus-5',             'high', 'NO BASIS')
   2  ('claude-opus-5[1m]',      'default', 'NO BASIS')
```

Two different defects: the 15 have the right value and no record of its provenance; the 2 carry a value CLAUDE.md forbids. Pinned at ≤17 so the debt can only shrink, and reported on #470.

Full suite: **1588 passed, 4 skipped, 0 failed.**

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -433,7 +433,10 @@ prompt header (#397). No generic prompt names it, so records made before this
 simply lacked it — 12 of the 2026-08 API runs. Three sources, in order:
 
 1. the model route, where the provider expresses effort as a name suffix
-   (`google/claude-opus-5-high`) — recorded as observed;
+   (`google/claude-opus-5-high`) — recorded as observed. `d4d provenance
+   backfill-effort` applies this retroactively; it reports by default and
+   writes only under `--execute`, and has already populated the 49 records
+   whose route named an effort they did not carry (#448);
 2. `d4d provenance record --reasoning-effort <x>` — recorded as asserted by the
    launcher, and still listed under `unverified`;
 3. nothing — the field is left absent and the gap is named.
--- a/data/d4d_concatenated/claudecode_agent_core/2026-07-29_claude-opus-5-api-generic_rep1/CHORUS_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-07-29_claude-opus-5-api-generic_rep1/CHORUS_provenance.yaml
@@ -25,6 +25,9 @@ model:
     report: 12000
   temperature: null
   temperature_basis: not applicable to this model
+  reasoning_effort: high
+  reasoning_effort_basis: read from the model route; the provider exposes effort as
+    a model-name suffix, not a parameter
 prompts:
   hash_algorithm: sha256
   files:
--- a/data/d4d_concatenated/claudecode_agent_core/2026-07-31_canary_rep1/CHORUS_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-07-31_canary_rep1/CHORUS_provenance.yaml
@@ -25,6 +25,9 @@ model:
     report: 12000
   temperature: null
   temperature_basis: not applicable to this model
+  reasoning_effort: high
+  reasoning_effort_basis: read from the model route; the provider exposes effort as
+    a model-name suffix, not a parameter
 prompts:
   hash_algorithm: sha256
   files:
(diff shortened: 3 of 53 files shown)


realmarcin:
## Review rounds

**Round 1 — #472, the #444 failure reintroduced.** `observed_effort_gap` guarded `yaml.YAMLError`, but `read_text` sits inside the same `try` and raises `UnicodeDecodeError` and `OSError`, which that guard does not cover:

```
undecodable record -> RAISED: UnicodeDecodeError 'utf-8' codec can't decode byte 0xff
```

The function runs once per record across 162 records, so one unreadable file aborts the whole pass — exactly what #444 fixed for the scope sweep, three days earlier. Worse under `--execute`, which writes as it goes and would stop partway with no report of where.

The existing test gave false confidence: `test_unparseable_yaml_is_skipped_not_raised` used *malformed YAML*, which reaches the parser. Undecodable bytes never do. Different code path, only one exercised. Both now covered, plus `OSError` via a directory-where-a-file-is-expected.

**Round 2 — atomic write.** The write is now temp-file-plus-rename, deliberately unlike `Record.write`. That method creates a record for a run that just happened, so a failed write loses something not yet relied on. This mutates 49 records that already exist and are already attested, where a truncated write *destroys* provenance rather than failing to add it.

**Idempotence verified against the real corpus**: re-running `--execute` reports nothing to do and leaves the working tree clean, so the pass cannot churn records on a second invocation.

Full suite: **1591 passed, 4 skipped.** All four CI checks green.

PR #473, opened 2026-08-11
Match core distributions against nested resources, not just collections (#401)

### LINKED ISSUE #401: d4d_pair_consistency matches distributions only at collection level, reporting spurious unmatched core distributions
Found verifying `VOICE_PEDIATRIC` in label `2026-08-07_claude-opus-5-claudecode-generic-v3_rep1`.

The `file_collections` ↔ `distributions` check matches core distributions against **top-level** `FileCollection` entries only. It does not descend into a collection's nested `resources` (`File[]`). But `FileCollection` is collection-level (`total_bytes`, `file_count`, `collection_type`, nested `resources`) while `CoreDistribution` is file-level (`bytes`, `hash`, `md5`, `sha256`, `path`, `media_type`), so a core record that enumerates distributions per file is doing the correct thing and still gets reported as broken:

```
WARNING [semantic-review-required] $.file_collections <-> $.distributions:
  deterministic matches=1, unmatched core distributions=[0..11]
```

Verified by hand that all 12 "unmatched" distributions do match, by `id`, a nested `File` under `file_collections[*].resources`, with `name`, `path` and `bytes`/`total_bytes` agreeing and zero disagreements. Nothing was wrong with the record.

Why this matters beyond noise:

- It costs a manual justification on every affected run. The VOICE_PEDIATRIC reconciliation report had to spend a section explaining that the validator's own count is misleading.
- It is the wrong way round for safety. A genuinely unmatched core distribution — one asserting a file that exists nowhere in full — produces the *same* warning as this benign case, so the signal that should be alarming is routine. Compare AI_READI and VOICE in the same label, which report `unmatched=[]` because their core enumerates at collection level.

Suggested fix: match core distributions against the union of top-level collections and their nested `resources`, keyed by `id`, and report which level each match came from. Then `unmatched` means what it says, and the semantic-review warning can be reserved for content that actually needs a human read.

realmarcin:
Closes #401.

`FileCollection` is collection-level — `total_bytes`, `file_count`, `collection_type`, nested `resources` — while `CoreDistribution` is file-level — `bytes`, `hash`, `md5`, `sha256`, `path`, `media_type`. A core record that enumerates one distribution per *file* is therefore doing the correct thing, and matching only against top-level collections reported it as broken.

On VOICE_PEDIATRIC in the canonical label, the record #401 was filed on and hand-verified:

```
before: deterministic matches=1, unmatched core distributions=[0..11]

after:  deterministic matches=13 (1 at collection level,
        12 at nested resource level), unmatched core distributions=[]
        0 errors
```

All 12 match a nested `File` by id, with `name`, `path` and `bytes` agreeing and zero conflicts — which is what the issue verified by hand.

## The size field follows the match level

A collection carries `total_bytes`; a nested `File` carries `bytes`. Comparing a file-level size against a collection total is a category error, and would have reported a conflict on the very record that agrees exactly. Tested both directions.

## Why this matters beyond the noise

The warning now names which level each match came from, so **`unmatched` means what it says**. Previously a genuinely unmatched core distribution — one asserting a file that exists nowhere in full — produced an identical warning to the benign case, so the signal that should alarm was routine.

## Corpus swept for regressions

158 pairs. **3** now match at nested level (the three VOICE_PEDIATRIC replicates); **29** still report unmatched. Sampled two of the 29: both have *zero* nested files, and their distribution ids sit in a different namespace from their collection ids (`distribution:adult-v3-1-0-parquet-features` against `filecollection:adult-v3-1-0-features`). They legitimately have no counterpart at either level — real cases for semantic review, which is now what the warning means.

## Tests

Nine, including that a real conflict against a nested file is still caught (descending must not turn the check off), that an id duplicated across two collections is *reported* rather than settled by iteration order, and that a genuinely absent distribution still reports unmatched.

Full suite: **1599 passed, 4 skipped, 0 failed.**

### DIFF
--- a/src/data_sheets_schema/d4d_pair_consistency.py
+++ b/src/data_sheets_schema/d4d_pair_consistency.py
@@ -419,7 +419,15 @@ def _append_distribution_relation_issues(
             distribution, mapping_collections
         )
         level = "collection"
-        if collection is None:
+        if collection is None and match_basis is None:
+            # Only on a *clean* miss. `_related_match` returns
+            # (None, "ambiguous <key>") when two collections share an
+            # identifier, which is a data defect; descending past it and
+            # finding one nested file would report a tidy resource-level match
+            # and delete the ambiguity from the output (#474). That would make
+            # the defect less visible than before this change, inverting what
+            # #401 is for.
+            #
             # Descend into each collection's nested `resources` (#401).
             # `FileCollection` is collection-level — total_bytes, file_count,
             # collection_type — while `CoreDistribution` is file-level — bytes,
--- a/src/data_sheets_schema/d4d_pair_consistency.py
+++ b/src/data_sheets_schema/d4d_pair_consistency.py
@@ -330,6 +330,26 @@ def _append_resource_projection_errors(
             )
 
 
+def _nested_resources(
+    collections: Sequence[Mapping[str, Any]],
+) -> list[Mapping[str, Any]]:
+    """Every `File` nested under a collection's `resources` (#401).
+
+    Flattened across collections rather than searched per collection, because
+    a core distribution names the file, not the collection it sits in — and
+    `_related_match` already refuses an ambiguous match, so two files sharing
+    an id anywhere in the record are reported rather than silently resolved to
+    whichever came first.
+    """
+    out: list[Mapping[str, Any]] = []
+    for collection in collections:
+        resources = collection.get("resources")
+        if not isinstance(resources, list):
+            continue
+        out.extend(item for item in resources if isinstance(item, Mapping))
+    return out
+
+
 def _related_match(
     distribution: Mapping[str, Any],
     collections: Sequence[Mapping[str, Any]],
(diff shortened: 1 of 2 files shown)


realmarcin:
## Review round — #474, an inversion of the PR's own purpose

`_related_match` distinguishes two failures and returns them differently:

```python
if len(matches) > 1:
    return None, f"ambiguous {key}"     # a data defect
return None, None                       # simply not here
```

The descent as first written treated them alike. So a distribution whose id matched **two top-level collections** — two collections sharing an identifier — would fall through to the nested search and, finding one nested file, report a tidy `1 at nested resource level` match. The collection-level ambiguity would vanish from the output entirely.

That is worse than before this PR, where the case at least reported `unmatched`. It inverts the intent: #401 is about making `unmatched` mean what it says, and this would have made a real defect *less* visible than it was.

Descent is now conditional on `match_basis is None`. The nested-level ambiguity case was already correct and tested; this is the same rule one level up, and it was missing.

VOICE_PEDIATRIC unaffected — still 13 matches, 1 at collection level and 12 nested, `unmatched=[]`.

Issue #463, opened 2026-08-11
form_defects._value_index hardcodes four projects, so every VOICE_PEDIATRIC failure would attribute as 'unattributed'


realmarcin:
Round-5 review finding on #459. Blocks the five-project measurement the PR registers.

`_value_index` (form_defects.py:150) hardcodes the project list:

```python
for cfg, label in configs.items():
    tag = cfg.split()[0]
    for project in ("AI_READI", "CHORUS", "CM4AI", "VOICE"):
        for record in load_replicates(root, method, label, project).values():
```

`VOICE_PEDIATRIC` is absent. It has been a project since #298 and is in `PROJECTS`:

```
['AI_READI', 'CHORUS', 'CM4AI', 'VOICE', 'VOICE_PEDIATRIC']
```

## Why it matters, and why it is quiet

`attribute()` recovers a failure's arm by matching the cached value back against the records — exact match on canonical JSON, deliberately, because *"an approximate attribution would silently mix the two arms being compared, which is the one error this whole exercise exists to undo."*

A value that appears in no indexed record gets `tags = set()` and therefore `config = ""`, which `table()` buckets as `"unattributed"`. So every form failure originating in a VOICE_PEDIATRIC record lands in a column that reads as a data problem rather than as a project the index never looked at.

It is silent today only because no VOICE_PEDIATRIC record has been fitness-scored — the 106 cached failures predate the split. `notes/generic_v1_v3_analysis_plan.md` schedules exactly that: five projects × three replicates × two arms.

There is an existing guard that would catch it, but only after the fact — `test_the_form_failures_are_the_two_arms_exactly` asserts `{"v1": 50, "v2": 56}` with nothing unattributed. On the new corpus that test's premise no longer holds, so it will be rewritten for the new arms, and if it is rewritten to match observed output rather than expected coverage the omission passes through.

## Fix

Take the project list from `PROJECTS` rather than a literal, so a sixth dataset needs no edit here — the property `notes/PIPELINE_GENERICITY_AUDIT.md` (#418) checked for and found the *code* already had. This is one of the few places it does not.

Worth also asserting that the index covers every project in `PROJECTS`, so the next addition fails a test rather than producing an "unattributed" column.

PR #477, opened 2026-08-11
Stop the byte-identity guard shadowing PROJECTS, and state why agreement's default is four (#467)

### LINKED ISSUE #467: Two four-project literals are logic, not prose: agreement.DEFAULT_PROJECTS, and the byte-identity test that claims 'every project'
Found by an adversarial Codex review of #459, checking NEXT_TASKS §9's claim that *"adding the fifth project broke nothing; the four-project list in 26 files is prose, not logic."*

Mostly true. Three exceptions, one of which #459 fixes.

## 1. `form_defects._value_index` — fixed in #459 (#463)

Was a runtime literal. Any VOICE_PEDIATRIC failure attributed as `unattributed`, indistinguishable from a value that matched nothing. Now reads `PROJECTS`.

## 2. `agreement.DEFAULT_PROJECTS` — open

```
src/data_sheets_schema/agreement.py:551
DEFAULT_PROJECTS = ("AI_READI", "CHORUS", "CM4AI", "VOICE")
```

Used as the default for `build_matrix(projects=...)` (:574) and by the CLI when `--projects` is not passed (:667). So a replicate-agreement matrix built with default arguments silently omits VOICE_PEDIATRIC. Unlike #463 this one is at least visible in the output — the project simply has no row — but nothing states that the absence is a default rather than a result.

Worth deciding rather than mechanically changing: the agreement figures published in §7 were computed over four projects, and `SHARED_CORPUS_GROUPS` exists because VOICE and VOICE_PEDIATRIC are not independent samples. Widening the default changes what a bare invocation reports against what §7 published. The alternative — read `PROJECTS` and let the shared-corpus grouping handle non-independence — is probably right, but it is a decision about the published figures, not a lint fix.

## 3. `tests/test_download/test_generic_v2_prompt.py:17` — open

```
PROJECTS = ("AI_READI", "CHORUS", "CM4AI", "VOICE")
```

This shadows the real registry inside the test that asserts **every project receives byte-identical prompt text after mechanical substitution**. NEXT_TASKS §6 cites that assertion as a guard on the generic condition. It does not hold for VOICE_PEDIATRIC, because the test never asks.

That is the one that matters most of the three: it is a guard whose stated scope ("every project") is wider than its actual scope, in exactly the area — per-project prompt divergence — that #419/#422 were about. A test importing `PROJECTS` would have failed when the fifth project arrived, which is the behaviour wanted.

## Suggested

- 3 first, and mechanically: import `PROJECTS` rather than shadowing it.
- 2 as a decision, noting the effect on §7's published figures.
- Correct §9's "prose, not logic" claim. Done in #459.

realmarcin:
Closes #467. Two four-project literals that NEXT_TASKS §9 counted as "prose, not logic". They are not.

## 1. The byte-identity guard was narrower than it claimed

`tests/test_download/test_generic_v2_prompt.py` declared its own `PROJECTS` tuple, shadowing the registry. So the assertion its own module docstring opens with —

> The arm's whole value is that **every project** receives identical text.

— excluded VOICE_PEDIATRIC from the day #298 made it a project. A guard whose stated scope is wider than its actual scope, in precisely the area #419 and #422 were about: per-project prompt divergence.

Now imports `PROJECTS`, so a sixth dataset fails this file rather than passing it by omission.

**The guarantee did hold.** All 18 tests pass unchanged with the fifth project included, and `VOICE_PEDIATRIC_preprocessed.txt` is present so the case is genuinely exercised rather than silently skipped. It simply was not being asserted — which is the difference between a property and a checked property.

## 2. `agreement.DEFAULT_PROJECTS` stays four, and now says why

Not a mechanical widening, as the issue argued. This is the project set §7's published agreement figures were computed over — 261/540 shared slots, 48.3%, between-config effect −2.9 against a per-project delta sd of 10.9. Changing the default would silently alter what a bare invocation reports against what was published: a decision about the figures, not a lint fix.

VOICE_PEDIATRIC is reachable with `--projects`, and is excluded from the *default* because it and VOICE share a source corpus — counting them as two samples overstates the independent n that §7's power argument rests on (`SHARED_CORPUS_GROUPS`).

## The two omissions are not the same kind

`form_defects._value_index` (#463, fixed) silently produced an `unattributed` column that read as a data problem. An absent project here simply has no row, which is visible. Recorded in the comment so the next reader does not "fix" it.

Full suite: **1600 passed, 4 skipped, 0 failed.**

### DIFF
--- a/src/data_sheets_schema/agreement.py
+++ b/src/data_sheets_schema/agreement.py
@@ -548,6 +548,21 @@ def compare_records(records: dict[str, dict[str, Any]], *,
 DEFAULT_METHOD = "claudecode_agent"
 DEFAULT_CONFIGS = {"v1  (2026-07-28 generic)": "2026-07-28_claude-opus-5-generic",
                    "v2  (2026-07-31 generic-v2)": "2026-07-31_claude-opus-5-generic-v2"}
+#: Four, not `PROJECTS`, and deliberately so (#467). This is the project set
+#: §7's published agreement figures were computed over — 261/540 shared slots,
+#: 48.3%, and the between-config effect of −2.9 against a per-project delta sd
+#: of 10.9. Widening the default would silently change what a bare invocation
+#: reports against what was published, which is a decision about the figures
+#: rather than a lint fix.
+#:
+#: VOICE_PEDIATRIC is reachable with `--projects`; it is excluded from the
+#: default because it and VOICE share a source corpus, so counting them as two
+#: samples overstates the independent n that §7's power argument rests on. See
+#: `SHARED_CORPUS_GROUPS`.
+#:
+#: Unlike `form_defects._value_index`, whose omission of the fifth project was
+#: an oversight that silently produced an `unattributed` column, an absent
+#: project here is visible: it simply has no row.
 DEFAULT_PROJECTS = ("AI_READI", "CHORUS", "CM4AI", "VOICE")
 DEFAULT_ROOT = Path("data/d4d_concatenated")
 DEFAULT_CACHE = Path("data/evaluation_llm/agreement_cache")
--- a/tests/test_download/test_generic_v2_prompt.py
+++ b/tests/test_download/test_generic_v2_prompt.py
@@ -13,8 +13,14 @@ from pathlib import Path
 from data_sheets_schema.api_runner import (
     GENERIC_PROMPT, GENERIC_PROMPT_V2, RunSpec, prompt_body, resolve_prompt,
 )
+from data_sheets_schema.constants import PROJECTS
 
-PROJECTS = ("AI_READI", "CHORUS", "CM4AI", "VOICE")
+# PROJECTS is imported, not restated. This file previously shadowed it with a
+# four-project literal, so the assertion "every project receives identical
+# text" excluded VOICE_PEDIATRIC from the day it became a project (#298) — a
+# guard whose stated scope was wider than its actual scope, in exactly the area
+# #419 and #422 were about. A fifth project now fails this file rather than
+# passing it by omission (#467).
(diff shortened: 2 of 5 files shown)


realmarcin:
## Review round — the v2 guard was not the only one, and widening found a real gap

Asking whether other files shadowed `PROJECTS` turned up three more with the identical literal: `test_generic_v3_prompt.py`, `test_generic_v4_prompt.py`, and `test_prompt_components.py`. **v3 is the condition actually up for promotion in §6**, so its guard being narrow mattered most.

Widening the components guard failed immediately:

```
FileNotFoundError: src/download/prompts/components/VOICE_PEDIATRIC.md
```

```
$ ls src/download/prompts/components/
AI_READI.md  CHORUS.md  CM4AI.md  README.md  VOICE.md
```

VOICE_PEDIATRIC has had **no component file since the fork**, so the `tuned` condition cannot be run for it, and four tests asserting properties over "every project" were covering four of five. This is exactly the shape #467 is about: the guard that would have caught it declared its own project list, so the only thing that looks was the thing that was narrowed.

It has not bitten in practice — every run since the fork used a `generic*` condition, and only `tuned` consumes components. Filed as **#478** rather than fixed here: authoring a component is content and a judgement call (it must state what distinguishes the pediatric cohort without importing adult-cohort facts, the `target` failure the fitness axis names) and it interacts with the manifest scope declaration (#422).

Pinned meanwhile as a **set equality**, not a skip — so filling the gap fails the test until `WITHOUT_COMPONENT` is updated, and a *new* gap fails it too. The three content assertions iterate the projects that have a file; the existence assertion iterates all five.

v3 and v4 pass unchanged at five projects, as v2 did.

Issue #479, opened 2026-08-11
The agentic playbook already carries v2/v3's rules, so the registered v1-vs-v3 sweep cannot answer its own question on that path


realmarcin:
Found while checking whether #385 and #380 gate the sweep registered in `notes/generic_v1_v3_analysis_plan.md`. They do not — but this does, and it is larger.

## The plan cannot answer its own question on the agentic path

The plan compares **v1 (`generic`) against v3 (`generic_v3`)**, five projects × three replicates, on the Claude Code agentic path. The rules v3 adds to the *prompt* are already present in the *playbook* both arms read.

`.claude/commands/d4d-agent.md:133-145`, added by #394 ("playbook parity") on 2026-08-07:

> **Slot-filling order** (same contract as the API pipeline's phase instructions): a class's structured slots first — `name`, `id`, `affiliations`, `grants` and kin **must not sit empty while their content sits in prose** — then `description` as the default home for narrative, then `notes` only for content `description` cannot hold. […] Never restate a sibling slot's value, and never invent a key.
>
> **Shape check before writing**: a value must match its slot's range — no prose where the schema requires a list […]

That is the substance of v2's rule 1, v3's rule 4, and #385's fix. Both arms of the planned comparison would read it, because the agentic launch instruction's first lines tell the agent to open the playbook.

## The empirical demonstration is already on disk

The 2026-08-07 sweep ran the **generic v1** prompt (#454) on the agentic path, after #394 landed the same day. Measured over every record, grouped by runtime — the two paths happen to have exactly the same creator count, which makes it unusually clean:

| runtime | creators | with `name` | with `affiliations` | using `notes` |
|---|---|---|---|---|
| Claude Code (agentic) | 692 | 692 (100%) | 674 (97%) | 0 |
| Claude API (direct) | 692 | 274 (39%) | 69 (9%) | 76 |

**A v1 prompt on the agentic path already produces v3-quality structure.** The affiliations are well-formed `Organization` objects carrying ROR ids — `{name: 'Washington University in St. Louis', id: 'https://ror.org/01yc7t268'}` — which is exactly what #378's enrichment wants, and the records validate.

So on the agentic path the prompt version is close to inoperative for the defects v2/v3 target. A v1-vs-v3 comparison there would measure ≈0 and the correct reading would be "both arms had the rules", not "the rules do nothing".

## Why this is not just a power problem

[...]

Issue #380, opened 2026-08-06
Undeclared-key inventions are mostly standard-vocabulary near-misses; notes escape hatch added for the rest


realmarcin:
Prompted by Harry Caufield's question about the structure hallucinations: *"Is there a general-purpose notes field the agent can store things in instead of inventing new structures?"* — hypothesis being the model finds content the instructions want but the schema has no place for.

**Analysis of all 79 undeclared-key findings in the corpus inventory:** the hypothesis holds for a minority. The dominant pattern is **standard-vocabulary near-misses** — the model reaches for the canonical external name where our schema has a different one:

| Invented key | Occurrences | The near-miss |
|---|---|---|
| `affiliation` (Creator) | 47× | Schema slot is `affiliations`; `Person` has singular `affiliation` |
| `download_url`, `checksum` (distributions) | 11× | DCAT's `downloadURL`, SPDX checksum |
| `format`, `md5`, `media_type` (subsets) | 6× | dcterms/DCAT standard file metadata |
| `grant` (funders) | 7× | Schema slot is `grants` |
| `same_as` (root) | 1× | `schema:sameAs` |
| `data_type`, `representation`, `dialect/description` | 7× | Genuinely unplaced — Harry's case |

This is the slot-level twin of the `IsNewVersionOf` enum-casing finding: the model knows the standard vocabularies and uses their canonical names.

**Done now (pre-rerun):**
1. `notes` slot (`schema:comment`, optional string) on `DatasetProperty`, `NamedThing` (→ Dataset, Person, File, Software…), `Organization`, `Grant` — the escape hatch that makes inventing keys unnecessary.
2. Full/core phase instructions now state: only declared keys; unplaced content goes in `notes`.

**Open for consideration:** slot-level aliases for the near-miss names (mirroring the enum-alias + write-path normalisation pattern) or adding the DCAT-standard slots outright (`download_url` etc.) where models repeatedly reach for them — that's schema modeling with interop upside, worth its own decision.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

realmarcin:
**Re-checked against the current corpus.** Of the invented keys catalogued here, all but two are now *declared slots*, so their appearance is correct behaviour rather than invention:

```
affiliation      declared        download_url    declared
affiliations     declared        checksum        declared
format           declared        md5             declared
media_type       declared        data_type       declared
dialect          declared        same_as         declared
grants           declared        notes           declared
                                 source_caveats  declared

[...]

PR #480, opened 2026-08-11
Clear #385/#380 as pre-sweep gates, and move the comparison off the agentic path (#479)

realmarcin:
Amends `notes/generic_v1_v3_analysis_plan.md` **before any run**, and records the corpus-wide re-measurement of #385 and #380.

## #385 and #380 do not gate the sweep

Re-measured across the corpus rather than one label, as #385's own last comment asked.

**Mode 1 (prose defaults to `notes`) — fixed.** The canonical agentic series is **971 `description` to 7 `notes`**, with **276 `source_caveats`** carrying evidence commentary in the channel mode 3 asked for. The 4:1 misuse this was filed on is 1:311 in the current arm.

**Mode 2 (structured slots bypassed) — fixed on the agentic path, and only there.** Both runtimes happen to have exactly 692 creators, which makes the cut unusually clean:

| runtime | creators | with `name` | with `affiliations` | using `notes` |
|---|---|---|---|---|
| Claude Code (agentic) | 692 | 692 (100%) | 674 (97%) | 0 |
| Claude API (direct) | 692 | 274 (39%) | 69 (9%) | 76 |

The agentic affiliations are well-formed `Organization` objects with ROR ids — `{name: 'Washington University in St. Louis', id: 'https://ror.org/01yc7t268'}` — which is exactly what #378's enrichment wants. Records validate.

**#380 — fixed.** Every invented key it catalogued except `grant` and `representation` is now a declared slot (#388, #381), and neither of those two appears in the canonical sweep at all.

## What does gate it — #479

The rules v3 adds to the **prompt** are already in the **playbook** the agentic path reads (`.claude/commands/d4d-agent.md:133-145`, added by #394 on 2026-08-07):

> a class's structured slots first — `name`, `id`, `affiliations`, `grants` and kin must not sit empty while their content sits in prose […] Never restate a sibling slot's value, and never invent a key.

Both arms open the playbook, because the launch instruction's first lines tell them to. So a v1-vs-v3 comparison there would measure ≈0 — and the honest reading would be *"both arms had the rules"*, not *"the rules do nothing"*.

**The demonstration is already on disk.** The 2026-08-07 sweep ran the **v1** prompt agentically, after #394 landed the same day, and produced v3-quality structure. That is the 100% / 97% row above.

## The amendment

The promotion decision moves to the **API path** — where the prompt is the only channel carrying these rules, where the defect is live, and where v1/v2/v3 were originally measured, so the result is comparable with `notes/generic_v2_results.md` rather than being a new baseline.

**The agentic v1 arm is still worth running, for a different question.** The canonical set needs replacing regardless (#454), and a fresh arm also clears the bundle drift on all five canonical records (#452). It answers replicate variance and corpus currency, not prompt promotion — and the write-up must not conflate them. Two runs of 15, not one run of 30.

Amended before the runs rather than discovered after them, which is the whole point of registering a plan.

Full suite: **1600 passed, 4 skipped, 0 failed.**

### DIFF
--- a/NEXT_TASKS.md
+++ b/NEXT_TASKS.md
@@ -257,12 +257,41 @@ guard could not do is announce that the arm it protected had never been
 generated — an absent cache entry and an absent run look identical.
 
 **Decided, 2026-08-11: generate both arms fresh.** v1 and v3, five projects ×
-three replicates = 30 runs, at today's digest `e802cdc3`, on the agentic path,
-with instructions rendered rather than typed. Registered in
+three replicates = 30 runs, at today's digest `e802cdc3`, with instructions
+rendered rather than typed. Registered in
 `notes/generic_v1_v3_analysis_plan.md` before any run. The pairing is v1-vs-v3
 rather than v2-vs-v3 because v1 is the playbook default, so that is the
 comparison the promotion decision rests on; the mechanism question (does rule 4
 alone do what it says?) stays open and needs a v2 arm at the same digest.
+
+**Amended the same day, before any run: on the API path, not the agentic one
+(#479).** The rules v3 adds to the prompt are already in the playbook the
+agentic path reads — `.claude/commands/d4d-agent.md:133-145`, added by #394 —
+so both arms would receive them regardless of `--condition` and the comparison
+would measure ≈0 for the wrong reason. The demonstration is on disk: the
+2026-08-07 sweep ran **v1** agentically and produced v3-quality structure.
+Grouped by runtime, over identical creator counts:
+
+| runtime | creators | with `name` | with `affiliations` | using `notes` |
+|---|---|---|---|---|
+| Claude Code (agentic) | 692 | 692 (100%) | 674 (97%) | 0 |
+| Claude API (direct) | 692 | 274 (39%) | 69 (9%) | 76 |
+
+The API path is where the prompt is the only channel carrying these rules, where
+the defect is live, and where v1/v2/v3 were originally measured — so the result
+is comparable with `notes/generic_v2_results.md`.
+
+**The agentic v1 arm is still worth running, for a different question**: the
+canonical set needs replacing regardless (#454), and a fresh arm also clears the
+bundle drift on all five canonical records (#452). It answers replicate variance
+and corpus currency, not prompt promotion, and the two must not be conflated.
+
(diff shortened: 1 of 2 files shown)


realmarcin:
## Review rounds

**Round 1 — #481, an error in my own causal claim.** I attributed the agentic path's structured creators to #394's playbook parity. The timing checks out (#394 landed 20:42 UTC, replicates recorded 23:16–00:58 UTC) but the attribution is wrong: **four agentic series predating #394 already sit at 94–100% under v1 prompts.**

So the operative variable is the **path**, not the prompt and not the playbook — most likely the mandatory `linkml-validate`-and-iterate step, which converges on schema-valid structured output whatever the prompt says. That makes the recommendation *better* supported: the agentic comparison measures ≈0 for a reason with nothing to do with the rules.

Two things the corrected table added. The effect this comparison exists to measure **is already visible on the API path** — v1/v2 at 0%, v3 at 69–100%. And **it does not survive schema 2.0.0**, which is what the planned run uses.

**Round 2 — checking whether the schema attribution was itself confounded.** The only schema-2 v3 data is also `1m`, so the 0% might have been a route effect. It is not — the three series decompose because each step changes exactly one variable:

| series | route | schema | affiliations |
|---|---|---|---|
| `2026-08-05_…-generic-v3` | `google/claude-opus-5-high` | `b065e1cd` (1.0.0) | **100%** |
| `2026-08-05_…-1m-generic-v3` | `claude-opus-5` | `b065e1cd` (1.0.0) | **69%** |
| `2026-08-06_…-1m-generic-v3-schema2` | `claude-opus-5` | `583d79c1` (2.0.0) | **0%** |

Step one holds the schema and changes the route → **100% → 69% is a route effect**. Step two holds the route and changes the schema → **69% → 0% is a schema effect**, and that is the step this run inherits.

One series, so it carries as a strong prior rather than a measurement — but the variable is isolated, which is more than the v2-vs-v3 comparison could say for itself (#458). Registered as prediction 5, with the explicit note that a 0% result is a finding about the *schema* — the trap-slot work (#376, #382) changed what a `Creator` looks like — not about v3 failing.

Scoring is now two-way throughout: **populated** vs **well-formed**. The 2026-08-05 run is 100% populated with bare strings; the agentic runs carry `Organization` objects with ROR ids. Averaging them would hide the distinction rule 4 is about.

Issue #462, opened 2026-08-11
form_defects ignores its own recorded_model(): the live pin moved, all 106 cached labels fell out of scope, and an online run would pool two instruments


realmarcin:
Round-3 review finding on #459. Pre-existing, and it blocks the measurement plan registered in that PR.

## What happens

```
$ python -m data_sheets_schema.form_defects --offline
106 form failure(s) loaded
OfflineCacheMiss: no cached subtype for 'intended_uses'
```

The cache is complete — 106 entries, all at the current rubric digest `6bd1d952`:

```
lines: 106
('6bd1d952', 'google/claude-opus-5-high') 106
106/106 cache entries usable at the current rubric
```

But the classifier loads **none** of them:

```
resolved model: 'claude-opus-5'
memo size: 0
missing: 106 of 106
```

`_load()` filters on `entry["model"] == self.model`, and `self.model` falls back to `_model_settings()["name"]` — the **live config pin**, now `claude-opus-5`. The cache was written under `google/claude-opus-5-high`. Every entry is out of scope.

## The function that fixes this already exists and is wired to nothing

`recorded_model(cache_path)` (form_defects.py:176) reads the one model the cache records, and its docstring describes this exact failure:

> A reproduction must run on the instrument the cache records, not on the live config pin: the pin is free to move (#345 moved it), and the moment it does, every model-scoped entry falls out of scope and an offline rebuild of a frozen artifact fails (#351). Requires exactly one recorded model for the current rubric — zero means there is nothing to reproduce, two means two instruments are pooled and neither table can be trusted (#277).

It predicted the failure precisely. Neither `main()` nor `FormSubtypeClassifier.__init__` calls it.

## Why this is worse than a failed offline run

`--offline` fails loudly, which is the good case. **Without** `--offline` it succeeds: 106 paid calls, and 106 new entries appended under a second model. After that, `recorded_model()` raises on its own cache — *"two means two instruments are pooled and neither table can be trusted"* — and the v1→v2 baseline can no longer be reproduced at all.

So the first person to run the documented command, without a flag they had no reason to pass, silently converts a reproducible measurement into an unreproducible one.

## Why it blocks #459

[...]

PR #482, opened 2026-08-11
Key sub-type labels on the schema they were judged against (#465)

### LINKED ISSUE #465: The sub-type cache key is (slot, value) with no schema, so a label survives a schema change that redefines it
Found by an adversarial Codex review of #459. **Not fixed there** — the plan works around it; the fix is a cache migration and wants its own decision.

The fitness judgement cache is keyed on `(axis, model, rubric, corpus, schema)`. That schema component is load-bearing: it is why all 1441 fitness judgements are scoped to `34d24ff3` and why no v3 record was ever a silent cache hit (#458).

The **sub-type** cache is not. `FormFailure.key` is:

```python
@property
def key(self) -> str:
    return _digest(json.dumps([self.slot, self.value[:VALUE_CHARS]],
                              ensure_ascii=False))
```

Slot and value only. The written entry records `rubric`, `model`, `chars`, `key`, `slot`, `subtype`, `reason` — no schema.

## Why it bites

The classifier's question is *"does this value populate the fields its declared range declares?"* — `hollow_object` is defined against the slot's specification, and `slot_spec()` reads the live schema. So the label is a function of the schema, and the key does not record it.

Concretely: `Organization.id` became optional (#382). An object that was hollow under the old spec may not be under the new one. An identical slot/value classified before that change returns its old label from cache, with nothing indicating the specification underneath it moved.

This is exactly the failure the fitness cache's schema key was added to prevent, in the module layered on top of it.

## Why it has not caused a wrong number yet

Every cached label was written on 2026-08-03 against `34d24ff3`, and nothing has been classified since. The corpus has not yet contained two schemas' worth of values.

`notes/generic_v1_v3_analysis_plan.md` schedules the first run that would: 30 records at `e802cdc3`. Its workaround is to classify into a **separate cache file** via `--cache`, which is correct and is discipline rather than structure — the same distinction #432 drew for prompt pins.

## Fix

Add the schema digest to the key, or at minimum record it on each entry and scope `_load()` by it the way `model` and `rubric` already are. The existing 106 entries would need stamping with `34d24ff3`, which is safe: it is the only schema they could have been judged against.

realmarcin:
Closes #465.

The fitness cache is keyed on `(axis, model, rubric, corpus, schema)`. The sub-type cache was keyed on `(slot, value)` alone — though its labels are just as schema-dependent: `slot_spec()` renders the **live** schema into the classifier prompt, and the question it asks is whether a value populates the fields its declared range declares.

`Organization.id` becoming optional (#382) can turn a hollow object into a well-formed one **without changing a byte of the value**. An unscoped key returns a verdict reached against a specification that no longer exists.

## The migration is recoverable, not assumed

The 106 existing entries are stamped `34d24ff3`. That is not a guess: every fitness judgement they classify is keyed to that digest, and a failure cannot be classified against a schema its own judgement was not made under. An unstamped entry is honoured **only** at `LEGACY_SCHEMA`, so it expires when the schema moves rather than silently answering for it.

## Scoping alone would have broken reproduction

The live schema is now `e802cdc3`, so scoping without more would have put every label out of scope and broken the offline rebuild of the published v1→v2 table. The schema therefore resolves **from the cache first, live second** — mirroring how the model resolves (#462), and for the same reason: a populated cache is a frozen measurement and reproducing it must not depend on the working tree, while an empty one (what `--cache <new path>` gives a new arm) has nothing to reproduce.

Verified end to end:

```
instrument: google/claude-opus-5-high
0 call(s), 106 cached
collapsed_cardinality  34   2      folded:  42   7
hollow_object           0  45               8  50
cache pins schema: 34d24ff3
```

## Retires a workaround in the analysis plan

The plan told the new arms to classify into a separate cache file, which was discipline rather than structure. No longer needed: their entries will carry `e802cdc3`, the v1/v2 entries carry `34d24ff3`, and the two cannot be mixed.

Seven tests, including that a pre-#465 entry is honoured at the legacy schema and expires anywhere else, and that two schemas in one cache refuse via `PooledInstruments`.

Full suite: **1607 passed, 4 skipped, 0 failed.**


realmarcin:
## Review round — #483, the PR contradicting its own claim

This PR said the plan's separate-cache workaround was retired and "the new arms can share the cache". They could not:

```
# cache holding one 34d24ff3 entry and one e802cdc3 entry
RAISED PooledInstruments: … records 2 schemas for the current rubric …
```

`_default_schema` called `recorded_schema`, which requires exactly one. So the moment a second schema's labels landed, the classifier refused to construct at all unless every later invocation passed an explicit schema — **worse than the workaround it claimed to retire.**

**The refusal is right for models and wrong for schemas.** Two instruments in one file means neither table can be trusted (#277) and there is no principled pick. Two *schemas* is the expected end state: every entry is correctly scoped, `_load` filters on it, and accumulating across schema versions is what keying on schema is *for*. There is a right answer — the schema the caller is working at.

Resolution when the caller does not say:

| cache records | resolves to |
|---|---|
| one schema | that one, regardless of the working tree — the frozen case, and what keeps the published table reproducible after a slot description is edited |
| several, one of them live | the live one |
| several, none live | refuse, naming the candidates; pass `--schema` |

Also adds `--schema` to the command line, which this PR had exposed at the API level but not the CLI, and prints the resolved schema next to the instrument so a run states its scope rather than leaving it inferred:

```
instrument: google/claude-opus-5-high  schema: 34d24ff3
```

Four more tests, including that the frozen single-schema case still pins independently of the live schema — the property the accumulation must not break.

Full suite: **1611 passed, 4 skipped.** All four CI checks green.

PR #484, opened 2026-08-11
Say where every recorded reasoning effort came from (#470)

### LINKED ISSUE #470: Two records carry reasoning_effort: default, the value CLAUDE.md explicitly forbids, on a route with no effort ladder

realmarcin:
Closes #470.

Seventeen records stated an effort with no `reasoning_effort_basis`, so the value could not be placed on the ladder #397 defines — route (observed), agent (asserted), or absent (gap named). Anything grouping runs by effort read them as a third condition alongside `high` and absent.

```
record_basis      15  effort=high     route=claude-opus-5
drop_placeholder   2  effort=default  route=claude-opus-5[1m]
```

Two defects needing **opposite** fixes, kept separate throughout.

## The 15: the value is right, its provenance was missing

The 2026-08-07 sweep. Header says `high`, the runtime does expose `CLAUDE_EFFORT` (#449), and only the basis field postdates them. The basis written is the same string `build_record` writes in the same situation — *"asserted by the generating agent, not observed"* — because that is exactly what is known: the header carries the value and nothing here observed it. Claiming more would be the fabrication the recorder exists to prevent, so each also gains an `unverified` entry.

## The 2: the value names no effort, so no basis can be honest about it

`default` is forbidden outright by CLAUDE.md, on a route with no effort ladder. **Removed rather than relabelled**, with the gap named:

```yaml
unverified:
- field: model.reasoning_effort
  value: null
  reason: 'the record carried ''default'', which names no effort; the route
    ''claude-opus-5[1m]'' exposes no effort ladder, so nothing was chosen.
    Removed rather than kept: a run that did not choose an effort is a
    different claim from one at any level (#470).'
```

That is a deletion, so it is reported separately and never folded into the other count.

## Canaried per action before the batch

`drop_placeholder`: one line removed, gap named, nothing reformatted. `record_basis`: one line added to `model`, one `unverified` entry, nothing else. Then the remaining 15. The whole diff is **17 files, 102 insertions, 2 deletions**, and every touched line concerns `reasoning_effort`.

## Corpus after

```
49  ('high', 'read from the model route; the provider expo…')
15  ('high', 'asserted by the generating agent, not observ…')
```

No record carries a placeholder. `d4d runs check --strict` still exits 0.

The guard added with #448 pinned this debt at ≤17 so it could only shrink; **it now asserts zero**, and a second guard asserts no record states an effort that names none.

Full suite: **1623 passed, 4 skipped, 0 failed.**

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -441,7 +441,9 @@ simply lacked it — 12 of the 2026-08 API runs. Three sources, in order:
    launcher, and still listed under `unverified`;
 3. nothing — the field is left absent and the gap is named.
 
-Never write "default", "unspecified" or a guess. A run that did not choose an
+Never write "default", "unspecified" or a guess — `d4d provenance
+backfill-effort-basis` removes such a value and names the gap it leaves, rather
+than relabelling it (#470). A run that did not choose an
 effort is a different claim from a run whose effort is unknown, and neither is
 a run at high. The fix lives in the recorder because adding a header line to a
 generic prompt would re-baseline that condition for every project and require a
--- a/data/d4d_concatenated/claudecode_agent_core/2026-07-27_claude-opus-5_rep1/VOICE_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-07-27_claude-opus-5_rep1/VOICE_provenance.yaml
@@ -15,7 +15,6 @@ model:
   agent_runtime: Claude Code
   provider: Anthropic
   model: claude-opus-5[1m]
-  reasoning_effort: default
   mode: four-phase project agent
   temperature: '0.0'
 schema:
@@ -75,3 +74,9 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-07-27_claude-opus-5_rep1/VOICE_d4d_core.yaml
       sha256: 7d689413e4c5a095de6e996b16d6f6fbcdb41f2e736bf1d65ca599d92f97eaa2
   recorded_by: d4d runs validate
+unverified:
+- field: model.reasoning_effort
+  value: null
+  reason: 'the record carried ''default'', which names no effort; the route ''claude-opus-5[1m]''
+    exposes no effort ladder, so nothing was chosen. Removed rather than kept: a run
+    that did not choose an effort is a different claim from one at any level (#470).'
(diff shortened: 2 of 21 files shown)


realmarcin:
## Review round

**No duplicate `unverified` entries created** — checked across all 162 records: `0` with a repeated field. Some records already carried a `model.temperature` entry, and those survive alongside the new one.

**The existing counter now surfaces the honest state**, which is #447's counter doing exactly its job:

```
ⓘ  94 value(s) across these runs were recorded but not observed:
     model.temperature            77
     model.reasoning_effort       17
```

17 = the 15 asserted values plus the 2 dropped-gap entries. Before this PR the same 17 records stated an effort and reported nothing — the gap existed and was invisible.

**Nothing downstream regressed.** `d4d runs check --strict` exits 0, `d4d runs canonical` still marks all five projects, and **15/15** canonical-label records still carry a passing validation verdict — the property #396 exists to protect, since this pass rewrites records that are already attested.

**The deletion is recorded, not silent.** The two dropped values are named in the `unverified` reason (`the record carried 'default'…`), so what was removed is readable from the record itself, not only from git history.

Full suite: **1623 passed, 4 skipped.** All CI checks green.

PR #485, opened 2026-08-11
Author and pin the VOICE_PEDIATRIC prompt component (#478)

### LINKED ISSUE #478: VOICE_PEDIATRIC has no prompt component, so the tuned condition cannot be run for it
Found by widening `tests/test_download/test_prompt_components.py` to read `PROJECTS` instead of shadowing it with a four-project literal (#467). The test named `test_every_project_has_a_component_file` immediately failed:

```
FileNotFoundError: src/download/prompts/components/VOICE_PEDIATRIC.md
```

```
$ ls src/download/prompts/components/
AI_READI.md  CHORUS.md  CM4AI.md  README.md  VOICE.md
```

VOICE_PEDIATRIC has been a project since #298 and has no component file, so **the `tuned` condition cannot be run for it**. Four tests in that file assert properties over "every project" and all four were silently covering four of five.

## Why it stayed invisible

Exactly the shape #467 is about: the guard that would have caught it declared its own project list. The gap has existed since the fork landed and nothing could report it, because the only thing that looks is the thing that was narrowed.

It also has not bitten in practice, because every run since the fork has used a `generic*` condition, and `PROMPT_CONDITIONS` marks only `tuned` as component-consuming. So this is a latent gap in an arm that is not currently exercised — not a defect in any existing record.

## What it needs

Authoring a component, which is content and a judgement call, not a mechanical fix:

- The component states per-dataset context for the tuned arm. VOICE_PEDIATRIC's would need to say what distinguishes the pediatric cohort — its own DOI (`10.13026/h995-bt35`), protocol, and REB approval — without importing adult-cohort facts, which is the `target` failure the fitness axis names.
- It interacts with the scope declaration (#422): the manifest already declares VOICE and VOICE_PEDIATRIC as related-but-distinct, so the component must not restate that as prose or the two can drift.

Worth deciding alongside whether the `tuned` arm is still live at all. It is pinned to generic v1 as its base (`api_runner.py:78`) precisely so the 2026-07-27 tuned series stays comparable, and nothing has run it since.

Pinned meanwhile: the four tests skip VOICE_PEDIATRIC by name with a pointer here, so the gap is visible and the other four projects stay guarded.

realmarcin:
Closes #478. VOICE_PEDIATRIC has had no component since the fork, so the `tuned` condition could not be run for it. Found by widening a test that declared its own project list (#467) — which is why it hid for months.

## This was a decision, not a hole to fill

The components README is explicit:

> A project with no legitimate component gets an empty block, and its tuned condition is then identical to generic. **That is a valid result, not a gap to fill.**

So the question was empirical: does the pediatric bundle actually carry content that could be conflated? It does, and the offending line is unlabelled:

```
line 1786:  "There are currently around 833 instances."
line  391:  "23,533 recordings collected from 300 participants aged 2-18"
```

**833 is the adult count**, sitting in the pediatric bundle with no cohort attached, against this dataset's own **300**. That is exactly the conflation VOICE's component prevents in the other direction — the fact that moved three-way agreement from **76.7% to 92.4%**, the largest single effect measured in the study. The mirror of it is legitimate here.

## Everything is quoted from the bundle

23,533 recordings, 300 participants aged 2–18, DOI `10.13026/h995-bt35`, Hospital for Sick Children, Synapse for raw audio, adult release `3.1.0` — each checked against `VOICE_PEDIATRIC_preprocessed.txt` rather than copied from `VOICE.md`.

## The referent is deliberately not restated

It is declared in the manifest's `scope:` block, where `d4d download scope --check` verifies it and where the next dataset inherits the mechanism (#422). A component repeating it would be a second copy free to drift from the checkable one — the risk named when this issue was filed.

## Registry

The registry correctly refused the new file as `unpinned`, which is what #432 built it to do: a prompt never declared canonical is not a condition's text, just a file someone added. Pinned **after** committing, never before — a pin records the commit it was taken at, so pinning uncommitted bytes is refused by design.

```
10 prompt file(s), 0 not at their pin
```

`WITHOUT_COMPONENT` is now empty and all five projects are guarded by the four content assertions. Kept as a named set rather than deleted: a project with no legitimate component gets an *empty block*, which is a different state from a missing file and belongs recorded.

Full suite: **1623 passed, 4 skipped, 0 failed.**

### DIFF
--- a/src/download/prompts/canonical_hashes.yaml
+++ b/src/download/prompts/canonical_hashes.yaml
@@ -42,6 +42,16 @@ files:
       recorded prompt hashes in the corpus match these bytes, so pinning here fails nothing
       retroactively.'
     sha256: 07b59ca750a1d24ef7c7eb42d310fbfa064ddf300b4cd118383aaaebb9074186
+  src/download/prompts/components/VOICE_PEDIATRIC.md:
+    bytes: 1887
+    pinned_at_commit: 5f1ab65796518f355eec7b5bb1afc8822a79640e
+    pinned_on: '2026-08-11'
+    reason: "First canonical text for the pediatric tuned condition (#478). States the adult\
+      \ figures present in the pediatric bundle \u2014 833 instances against this dataset's\
+      \ 300 participants \u2014 so each is attributed to the cohort its source names. Mirrors\
+      \ the VOICE cohort fact, which is the study's largest measured single effect. The referent\
+      \ is left to the manifest scope declaration rather than restated here (#422)."
+    sha256: b7db067c2852b57287cb68bd1fd4f3bfc229496cc2a4bd33065bb9558a35b215
   src/download/prompts/d4d_generic_arm_prompt.md:
     bytes: 5754
     pinned_at_commit: 225f224544aa8f9ca8ac7f61f57aab9813536866
--- a/src/download/prompts/components/VOICE_PEDIATRIC.md
+++ b/src/download/prompts/components/VOICE_PEDIATRIC.md
@@ -1,0 +1,36 @@
+# VOICE_PEDIATRIC — GC-specific prompt components
+
+## fact
+
+The input bundle is scoped to the **pediatric** cohort, but it also carries
+programme-level documents that state **adult** figures, and at the point where
+they appear the cohort is not named.
+
+- **This dataset** — Bridge2AI-Voice Pediatric, PhysioNet `b2ai-voice-pediatric`
+  release 1.1.0, DOI `10.13026/h995-bt35`. Derived audio features for **23,533
+  recordings collected from 300 participants aged 2–18**, recruited at the
+  Hospital for Sick Children. Raw audio is distributed via Synapse rather than
+  the PhysioNet route.
+- **Also present in the bundle** — the Bridge2AI-Voice adult dataset,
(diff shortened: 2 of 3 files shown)


realmarcin:
## Review round

**File existence was the wrong property to guard.** The check asserted a component file was on disk; a component that rendered to *nothing* would have passed it. Verified the actual deliverable instead:

```
$ d4d api render-prompt --project VOICE_PEDIATRIC --condition tuned
# rendered tuned for VOICE_PEDIATRIC / baseline / runtime=Claude API (direct)
# sha256 132f7fab…  (6188 bytes)
```

6188 bytes against generic's 4059, with the component's content reaching the instruction. Before this it could not be rendered at all.

Three guards added: every project renders under `tuned`; a probe phrase from each component appears in its rendered instruction; and tuned is longer than generic wherever a component exists — which is what a component rendering to nothing would fail.

**Re-verified every factual claim against the bundle**, not against `VOICE.md`, since this is prompt content and an error here propagates into every tuned record:

```
line  308  "Raw Audio Data Access for Bridge2AI Voice Pediatric Cohort is via Synapse"
line  397  enrollment at the Hospital for Sick Children
line  444  REB approval at the Hospital for Sick Children
line  391  23,533 recordings from 300 participants aged 2-18
line 1786  "There are currently around 833 instances"   <- adult
```

Full suite: **1626 passed, 4 skipped.** All four CI checks green.

Issue #436, opened 2026-08-11
The canonical pin is keyed on path, so a copy at another path bypasses it


realmarcin:
Found reviewing #435, which added the registry. The gate it adds is bypassed by `cp`.

The pin is keyed on the file's path. A prompt at a path the registry does not cover is `unpinned` — reported, not fatal — so the whole check is avoided by not using the pinned path:

```
cp src/download/prompts/d4d_generic_arm_prompt_v3.md my_prompt_copy.md
# edit my_prompt_copy.md
d4d provenance record ... --prompt my_prompt_copy.md --prompt-text sent.txt
```

Measured on a scratch tree:

```
copied-path record -> ('unpinned', 'no canonical hash pinned for my_prompt_copy.md')
condition_of       -> None
label mismatch     -> None
```

Three checks and all three are silent. `canonical_prompt_status` says `unpinned`, which `d4d runs check` prints and does not fail. `condition_of()` matches on filename, so it cannot name the condition, and `prompt_condition_mismatch` therefore reports nothing either — while the run label still says `generic-v3`.

## What closes it

Coverage, not just agreement. The label already claims a condition (`condition_from_label`), and `CONDITION_PROMPTS` says which file that condition is. So require the claim's pinned file to be **among** the record's prompt files:

- label claims `generic_v3`, record hashes the pinned `d4d_generic_arm_prompt_v3.md` → as now
- label claims `generic_v3`, record hashes nothing pinned for that condition → `uncanonical`, fatal under `--strict`

A run that names no condition in its label stays silent, for the same reason `prompt_condition_mismatch` is silent there: labels predating the convention name none, and failing them punishes history.

That still leaves an unlabelled run free to use any prompt it likes. Worth saying plainly rather than fixing by making labels mandatory in the same change.

realmarcin:
Fixed in #435 (`abed21e2`) with a coverage check: a label that claims a condition must hash some pinned condition prompt, else `uncanonical`, fatal under `--strict`. Not the *claimed* condition's exact file — that would fail the 16 known #420 records retroactively. An unlabelled run is still unconstrained; that part stands.

realmarcin:
Closed by #435, merged as 1f3a9db8.

PR #491, opened 2026-08-11
Recover historical prompts as of each run's own commit (#399)

### LINKED ISSUE #399: No route to record prompts for historical agentic runs (backfill has no --prompt)
Follow-up to #395, which adds `--prompt` to `d4d provenance record` (the live path only).

Every Claude Code agentic record written before that PR carries `prompts: null`, because the CLI never exposed `prompt_paths`. `d4d provenance backfill` does not accept prompt paths either, so there is currently no way to record a prompt for an existing run even when the prompt file is known with certainty — for example, a run whose `# Prompt:` header names `src/download/prompts/d4d_generic_arm_prompt.md` and whose label sits inside the window where that file was unchanged.

Worth being careful here rather than bulk-filling, because the honest answer differs per run:

- **Recoverable**: the header names a prompt file, and the file's contents at that run's commit can be recovered from git. The hash should be of *the bytes at that commit*, not of today's file — the generic prompt has been edited (v1/v2/v3), so hashing the current file would assert that an older run used today's prompt.
- **Not recoverable**: runs launched with an inline prompt that was never saved. `prompt_facts()` already has the right note for these, and the 2026-07-27 tuned prompts are called out in its docstring as exactly this case. They should stay null.

So this is not a `--prompt` flag on `backfill` so much as a resolver: given a run's label, commit, and `# Prompt:` header, recover the prompt bytes as of that commit or mark them unrecoverable. That keeps `backfill`'s existing contract — record what can be honestly recovered, mark the rest — rather than letting an operator assert a hash by hand.

realmarcin:
Closes #399. A **resolver**, not a `--prompt` flag on `backfill` — the honest answer differs per run, and an operator asserting a hash by hand is exactly what must not be possible.

```
already_recorded   77
no_prompt_header   66     supplied inline, never saved — stays null
recovered          15     from commit fa90b4ec
```

## A bug found on the way in

`Prompt` was missing from `HEADER_FIELDS`, so `parse_header` silently discarded the one field a historical prompt could be recovered from. A survey using it reported **0 of 85** records carrying a prompt header; reading the raw files reported **15**.

## The hash is of the bytes at the run's commit, never today's

Not a hypothetical precaution. `d4d_generic_arm_prompt.md` was edited on 2026-07-29 — the day *after* the 15 runs that name it — 16 lines apart:

```
at 2026-07-28 (what those runs read)   7e9a67f7…
at 2026-07-29 and today                0fbc626b…
```

Recording today's hash would assert a 2026-07-28 run used a prompt that did not yet exist: the fabricated-provenance failure this module exists to prevent.

## Running it surfaced a design question

The 15 promptly failed `runs check --strict` as `uncanonical`. That is *literally* true — the registry never pinned those bytes, because it postdates them — but it puts honest recovered evidence in the same **fatal** bucket as the `cp`-to-another-path bypass (#436), where the bytes are attested by nothing.

These are attested by git. `git show <commit>:<path>` reproduces **all 15** hashes and byte counts exactly, asserted by test.

So `canonical_prompt_status` gains **`pre_registry`**, reported like `superseded` and `unpinned` rather than failed:

```
⚠️  AI_READI  2026-07-28_claude-opus-5-generic_rep1  pre_registry: recovered from
    fa90b4ec036f as of 2026-07-28; the bytes are attested by git rather than by
    the pin registry, which postdates this run
```

Without it, recovering honest evidence would make a record *worse* than leaving it blank — exactly backwards.

## Each record says it is a recovery

A `recovery` block names the commit, the date resolved as of, and a note that the hash is not of the file today. Otherwise a reader cannot tell a recovery from a live capture, and the two are not the same evidence.

Twelve tests, including that the recovered hash still differs from the current file's — if they ever coincide, the test has stopped proving anything.

Full suite: **1669 passed, 4 skipped.** `d4d runs check --strict` exits 0.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -421,8 +421,19 @@ edited into the prompt file *before* rendering re-renders to itself and reports
 check` is fatal under `--strict` on `uncanonical` (a prompt that was never
 pinned, or a labelled condition whose record hashes no condition prompt at all —
 the `cp`-to-another-path bypass, #436) and on `missing` (a pinned path the
-record hashed nothing for). `superseded` (pinned once, since rotated) and
-`unpinned` are reported and not failed.
+record hashed nothing for). `superseded` (pinned once, since rotated),
+`unpinned`, and `pre_registry` are reported and not failed.
+
+`pre_registry` is a prompt recovered from git at the run's own commit
+(`d4d provenance backfill-prompts`, #399). Those bytes are attested by a
+different instrument than the registry and predate it, so calling them
+`uncanonical` would put honest recovered evidence in the same bucket as the
+`cp`-to-another-path bypass, where the bytes are attested by nothing. Every
+recovered hash is reproducible with `git show <commit>:<path>`, and a test
+asserts it. **The hash is of the bytes at that commit, never today's** —
+`d4d_generic_arm_prompt.md` was edited the day after the 15 runs that name it,
+16 lines apart, so today's hash would assert they used a prompt that did not
+yet exist.
 
 This is not tamper-proofing. Whoever can edit a prompt can rotate its pin.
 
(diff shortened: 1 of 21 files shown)


realmarcin:
## Review round — CI caught a portability defect the local suite could not

All three Python versions failed while the local suite passed:

```
AssertionError: 128 != 0 : src/download/prompts/d4d_generic_arm_prompt.md
```

The cause is the **clone, not the corpus**. `actions/checkout` fetches at depth 1, so every commit these records name is unreachable and `git show` exits 128. A failure there reports on the checkout depth and says nothing about whether the recovered hashes are right.

The test now probes `git rev-parse --is-shallow-repository` and skips. Locally — where the history is present — all 15 hashes and byte counts are verified for real.

Worth noting the **resolver already degrades the same way by design**: in a shallow clone it returns `no_commit_at_or_before_run` rather than falling back to today's file. That is the behaviour that actually matters, since the alternative is exactly the fabricated hash this whole issue is about.

Full suite: **1669 passed, 4 skipped.** All four CI checks green.

Issue #456, opened 2026-08-11
unit, data_substrate and data_topic are ranged uriorcurie but hold quantities and prose — a modelling error, not bad data


realmarcin:
Surfaced by `d4d runs identifiers` (#402, landed in #455 as `48b26f4a`).

Three slots ranged `uriorcurie` are not holding identifiers, and were never going to:

| slot | values | unresolvable | what they actually hold |
|---|---|---|---|
| `unit` | 148 | **148 (100%)** | `%`, `years` |
| `data_substrate` | 68 | 22 | *"Multimodal per-participant records comprising…"* |
| `data_topic` | 120 | 22 | *"Type 2 diabetes mellitus and associated health outcomes"* |

`unit` is the clearest: every value in the corpus is a quantity unit, and not one is an IRI or a CURIE. That is not a data-entry problem that a validator pattern would catch and generators would then fix — it is a slot whose declared range does not describe what the slot is for. Adding a `pattern` to `uriorcurie` (see the migration issue) would turn all 148 into validation failures overnight and the correct remedy would still be a schema change.

`data_substrate` and `data_topic` are more mixed — some values *are* CURIEs, which is presumably the intent, and the schema declares `B2AI_SUBSTRATE` and `B2AI_TOPIC` prefixes that look built for exactly this. So those two may be a real vocabulary that records fall back to prose when they cannot find a term in. Worth checking whether the registry has the terms before concluding either way.

Suggested:

1. Re-range `unit` to a unit vocabulary or a plain string with a pattern. `qudt` is already a declared prefix if a term-based unit is wanted.
2. For `data_substrate` / `data_topic`, decide whether the B2AI registry can supply terms for the prose values. If it can, this is data entry; if it cannot, the range is wrong in the same way `unit`'s is.
3. Either way, decide before the migration in the sibling issue — a migration that assumed these were mis-entered identifiers would repair the wrong thing, and 192 of the corpus's 7,415 unresolvable values are in these three slots.

Issue #426, opened 2026-08-10
A validation verdict does not pin the schema it was reached against


realmarcin:
Raised reviewing #423 before merge. Partly mitigated there; the underlying gap remains.

"Validates" is a claim about a record **against a schema**. `validation.artifacts` pins only the record:

```yaml
validation:
  passed: true
  artifacts:
    full: {path: …/VOICE_d4d.yaml, md5: a050a231…}
    core: {path: …/VOICE_d4d_core.yaml, md5: fd47c062…}
  recorded_by: d4d runs validate
```

No schema hash. The record carries one separately —
`schema.full_sha256`, `schema.core_sha256` — but the verdict does not reference it.

So `validation_status()` re-hashes the artifacts, finds them unchanged, and reports VALID for a record that may never have been checked against the schema now in the tree. A schema change that tightens a constraint makes every prior verdict a claim about a check that no longer exists, and nothing notices.

## Why it surfaced now

Before #423, the gap was bounded by accident: `d4d provenance record` deleted the `validation:` block on every re-record, so a verdict rarely outlived much. #423 carries it forward when the artifacts are unchanged — correct on its own terms, and it would have let a verdict outlive its schema.

**Mitigated in #423**: `preservable_validation()` also compares the record's `schema` block and drops the verdict if either hash moved. So a *carried* verdict cannot outlive its schema.

**Not fixed**: a verdict written by `d4d runs validate` and never re-recorded still has no schema binding, and `validation_status()` still does not check one. The 55 records currently carrying a verdict are in that state.

## Suggested fix

Record the schema hashes inside the verdict at the point it is written:

```yaml
validation:
  passed: true
  artifacts: {...}
  schema: {full_sha256: …, core_sha256: …}
```

and have `validation_status()` return STALE when they differ from the schema on disk — the same treatment an edited artifact already gets. The distinction matters for the same reason it does there: "validated, then the schema moved" is a different problem from "never checked", and both mean do not rely on it.

Cheap, because both hashes are already computed for the `schema` block of every record.

Issue #403, opened 2026-08-07
No vocabulary for data standards: conforms_to is free text, conforms_to_schema is dead, and standards cannot project into core


realmarcin:
Found analysing label `2026-08-07_claude-opus-5-claudecode-generic-v3_rep1`. Related to #402 (both are cases where the schema gives no usable slot and records improvise).

## Three overlapping slots, one of them used

`Dataset` offers all of:

| slot | range | used in this label |
|---|---|---|
| `conforms_to` | string | **8 distinct values across 4 of 5 projects** |
| `conforms_to_schema` | string | never |
| `conforms_to_class` | string | never |

Every record put a *schema or data standard* into `conforms_to` while `conforms_to_schema` — the slot whose description is literally "The schema or data model to which the resource conforms" — sat empty in all five. When two slots are near-synonyms and neither is constrained, records pick the shorter name. That is a slot-selection trap, and it means the corpus cannot be queried for "what schema does this conform to".

## The values are unconstrained and inconsistently shaped

```
ASCII File Format Guidelines for Earth Science Data (ESDS)          AI_READI
Brain Imaging Data Structure (BIDS) v1.9.0                          VOICE_PEDIATRIC
Clinical Dataset Structure (CDS) v0.1.1                             AI_READI
Digital Imaging and Communications in Medicine (DICOM)              AI_READI
Open mHealth                                                        AI_READI
The Observational Medical Outcomes Partnership (OMOP) CDM           AI_READI
WaveForm DataBase (WFDB)                                            AI_READI
https://www.researchobject.org/ro-crate/                            CM4AI
```

Long-name-plus-parenthetical, bare name, name-with-version, and a raw URL — four shapes for one concept. None is resolvable to a term.

## They cannot reach the core record at all

`CoreDistribution.format` is ranged on `FormatEnum`, whose members are file serializations only:

```
BZ2 CSV DOCX GZ HTML JSON JSONL MD PDF PPTX TAR TSV TXT XLSX XML XZ YAML ZIP
```

DICOM, WFDB, OMOP CDM, BIDS, Open mHealth and ESDS are **data standards, not serializations** — a different concept from the one `FormatEnum` models. So there is no slot on `CoreDistribution` that can hold them, and the semantic exchange layer silently loses the single most interoperability-relevant fact about the distribution.

[...]

PR #506, opened 2026-08-12
Give governance and stewardship somewhere to live (#503)

### LINKED ISSUE #503: No governance or stewardship class exists; the one DAC contact slot is nested under ExportControlRegulatoryRestrictions

> Thin on Governance and Stewardship details including DAC members and contacts and related info.

## The content is thin because the schema has almost nowhere to put it

Checked against the merged schema. **`Dataset` has no governance or stewardship slot at all** — no class named for governance or stewardship exists:

```
governance classes in merged schema: []
```

`D4D_Data_Governance.yaml` exists and is imported, but the classes it declares are:

```
LicenseAndUseTerms
IPRestrictions
ExportControlRegulatoryRestrictions
```

— licensing, IP and export control. The module is named for governance and contains none.

## The one governance field is nested under export controls

There is exactly one relevant slot, and it lives in the wrong place:

```
governance_committee_contact lives on: ExportControlRegulatoryRestrictions
```

So to record who chairs the data access committee, a record must nest that person inside a statement about **export control regulations**. Two of the fresh records do exactly that:

```yaml
governance_committee_contact: https://orcid.org/0000-0003-4535-3486
governance_committee_contact: mailto:DACO@b2ai-voice.org
```

Both are correct answers filed under a heading that does not describe them.

## What is absent entirely

No slot for: the data access committee as an entity, its **members**, its composition or appointment, its review process or turnaround, escalation or appeal, stewardship roles distinct from the PI, or the body accountable after the grant ends. A run cannot be thin or thorough on these — it has no field to be either in.

## Why this is a schema issue and not a prompt issue

Prompt changes cannot fix it. This is the same shape as #403 (`conforms_to*`) and #456 (`unit`): the record improvises because the schema offers no fitting slot, and the improvisation is then judged as the record's failure.

## Suggested

A `DataGovernance` class reachable from `Dataset`, holding the committee as an entity with members, contacts, remit and process, and moving `governance_committee_contact` onto it. Adding slots changes the schema digest and therefore the comparison pin, so it wants sequencing against the in-flight sweep (`notes/generic_v1_v3_analysis_plan.md`).

realmarcin:
Closes #503. From Camille Nebeker's review: *"thin on Governance and Stewardship details including DAC members and contacts"*.

**The content was thin because there was almost nowhere to put it.** `Dataset` had no governance or stewardship slot at all, and the module *named* `D4D_Data_Governance` declared only `LicenseAndUseTerms`, `IPRestrictions`, `ExportControlRegulatoryRestrictions`. The one relevant field sat on the export-control class:

```
governance_committee_contact lives on: ExportControlRegulatoryRestrictions
```

So naming a data access committee chair required nesting that person inside a statement about export regulations — and two records in the 2026-08-11 arm do exactly that. Both are correct answers filed under a heading that does not describe them.

## What was added

A `DataGovernance` class reachable from `Dataset`: committee name, **members** (one per person, not collapsed), contact, review process, decision timeframe, appeal route, stewardship roles distinct from creators, and the organization accountable **after the funding ends**.

The descriptions carry the constraints that keep it honest — record only members the sources name, since committee membership is frequently undisclosed and an absent list is a correct answer. The class also says plainly how it differs from `LicenseAndUseTerms` (the terms) and `ExportControlRegulatoryRestrictions` (the regulations): this is the people and process that *apply* them.

The old slot is **deprecated in place, not moved**. Removing it would invalidate the two records that already populate it.

No prompt change could have fixed this — a run cannot be thorough about a committee's membership when no field accepts it. Same shape as #403 and #456: the record improvises because the schema offers no fitting slot, and the improvisation is then read as the record's failure.

## Two consequences handled rather than left

- **The digest moves `488bd732` → `659aae67`.** Every 2026-08-11 record was re-validated so its verdict pins the new schema instead of going STALE — #426's mechanism working as designed.
- **The structural SSSOM mapping** gained exactly two regenerable rows (`DataGovernance/name`, `/description`). Added by hand rather than regenerating, because regeneration still drops the 10 rows #234 tracks. The summary *was* regenerated, and the check now reports only that known gap.

## A test of mine was wrong

`test_the_digest_is_the_one_on_disk` asserted a record's digest equals the **live** one — true only at the moment I wrote it. A record whose digest tracked the working tree would be worthless; the point is to pin what *that run* saw. Corrected, plus a new test that a superseded pin makes the verdict stale rather than silent.

Full suite: **1701 passed, 4 skipped.**


realmarcin:
**Note on the diff size.** This PR is 61 files / ~54.7k insertions, and only four of those files are the schema change. The rest is the 2026-08-11 sweep output, swept in by a broad `git add -A` while I was working on both at once.

Some of it genuinely belongs here — every 2026-08-11 provenance record was re-validated *because* this change moved the digest, which is #426's mechanism and part of what this PR should be judged on. The five fresh full/core records and their reconciliation reports do not; they are generation output from the authorized run and would have been clearer in their own PR.

Flagging rather than rewriting history on an open branch. **To review the actual change, use:**

```
git diff main...close-503-governance -- src/ tests/ data/semantic_exchange/
```

which is the four schema/test files plus the two SSSOM mapping rows.

realmarcin:
## Review

Checked the substance rather than restating it.

**No collision** — verified no governance or stewardship class existed before this; `Dataset` genuinely had nowhere to put any of it.

**The deprecated slot still works** — both records that populate `ExportControlRegulatoryRestrictions.governance_committee_contact` validate unchanged. Removing it would have invalidated them, which is why it was deprecated in place rather than moved.

**`d4d runs check --strict` exits 0** on this branch, and all six CI checks pass.

**One finding, filed as #510 rather than widening this PR.** `CoreDataset` carries `license_and_use_terms` and `regulatory_restrictions` — the other two classes from this same module — but not `data_governance`. So the exchange layer can say what the licence permits and which regulations apply, but not who decides access, which is arguably the one a consumer most needs to know whether they can obtain the data at all.

Leaving core alone kept this PR's blast radius to `Dataset`, which was the cautious call, but it is not a reasoned position about what belongs in core. The VOICE_PEDIATRIC agent independently hit the same shape (`related_datasets` missing from core), and #403 was a third instance. Worth deciding once — *what is `CoreDataset` for?* — rather than per-slot each time, which is how the asymmetry accumulated.

Merging.

Issue #515, opened 2026-08-12
The rendered spec says use the header block 'exactly'; the playbook requires two more lines on core files


realmarcin:
Reported independently by three of the fifteen runs in the 2026-08-11 agentic v1 arm (AI_READI rep3, CM4AI rep3, VOICE rep3), each flagging it rather than silently choosing.

Two instruction sources disagree about the core record's header, and each is authoritative in its own terms.

**The rendered launch spec** says:

> HEADER BLOCK — use exactly:
> […]
> The core file uses "phase 2" and the core schema path in the corresponding lines.

So: this block, with two substitutions, and nothing else.

**The playbook** (`.claude/commands/d4d-full-core.md`) requires more on the core file:

- `# D4D Core Datasheet …`
- `# Sources:` naming the same-run full record — required by the provenance guard, so a core record states what it was projected from
- `# Phase 4 reconciliation: completed` — a completion criterion, and `--sync-core` appends it automatically

All three runs followed the playbook and documented the deviation. That is the right call — a core record that does not name its full input is worse than one with two extra header lines — but they should not have had to make it, and a run that followed the spec literally would produce a record failing the playbook's own completion criteria.

## Why this is not cosmetic

`# Sources:` is what ties a core record to the full record it was projected from. If a run takes the spec literally and omits it, the pair's relationship exists only in the reconciliation report, not in the record.

It also means the header is **not** byte-reproducible from the spec for core files, so any future check comparing a core header against its rendered spec will report a mismatch on every correct record — the same false-alarm shape as #500.

## Options

1. **Have `render-prompt` emit the core header block in full**, including the two playbook-required lines, so the spec and the playbook cannot disagree. The renderer already knows the label, project and method; it can name the full record's path. Changes the rendered text, so it re-baselines the condition and needs a pin rotation.
2. **Change the spec's wording** from "use exactly" to something that acknowledges the playbook adds lines. Cheapest, and leaves the header unreproducible.
3. **Move the requirement out of the playbook** into the renderer only. Same effect as (1) with more churn.

Recommend (1), sequenced with whatever other prompt change lands next, since it re-baselines the condition either way.

## Scope

Does not affect the 15 records already produced — all validate, all reconcile, and all three affected runs documented the deviation in their reports.

realmarcin:
Second review pass on #516 measured the actual headers of all 15 core records. The conflict is worse than "two sources disagree" — it has already produced four spellings of the same line, and nothing parses any of them.

[...]

Issue #497, opened 2026-08-12
The agentic path records no schema digest and no render spec, so its instruction is unverifiable — the defect #454 is about


realmarcin:
Found by the CHORUS canary that precedes the 30-run sweep — one run on each path, compared field by field.

Both records validate and both carry live provenance. Two fields the API path records and the agentic path does not:

| field | API | agentic |
|---|---|---|
| `schema.digest_md5` | `488bd7321fe5` | **absent** |
| render spec behind the instruction | absent | absent → `unverifiable` |

## 1. The agentic arm records no schema digest

`d4d api run` computes `schema_digest.fingerprint(...)` and passes it to `build_record`. `d4d provenance record` — the agentic path's only route — has no such flag, so the field is simply missing.

That matters more than usual right now: the `Dataset` digest moved **twice today** (`e802cdc3` → `488bd732`, from #456 and #403). A record with no digest cannot be placed on the axis the whole v1-vs-v3 comparison is stratified by, and `validation_status` cannot detect that its verdict was reached against a schema that has since moved (#426) — the gap #433 counts.

## 2. The recorded instruction cannot be confirmed

```
1 run(s) whose recorded instruction could not be confirmed against its spec:
   ⚠️  CHORUS  2026-08-11_…-claudecode-generic_rep1
       unverifiable: request hash recorded without the spec that produced it
```

The render gate (#429) re-renders a record's spec and compares it to the recorded instruction. `d4d provenance record` accepts `--prompt-text` but has no flag for the spec that produced it, so an agentic run can record *what it was sent* and never *what should have been sent*. The comparison is one-sided by construction.

**This is the defect #454 is about.** That sweep's fault was carrying no recorded instruction; an arm launched to replace it should not carry an instruction nobody can verify. The instruction here *was* rendered — with `d4d prompt render`, exactly as the playbook now says — so the evidence exists and only the recording lacks a place to put it.

## Why this blocks the fan-out rather than being filed for later

The standing canary rule: if the canary needs a fix, fix it and re-canary; do not fix and fan out in the same step. Fifteen agentic runs whose instruction cannot be verified would reproduce, at the same scale, the exact defect the arm exists to repair.

## Fix

`d4d provenance record` gains `--schema-digest` (or computes it, since the schema is on disk and the run just validated against it) and `--render-spec`, mirroring what `api_runner` already passes to the same `build_record`.

PR #516, opened 2026-08-12
Replace the mislabelled canonical sweep with a fresh v1 arm (#454)

### LINKED ISSUE #454: The 2026-08-07 agentic sweep is labelled generic-v3 but ran the generic prompt: 15 runs, 30 records

realmarcin:
Closes #454.

## What was wrong

The canonical records for all five projects were labelled `generic-v3` and were not generated under v3. They also carried no recorded render spec, so the render gate could not re-render them and reported `unverifiable`.

Neither is fixable after the fact. A run's instruction is evidence; it is collected at run time or not at all. Relabelling would assert a provenance no artifact supports.

## What this is

A fresh arm: 15 runs, 5 projects × 3 replicates, condition `generic` (v1), agentic path, with `d4d provenance record` writing the render spec (#497).

All 15: `gate=match`, `pin=canonical`, `drift=current`, `verdict=valid`. `d4d runs check --strict` exits 0.

## Replicate variance

| project | slots | sd |
|---|---|---|
| AI_READI | 80 / 77 / 82 | 2.52 |
| CHORUS | 50 / 51 / 46 | 2.65 |
| CM4AI | 53 / 63 / 50 | **6.81** |
| VOICE | 81 / 82 / 82 | 0.58 |
| VOICE_PEDIATRIC | 66 / 69 / 67 | 1.53 |

Mean within-project sd 2.81 slots.

**Read CM4AI before drawing a single-replicate comparison from this arm.** A 13-slot spread between replicates of one condition is larger than several between-condition effects reported elsewhere in this repo.

## Canonical marks

Moved here. The superseded records keep a `canonical_superseded_by` tombstone rather than losing the `canonical` block, so the earlier selection stays readable.

## Self-caught defects worth noting

Three runs caught and corrected their own errors before writing, which is the behaviour the reconciliation phase is for:

- **AI_READI rep3** had silently re-mapped BMJ protocol ICD-10 codes to the clinically expected assignment — a plausible correction of the source, which is exactly what the record must not do.
- **VOICE_PEDIATRIC rep3** caught four fabrications: author count 130 against 121 verified, an ORCID-shaped id in no source, twelve institution URLs from model memory, and an over-claimed `conforms_to: BIDS`.
- **VOICE rep3** removed invented IANA media types.

## Filed, not fixed here

#515 — the rendered spec says to use its header block "exactly"; the playbook requires `# Sources:` and `# Phase 4 reconciliation: completed` on core files. Three runs hit it independently and all three followed the playbook. Fixing it re-baselines the condition and needs a pin rotation, so it does not belong in this PR.

## Not in scope

The API arms are stopped and this PR does not restart them. §6's v1-vs-v3 promotion decision needs the API path, where v1 and v3 actually differ; this arm establishes the agentic v1 baseline and the variance floor to read any such comparison against.


realmarcin:
## Review, rounds 1–2

Read-only pass over the arm. Two findings survived verification; three did not and are recorded below so they are not re-derived.

### Confirmed and addressed in this PR

**#517 — the arm straddles two schemas.** rep1 was generated at digest `488bd732`, rep2 and rep3 at `659aae67`, because #503 landed mid-arm and added `data_governance`. All five rep1 records lack the class; all ten rep2/rep3 records have it. A 15/15 split along the schema boundary is structural, not model variance.

The published sd figures were therefore not a clean replicate estimate: mean 2.81 raw, 2.67 with the forced slot removed. Both numbers are now stated rather than the first alone.

Fixed at the mechanism, not just the claim: `d4d runs check` now reports a replicate series carrying more than one generation digest. It finds exactly this one series across the whole corpus and nothing else. Reported, never fatal — a straddled series is usable if it is known to be straddled.

The analysis plan already contained the instruction that would have prevented it (*"if it has moved again, generate both arms after the move rather than one either side of it"*) and it did not fire, because a plan is read once and a sweep runs unattended. Plan amended with what happened; its stale `488bd732` pin corrected.

### Confirmed, filed, not fixed here

**#515 — the core header.** Measuring all fifteen records found the line naming the projected-from full record spelled four ways: `# Sources:` (8), `# Full D4D input:` (5), `# Full record input:` (1), and once wrapped with no key. None are in `HEADER_FIELDS`, so `parse_header` discards it in all fifteen.

Severity is bounded: the full record is hashed at `validation.artifacts.full.md5` in all 15, and `validation_status` re-hashes it and goes `STALE` on change. The link is verified; the header line is documentation of it. Fixing this re-baselines the condition and needs a pin rotation, so it stays filed.

### Checked and dissolved

- **Core records carry top-level keys absent from their full record** (10 of 15). Not a projection defect — `distributions` and `dialect` are `CoreDataset` slots that do not exist on `Dataset`.
- **rep1 reports `valid` while pinned to a superseded schema.** Not a missed `STALE`. `validation.schema` pins the schema a *verdict* was reached against; generation happened before #503 and validation after, so naming the old digest in `schema.digest_md5` and the current schema in `validation.schema` is the only way to state both facts. A straddle check reading the validation pin would have seen nothing, which is why #517's check reads the generation digest.
- **`Generated: 2026-08-12` in headers on a `2026-08-11` label.** Real but not a defect: the core phase ran 00:51–06:55 UTC. The label is the arm's date, not each record's timestamp, and `record_generated_at` carries the exact time.

### Not reviewed

Factual accuracy of the extracted content beyond the three self-caught cases in the PR body. That is a corpus question, not a diff question.

Issue #517, opened 2026-08-12
The 2026-08-11 v1 arm straddles two schemas: rep1 predates #503, rep2/rep3 do not


realmarcin:
Found reviewing #516. The 15-run agentic v1 arm of 2026-08-11 is described as five projects × three replicates of one condition. On the generation axis it is not: **rep1 ran against a different schema than rep2 and rep3.**

```
schema.digest_md5 at generation time
  rep1                    488bd732     (pre-#503)
  rep2, rep3              659aae67     (post-#503, adds `data_governance`)
```

The split is perfect and it shows in the output:

```
                 rep1              rep2             rep3
AI_READI         80  dg=False      77  dg=True      82  dg=True
CHORUS           50  dg=False      51  dg=True      46  dg=True
CM4AI            53  dg=False      63  dg=True      50  dg=True
VOICE            81  dg=False      82  dg=True      82  dg=True
VOICE_PEDIATRIC  66  dg=False      69  dg=True      67  dg=True
```

All five rep1 records lack `data_governance`; all ten rep2/rep3 records have it. A 15/15 split along the schema boundary is the signature of a structural cause, not model variation — #503 landed mid-arm and rep1 could not have populated a class that did not exist when it was generated.

## What this does and does not invalidate

**Does not**: the records are correct and internally coherent. Each names the schema it was generated against, which is the digest working as designed. `validation.schema.*_sha256` is current for all fifteen because validation is a later act than generation, correctly pinned to the schema it used — that is not a discrepancy.

**Does**: the variance figure published in #516, and the framing of the arm as a homogeneous replicate set.

```
                  raw sd     sd with the forced governance slot removed
AI_READI          2.52       2.65
CHORUS            2.65       2.89
CM4AI             6.81       6.66
VOICE             0.58       0.00
VOICE_PEDIATRIC   1.53       1.15
mean              2.81       2.67
```

The headline moves 2.81 → 2.67, which is small — but "small" is a fact established after the check, not a reason to have skipped it, and the adjustment is crude: it assumes the class contributes exactly one top-level slot and nothing else. Whether the presence of a new governance class changed how a run allocated facts across *other* slots is not measured here.

## Why it happened

Nothing forbids the schema moving mid-arm, and nothing detects it afterwards. `d4d runs check --strict` exits 0 on this arm — correctly, since every individual record is sound. The defect is only visible across records, and no check looks across them.

## Proposed

[...]

Issue #499, opened 2026-08-12
conforms_to_class and conforms_to_schema are strict-identity but describe the record, so they are unrepresentable in a valid full/core pair


realmarcin:
Found by the AI_READI agentic run of the 2026-08-11 sweep, which hit it in Phase 4 and resolved it by omitting the slot from both records — the only option that neither fails the gate nor writes a false claim.

## The contradiction

`conforms_to_schema` and `conforms_to_class` describe **the record itself**, so their correct values necessarily differ between a full/core pair:

| slot | correct in full | correct in core |
|---|---|---|
| `conforms_to_schema` | `…/data-sheets-schema` | the *core* schema |
| `conforms_to_class` | `Dataset` | `CoreDataset` |

But all three are strict-identity slots in the pair check:

```
conforms_to_class  is a strict-identity slot: True
conforms_to        is a strict-identity slot: True
conforms_to_schema is a strict-identity slot: True
```

Strict identity means full and core must hold the *same* value. So any honest population of `conforms_to_class` fails `d4d_pair_consistency`, and `--sync-core` would actively copy full's `Dataset` into core, writing a false claim about what the core record instantiates.

**The slot is unrepresentable in a valid pair.** Omitting it from both is the only passing state, which is what the run did — but that means a self-description slot the schema offers can never be used.

`conforms_to` is *correctly* strict-identity: a data standard the dataset follows is a fact about the dataset and is the same in both records. The three are not the same kind of slot, which is exactly the distinction #403 drew.

## My part in surfacing it

#403 rewrote these descriptions to say plainly that `conforms_to_schema` and `conforms_to_class` are statements about the record while `conforms_to` is about the data — and gave `conforms_to_class` the explicit values `Dataset` and `CoreDataset`. The contradiction predates that change; the old description ("the specific class or type within a schema to which the resource conforms") was vague enough that nobody had to confront it. Making the semantics explicit made the conflict explicit.

## Options

1. **Exempt record-self-description slots from strict identity.** `_slot_value_signature` derives identity from the schema, so this needs a marker — an annotation such as `d4d:perRecord: true` on the two slots, read by `load_pair_schema`. Correct, and it makes the category visible in the schema rather than in a hardcoded list.
2. **Hardcode the two names as projected rather than identical.** Cheaper; reintroduces exactly the hand-maintained-list problem #431 just removed elsewhere.
[...]

Issue #441, opened 2026-08-11
The scope check reads only the record's `id`; 32 records absorb the other cohort one level down


realmarcin:
Found reviewing #440, which added the check. It reads the record's `id` and nothing else, reports 171/171 in scope, and misses the failure mode the scope declaration exists to describe.

Walking every record's whole tree for the identifiers VOICE declares related-but-distinct:

```
32 record(s) with a related-but-distinct id outside related_datasets

2026-07-27_claude-opus-5_rep1/VOICE_d4d.yaml
    resources[1].doi = https://doi.org/10.13026/h995-bt35
    resources[1].download_url = https://physionet.org/content/b2ai-voice-pediatric/1.1.0/
    resources[1].version_access.latest_version_doi = https://doi.org/10.13026/mf9s-5r03
    resources[1].distribution_formats[0].access_urls[0] = …
2026-07-28_claude-opus-5-deprimed_rep2/VOICE_d4d.yaml
    resources[1].id = https://doi.org/10.13026/h995-bt35
    file_collections[3].download_url = …
```

`distribution_formats[].access_urls` and `file_collections[].download_url` pointing at the pediatric release say *you obtain this dataset there*. That is the #292 shape one level down from the identifier: not "the record is about the wrong dataset" but "the record absorbs the other dataset into its own distribution".

## What to do

Report it, do not fail it. Two reasons:

- 32 existing records carry it, and failing history retroactively is the error the live-provenance cutoff exists to avoid.
- Not every placement is wrong. A VOICE record may legitimately cite the pediatric PhysioNet page as a *resource* — the page is in its bundle. `access_urls` and `download_url` are the ones that make a claim about where the adult data lives, and even there the line between "cited" and "absorbed" is a judgement a rule should surface rather than settle.

So: a second verdict alongside `out_of_scope`, listing the paths, reported by `d4d download scope --check` and never fatal. The identifier check stays as it is — the part a rule can settle.

realmarcin:
Fixed in #440.

realmarcin:
Closed by #440.

Issue #431, opened 2026-08-10
AGENT_PLAYBOOKS is a hand-maintained list of which files carry decision rules


realmarcin:
Raised reviewing #429.

`AGENT_PLAYBOOKS` in `provenance.py` is a hardcoded tuple:

```python
AGENT_PLAYBOOKS = (
    Path(".claude/commands/d4d-full-core.md"),
    Path(".claude/commands/d4d-agent.md"),
    Path(".claude/agents/d4d-provenance-guard.md"),
)
```

Three properties follow, and two of them are wanted:

- **Fixed set, always the same three.** A missing file is recorded as `exists: false` rather than omitted, so a record always says what it looked for. Wanted — it distinguishes "this run followed no playbook" from "this field was not populated".
- **A renamed playbook silently becomes `exists: false`.** Nothing fails, nothing warns, and the record now attests that a file the agent actually read was absent. Not wanted.
- **A *new* playbook is invisible.** If a fourth instruction file is added under `.claude/` and the agent is told to read it, records will not mention it and nothing will notice.

The second and third are the same defect: the list is a claim about which files carry decision rules, maintained by hand, with nothing checking it against the tree.

## Why it matters here specifically

This is the file where `.claude/commands/d4d-agent.md` carries the slot-filling contract that gives the agentic path its v2/v3 behaviour (#394, #420). The whole reason for hashing playbooks is that instructions were reaching the model through files nobody was tracking. A hand-maintained list of which files those are reproduces the original problem one level up.

## Options

1. **A test that the tuple matches what the playbooks reference.** The launch prompt's first lines name `d4d-provenance-guard.md` and `d4d-full-core.md` explicitly; `d4d-full-core.md` names `d4d-agent.md`. So the set is derivable by following those references from the prompt body, and a test could assert the tuple equals the closure.
2. **Glob `.claude/commands/d4d-*.md` and `.claude/agents/d4d-*.md`.** Simpler, but hashes files no run reads, and then a record's playbook list changes whenever an unrelated command is added.
3. **A test that every file in those directories is either in the tuple or on an explicit ignore list.** Forces a decision when one is added.

Recommend (1) if the reference-following is reliable, else (3). Either way the failure mode to close is *silent* drift, not the tuple existing.

Not urgent — the three named files are correct today, and `exists: false` at least makes a rename visible to anyone reading a record. It is the "anyone reading a record" part that is doing too much work.

PR #519, opened 2026-08-12
Core exchange completeness and pair-model correctness (#510, #499, #518, #521)

### LINKED ISSUE #510: CoreDataset carries two of three governance classes but not the third, and related_datasets is missing too — what is core for?
Review finding on #506, and it extends beyond that PR.

`CoreDataset` carries two of the three Data Governance module classes and not the third:

```
core has license_and_use_terms:    True
core has regulatory_restrictions:  True
core has data_governance:          False    <- added by #506, full only
core has related_datasets:         False
```

So the exchange layer can say what the licence permits and which regulations apply, but not **who decides access or how** — which is the one of the three a downstream consumer most plausibly needs in order to know whether they can obtain the data at all.

## Why #506 left it out, and why that may be wrong

Adding a slot to `CoreDataset` widens the exchange surface, and #506 was already a schema change with a digest move. Leaving core alone kept the blast radius to `Dataset`. That was the cautious call in the moment, not a reasoned position about what belongs in core.

The reasoned position needs someone to answer: **is the core record for a machine deciding whether to ingest the data, or for one describing it once ingested?** Access governance matters enormously for the first and not at all for the second. The other two governance classes being present suggests the first, which argues for including it.

## The same gap, independently found

The VOICE_PEDIATRIC agent in the 2026-08-11 sweep reported:

> `related_datasets` does not exist in `CoreDataset` — the scope relation the manifest requires is expressible only in the full record. Core is in scope by `id`, but a core-only reader cannot see the distinct adult dataset.

Two different omissions, same shape: a fact the full record must state, that the exchange layer silently drops. And #403 was a third — data standards could not reach `CoreDistribution` until that PR added `conforms_to`.

Worth deciding as one question rather than three times separately: **what is `CoreDataset` for, and what does that imply about which full-record facts must project into it?** A per-slot answer each time is how the current asymmetry accumulated.

## Not blocking

#506 is an improvement on its own terms — `Dataset` had nowhere to put governance at all, and now it does. This is about the next step.

### LINKED ISSUE #518: check-schema-sync globs D4D_*.yaml, so a core-only edit demands a full-schema regen that produces only churn
Found while landing #510, which edits `D4D_Core.yaml` and nothing else.

`utils/check-schema-sync.sh` finds the newest module by globbing:

```bash
find "$MODULES_DIR" -name "D4D_*.yaml" -print0
```

`D4D_Core.yaml` matches that glob and is **not** imported by `data_sheets_schema.yaml` — it belongs to the core exchange schema (`data_sheets_schema_core.yaml` → `D4D_Core`). The full schema's `imports:` block lists 13 modules and `D4D_Core` is not among them.

So a core-only edit reports:

```
✗ Merged schema is out of date
  Newest source:  src/data_sheets_schema/schema/D4D_Core.yaml
  Fix: make full-schema
```

which is false. The full merged schema cannot be affected by a module it does not import.

## Why it is not cosmetic

Following the instruction produces a real commit with no semantic content: `make gen-project` re-emitted `project/owl/data_sheets_schema.owl.ttl` with **2226 changed lines** (1116 insertions, 1116 deletions — pure re-serialization ordering), plus two timestamp-only changes in `data_sheets_schema.jsonld` and `datamodel/data_sheets_schema.py`.

A churn commit is not harmless. It buries the actual change: the #510 diff is 4 files, and following `check-sync` would have made it 7 with 2200 lines of noise on top. This is the same failure mode as the `git add -A` sweep on #503's branch, arrived at by a different route — and here the tooling actively instructs it.

It also trains the reader to skip regenerated artifacts in review, which is where a real generated-artifact change would then hide.

## Fix

Derive the module list from the source schema's own `imports:` block instead of globbing:

```bash
sed -n '/^imports:/,/^[a-z#]/p' "$SOURCE_SCHEMA" \
  | sed -n 's/^[[:space:]]*-[[:space:]]*\(D4D_[A-Za-z_]*\)[[:space:]]*$/\1/p'
```

Yields the 13 imported modules and excludes `D4D_Core`. Cannot drift from what is actually imported: a module added to `imports:` is covered the day it is added, which the glob only achieved by accident of naming.

Verified both directions — touching `D4D_Ethics.yaml` (imported) still exits non-zero and names it; touching `D4D_Core.yaml` (not imported) no longer does.

## Not covered

`check-sync` is mtime-based throughout, so it reports a file touched-but-unchanged as out of date and would miss a content change that preserved mtime. Comparing content hashes would be the stronger check. Out of scope here — this fixes the wrong-input problem, not the weak-comparison one.

realmarcin:
Closes #510. Also closes #518, found while landing it.

## The decision

#510 asked a question rather than reporting a bug: **what is `CoreDataset` for?** Access governance matters enormously if core is for deciding whether to ingest a dataset, and not at all if it is for describing one already ingested.

This PR takes the first position, on the evidence of what core already carried:

| slot | the question it answers |
|---|---|
| `license_and_use_terms` | may I use it |
| `regulatory_restrictions` | what law applies |
| `conforms_to` on `CoreDistribution` (#403) | can I read it |

All ingest-decision facts. Access governance is the same kind and arguably the most decisive — a consumer who cannot find out who grants access cannot obtain the data at all.

**Two of three was worse than none.** A reader who finds licence terms and regulations present reasonably infers that governance was considered and found absent, rather than never expressible.

## What changed

`CoreDataset` gains `data_governance` (range `DataGovernance`) and `related_datasets` (range `DatasetRelationship`). Both ranges were already reachable — `D4D_Core` imports `D4D_Data_Governance` and `D4D_Composition` — so this is two slots, not a module.

`related_datasets` was the same gap found independently by the VOICE_PEDIATRIC agent in the 2026-08-11 sweep: it is the slot `source_manifest.yaml` names in `express_as`, so a core record could not state its own declared scope relation. Core was in scope by `id` while a core-only reader had no way to see the distinct adult dataset. Citing a related dataset there is also what keeps it out of `resources` and `distribution_formats`, where it would be absorbed into this dataset's own distribution (#441).

Both are **identity** slots, not projected: they state the same fact in either record. `resources` remains the only projected slot, and a test now pins that so a second one cannot arrive silently.

## Blast radius, measured

7 verdicts pin the core schema and went `STALE` on the hash change — the digest working exactly as #426 designed. All 7 revalidate unchanged, since widening a schema with optional slots is additive:

```
before: {'valid': 139, 'invalid': 19, 'stale': 7}
after:  {'valid': 146, 'invalid': 19}
```

Full suite: **1749 passed, 4 skipped**.

## Timing

Deliberate. The API sweep is stopped, and #517 established that a schema moving *during* an arm is the thing that must not happen — it turned part of a replicate-variance measurement into a schema effect. Between arms is the window for schema changes, and this is between arms.

Consequence to note: when #516 merges, its 10 rep2/rep3 records will go `STALE` against this schema until revalidated, for the same additive reason and with the same fix (`d4d runs validate --recheck`).

## #518, found doing this

`utils/check-schema-sync.sh` globbed `D4D_*.yaml` for the newest module. `D4D_Core.yaml` matches and is **not** imported by `data_sheets_schema.yaml`, so a core-only edit reported the full merged schema out of date. Obeying that instruction regenerated `project/owl/data_sheets_schema.owl.ttl` with 2226 changed lines of pure re-serialization ordering plus two timestamp-only diffs — a commit with no semantic content that would have buried this 5-file change under 2200 lines of noise.

The module list now derives from the source schema's own `imports:` block, so it cannot drift from what is actually imported. Verified both ways: touching `D4D_Ethics.yaml` still fails and names it; touching `D4D_Core.yaml` no longer does.

Not fixed: `check-sync` remains mtime-based throughout, so it flags touched-but-unchanged files and would miss a content change preserving mtime. Content hashing is the stronger check and is out of scope here — this fixes the wrong-input problem, not the weak-comparison one.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -203,7 +203,20 @@ gen-examples:
 .PHONY: full-schema
 full-schema: $(SOURCE_SCHEMA_ALL)
 
-$(SOURCE_SCHEMA_ALL):
+# Prerequisites are the source schema and every module it imports. Without
+# them this rule had none at all, so it fired only when the merged file was
+# missing and `make full-schema` was a silent no-op on an existing one — which
+# is why `regen-all` has to `rm -f` the target first. Combined with #518 that
+# made a closed loop: `check-sync` reported the merged schema stale and told
+# you to run `make full-schema`, which did nothing, so it stayed stale (#521).
+#
+# The module list is derived, not written out, so a module added to the source
+# schema's `imports:` is a prerequisite the day it is added.
+SCHEMA_IMPORTS = $(addprefix $(SOURCE_SCHEMA_DIR),$(addsuffix .yaml,$(shell \
+	sed -n '/^imports:/,/^[a-z\#]/p' $(SOURCE_SCHEMA_PATH) | \
+	sed -n 's/^[[:space:]]*-[[:space:]]*\(D4D_[A-Za-z_]*\)[[:space:]]*$$/\1/p')))
+
+$(SOURCE_SCHEMA_ALL): $(SOURCE_SCHEMA_PATH) $(SCHEMA_IMPORTS)
 	@echo "Generating D4D-Full schema with merged imports..."
 	$(RUN) gen-linkml -o $@ -f 'yaml' $(SOURCE_SCHEMA_PATH)
 	@echo '---' | cat - $@ > $@.tmp && mv $@.tmp $@
@@ -370,7 +383,16 @@ SSSOM_STRUCTURAL = data/semantic_exchange/d4d_rocrate_structural_mapping.sssom.t
 
 gen-core-schema: $(D4D_CORE_SCHEMA_ALL) ## Generate merged core exchange schema (data_sheets_schema_core_all.yaml)
 
-$(D4D_CORE_SCHEMA_ALL): $(D4D_CORE_SCHEMA) src/data_sheets_schema/schema/D4D_Core.yaml
+# Same derivation as SCHEMA_IMPORTS, over D4D_Core's own imports. Listing only
+# D4D_Core.yaml missed every module it imports: editing D4D_Base_import.yaml
+# left the merged core schema stale with make reporting nothing to be done
+# (#521). D4D_Core and the full schema import overlapping but not identical
+# module sets, so this is derived separately rather than shared.
+CORE_SCHEMA_IMPORTS = $(addprefix $(SOURCE_SCHEMA_DIR),$(addsuffix .yaml,$(shell \
+	sed -n '/^imports:/,/^[a-z\#]/p' $(SOURCE_SCHEMA_DIR)D4D_Core.yaml | \
+	sed -n 's/^[[:space:]]*-[[:space:]]*\(D4D_[A-Za-z_]*\)[[:space:]]*$$/\1/p')))
+
+$(D4D_CORE_SCHEMA_ALL): $(D4D_CORE_SCHEMA) $(SOURCE_SCHEMA_DIR)D4D_Core.yaml $(CORE_SCHEMA_IMPORTS)
 	@echo "Generating merged core exchange schema..."
 	$(RUN) gen-linkml -o $(D4D_CORE_SCHEMA_ALL) -f yaml $(D4D_CORE_SCHEMA)
 	@echo "✓ Core schema: $(D4D_CORE_SCHEMA_ALL)"
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
@@ -305,7 +305,7 @@ validation:
       md5: 19f4a17c60311d6bbc32fd06d266a76c
   schema:
     full_sha256: a392c073f2de00293702c7d31e6e1899f550d9b4bed2f01126630c25989da94b
-    core_sha256: 985460b61d58cf747fd67016bc0656fa204250fdd02ca69a8d638e45a22fcc6b
+    core_sha256: de0438d7df6b4b4e64abfde16e729531bc9c4fc72328872d84bc8b7dcad4db25
   recorded_by: d4d runs validate
 intermediates:
 - path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/intermediate/AI_READI_audit.json
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
@@ -304,8 +304,8 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_d4d_core.yaml
       md5: 19f4a17c60311d6bbc32fd06d266a76c
   schema:
-    full_sha256: a392c073f2de00293702c7d31e6e1899f550d9b4bed2f01126630c25989da94b
-    core_sha256: de0438d7df6b4b4e64abfde16e729531bc9c4fc72328872d84bc8b7dcad4db25
+    full_sha256: 060d9f542f817e291a5557e758715687b985a584df8d2fc74d2bfc99f857f9de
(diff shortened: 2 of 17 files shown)


realmarcin:
## Review

Read-only pass. Two findings; one dissolved on inspection.

### Confirmed — filed as #520

Both new slots are **identity** slots, and `d4d_pair_consistency` reads identity as *present in both or neither*. So every historical pair whose full record populates one while its core does not now reports `shared-slot-presence`: **88 occurrences across 70 pairs**, all `related_datasets`.

The error is true — those cores genuinely lack a fact their full records state, which is the gap #510 was filed about. Not a live gate either: the pair check runs at generation time, and `--sync-core` populates both slots automatically (verified below).

**The obvious fix is the wrong one**, which is why #520 exists: running `--sync-core` over the 70 would make a 2026-07-28 core record assert content no run of that date produced, break `validation.artifacts.*.md5` on every pair touched, and leave each `*_reconciliation.md` describing a record that no longer matches it. Same principle as `pre_registry` (#399) and the `canonical_superseded_by` tombstone (#516): annotate history, never rewrite it to satisfy a rule that postdates it.

### Verified end-to-end

The concern that mattered — does this add slots nothing will ever fill?

```
after --sync-core, core has:
   data_governance      True
   related_datasets     True
pair errors after sync: 0
```

Phase 4 of the generation playbook runs `--sync-core`, so future pairs are consistent by construction and the new slots populate without any prompt change.

A `CoreDataset` record using both also validates cleanly against the regenerated core schema.

### Dissolved

`committee_contact` (range `Person`) rejects an inline object while its siblings `committee_members` (`inlined_as_list`) and `accountable_organization` (`inlined`) accept one. That looked like an inconsistency in the class #503 added, and a candidate fix was to inline it.

Checking the records first showed the opposite: the 2026-08-11 runs populate it correctly as a reference — `mailto:jillianparker@health.ucsd.edu`, `d4d:CHORUS-person-ciera-mccrary` — and inlining it would have **invalidated all four**. The by-reference typing is right; a contact is a pointer and members are a list. No change.

Worth noting separately that `d4d:CHORUS-person-ciera-mccrary` resolves to no `Person` in the record, so the reference dangles. That is the #402 identifier family, already tracked, not something this PR introduces.

### Not reviewed

Whether `DataGovernance`'s slot set is the right one — that was settled in #503 and this PR only changes where the class can be attached.

realmarcin:
## Scope grew: #499 and #521 landed here too

Continuing the loop, since #499 is a defect in the same pair model #510 extends and could not be tested without #521.

### #499 — a record can now say what it is

`conforms_to_schema` and `conforms_to_class` describe the record they sit in, so a full/core pair must state them **differently** — `Dataset` in one, `CoreDataset` in the other. Both were strict-identity slots, which made them unrepresentable: any honest value failed the pair check, and `--sync-core` copied full's value into core and wrote a false claim about what the core record instantiates. The AI_READI run of the 2026-08-11 sweep hit this in phase 4 and resolved it by omitting the slot from both records — the only state that neither fails the gate nor asserts something false.

Now marked with a `d4d:perRecord` annotation on the slots, read into a third category beside identity and projected. The annotation lives in the schema, not a name list in the checker, so the category is visible where the slots are defined — a hardcoded list is what #431 removed elsewhere.

`conforms_to` stays strict-identity and tests hold that from both sides. A standard the *data* follows is the same in both records; an exemption leaking to it would stop checking a real fact, silently and for ever.

### #521 — two make rules had wrong prerequisites

`$(SOURCE_SCHEMA_ALL)` had **no prerequisites at all**, so `make full-schema` rebuilt only when the file was missing. That is why `regen-all` has to `rm -f` it first — the workaround was in the repo and the cause was not. Combined with #518 it closed a loop: check-sync reported the merged schema stale and told you to run `make full-schema`, which did nothing.

`gen-core-schema` was the same shape, listing `D4D_Core.yaml` but not the twelve modules it imports. Observed directly: the annotation added for #499 was in the module and absent from the merged core schema, so `load_pair_schema` read it as unset.

Both rules now derive their module list from the corresponding `imports:` block — the third instance of the same defect class, all now fixed the same way.

## Two self-inflicted defects caught in review

**The annotation did nothing at first.** `slot.annotations` is a `jsonasobj2.JsonObj` with no `.get`; the accessor called it inside `except AttributeError: return False`, so every slot read as unmarked and `per_record_slots` was empty. The schema advertised a guarantee nothing enforced. A test now covers the accessor directly — a check that cannot fail is worse than no check.

**The exemption was silent.** The report listed identity and projected slots and dropped the third category, so a passing pair read as fully verified when two slots had been exempted. `per_record_slots` now appears in `to_dict()` and the CLI summary, on pass as well as failure.

## State

**1763 passed, 4 skipped.** The 7 verdicts pinning the schema went `STALE` on each hash move and all 7 revalidate unchanged; corpus at 0 stale.

Closes #510, #499, #518, #521.

PR #523, opened 2026-08-12
Accept cross-slot repetition, and measure it (#501)

### LINKED ISSUE #501: The same fact is restated across five slots (Safe Harbor in AI-READI), and no rule says which slot should own it
From Camille Nebeker's review of the Ethics and Privacy content of AI-READI and CHORUS:

> Redundancy in content vs potential new content, some fields have redundant content (eg multiple mentions of SAFE Harbor). Partly this is expected based on schema design and overlapping questions and fields and structure. However, we also want to prioritize new informative content vs redundant content. So this could be more about picking the best content for the field given the content already generated for other fields.

## Confirmed, and it is project-specific rather than systemic

Measured on the fresh `2026-08-11_claude-opus-5-claudecode-generic_rep1` records — the same Safe Harbor fact appears **6 times across 5 distinct top-level slots** in AI_READI, in near-identical wording:

| slot | text |
|---|---|
| `preprocessing_strategies` | "The public release is stripped of protected health information as defined by the HIPAA Privacy Rule via the Sa…" |
| `human_subject_research` | "HIPAA — the public release is stripped of protected health…" |
| `participant_privacy` | "No direct identifiers are collected… The public release is stripped of protected hea…" |
| `regulatory_restrictions` | "HIPAA — the public release is stripped of protected health information…" |
| `is_deidentified` | "HIPAA Safe Harbor for the public release, as stated in the Nature Metabolism comment…" |

**CHORUS has zero** — its bundle does not discuss Safe Harbor. So this is a response to what the evidence contains, not a habit applied uniformly, which is worth knowing before any instruction change.

## Why the current rules produce it

The uniform decision rules tell a run to populate a slot where the bundle supports it, and to prefer omission over inference. Nothing tells it that a fact already stated in a sibling slot is *less* valuable there. Each of the five placements is individually defensible; the aggregate is repetitive.

#385 catalogued a related mode — "(c) redundant — restates a sibling", 11 in CHORUS and 33 in AI-READI — and was closed once prose moved out of `notes` into `description`. That fixed *where* prose goes, not *whether the same prose goes several places*. This is the residue.

## The tension to resolve before changing anything

A reviewer reading only `is_deidentified` should learn that the release is Safe Harbor de-identified. Suppressing the repeat makes each slot incomplete on its own; keeping it makes the document repetitive read end to end. Those are genuinely different products — a queryable record versus a readable one — and the rule should say which it optimises for.

Note also that one of the five is *not* redundant: `is_deidentified` records that **two source statements differ** and keeps both. That is the disagreement-preservation rule working, and a naive de-duplication pass would delete it.

## Suggested

Not a find-and-collapse. Either a decision rule stating that a fact belongs in its most specific applicable slot and is referenced rather than restated elsewhere, or an explicit decision that per-slot completeness wins and the repetition is accepted. Either way it changes instructions, so it is a pre-sweep decision.

realmarcin:
Closes #501.

## The question was a decision, not a bug

Camille Nebeker's review found the same Safe Harbor fact in five slots of AI-READI. #501 framed the tension honestly: a reviewer reading only `is_deidentified` should learn the release is Safe Harbor de-identified, but the document then reads repetitively end to end. Those are different products and the rule has to say which it optimises for.

## Measured before deciding

Across the whole canonical set (the 2026-08-11 v1 arm), sentence-level, content-word Jaccard ≥ 0.6:

| project | sentences | prose restatements | rate | structural |
|---|---|---|---|---|
| AI_READI | 451 | 39 | 8.6% | 3 |
| CHORUS | 192 | 17 | 8.9% | 0 |
| CM4AI | 324 | 19 | 5.9% | 30 |
| VOICE | 364 | 18 | 4.9% | 9 |
| VOICE_PEDIATRIC | 255 | 23 | 9.0% | 2 |
| **total** | **1586** | **116** | **7.3%** | 44 |

Between 4.9% and 9.0% on all five projects — uniform behaviour, so whichever way the rule went it would have moved every record.

## The decision: per-slot completeness wins

The core record is read one slot at a time — by a consumer deciding whether to ingest, by a mapping into RO-Crate or DCAT, by a reviewer checking one question. A reader who opens `is_deidentified` alone must find an answer there. Cross-referencing gives them a pointer instead, and a machine consuming one slot cannot follow a pointer at all.

The repetition is the price, paid knowingly. **No prompt change**, so no condition is re-baselined and no pin rotates.

## Two things measuring changed

**They are paraphrases, not copies.** "removed from", "stripped from", "excludes" — three slots, three wordings, one fact. An exact-match pass over the same records reports 8 restatements in AI_READI and misses the Safe Harbor family entirely. The alternative rule could therefore never have been a mechanical de-duplication pass; it would have had to be an instruction applied by every future run.

**Structural repetition is not redundancy.** A URL beside the format it belongs to, a nested `CoreDataset` under `resources` repeating its parent's title. 44 pairs, almost all CM4AI Dataverse URLs. Counted and reported separately — folding them into the headline would have overstated it by about a third.

## The alternative was worse than #501 flagged

The issue warned that a naive dedup would delete `is_deidentified`, which preserves that FAIRhub gives `deIdentType: "NoDeIdentification"` while the Nature Metabolism comment states Safe Harbor. Measuring shows **`participant_privacy` carries that same disagreement**. The contradiction is itself restated across two slots — and it is the most valuable content in either.

Any "state it once" rule needs an explicit exemption for preserved disagreements, or it deletes the best sentence in the record to save the third-best.

## What is built

`d4d runs redundancy` — reported, never fatal. The number exists to be watched, not met: at 7.3% the behaviour is as decided; a later arm at 30% would mean something changed that nobody chose.

`tests/test_redundancy.py` holds the exit code at zero. Turning this into a gate would reverse the decision silently, which is what that test prevents.

`notes/decision_501_cross_slot_repetition.md` records the decision, the numbers it was made on, and what it does not settle — so this is not re-filed in three months as an unnoticed defect.

**1795 passed, 4 skipped.**

### DIFF
--- a/notes/decision_501_cross_slot_repetition.md
+++ b/notes/decision_501_cross_slot_repetition.md
@@ -33,6 +33,30 @@ word Jaccard ≥ 0.6:
 Between 4.9% and 9.0% across five projects, so this is uniform behaviour rather
 than one project's quirk, and any rule would have moved every record.
 
+### The headline depends on the threshold, and the threshold is a judgement
+
+Same corpus, same code, varying only what counts as "the same statement":
+
+| threshold | restatements | rate |
+|---|---|---|
+| 0.5 | 191 | 12.0% |
+| **0.6** | **116** | **7.3%** |
+| 0.7 | 79 | 5.0% |
+| 0.8 | 60 | 3.8% |
+| 0.9 | 44 | 2.8% |
+| 1.0 (exact) | 34 | 2.1% |
+
+**7.3% is not a property of the corpus; it is a property of the corpus at 0.6.**
+0.6 was chosen by inspection against the Safe Harbor family, which sits at
+0.6–0.8 while unrelated sentences in the same record sit below 0.35 — so it is
+the point that separates the known case from the known non-cases, on one
+family, in one record.
+
+The decision does not rest on the exact figure: repetition is either the right
+product or it is not, and it is uniform at every threshold. But any future
+comparison must use the same threshold, which is why `d4d runs redundancy`
+prints it beside the rate.
+
 Two corrections came out of measuring rather than assuming:
 
 - **They are paraphrases, not copies.** "removed from", "stripped from",
--- a/notes/decision_501_cross_slot_repetition.md
+++ b/notes/decision_501_cross_slot_repetition.md
@@ -1,0 +1,96 @@
+# Decision: cross-slot repetition is accepted (#501)
+
+**Decided 2026-08-12. Records generated after this date repeat facts across
+slots on purpose.**
+
+## The question
+
+Camille Nebeker's review of the AI-READI and CHORUS ethics and privacy content:
+
+> Redundancy in content vs potential new content, some fields have redundant
+> content (eg multiple mentions of SAFE Harbor). … we also want to prioritize
+> new informative content vs redundant content.
+
+Confirmed. In the canonical AI_READI record the HIPAA Safe Harbor fact appears
+in five slots — `human_subject_research`, `participant_privacy`,
+`preprocessing_strategies`, `regulatory_restrictions`, `is_deidentified` —
+across eight values.
+
+## What was measured first
+
+Across the whole canonical set (the 2026-08-11 v1 arm), sentence-level, content-
+word Jaccard ≥ 0.6:
+
+| project | sentences | prose restatements | rate | structural |
+|---|---|---|---|---|
+| AI_READI | 451 | 39 | 8.6% | 3 |
+| CHORUS | 192 | 17 | 8.9% | 0 |
+| CM4AI | 324 | 19 | 5.9% | 30 |
+| VOICE | 364 | 18 | 4.9% | 9 |
+| VOICE_PEDIATRIC | 255 | 23 | 9.0% | 2 |
+| **total** | **1586** | **116** | **7.3%** | 44 |
+
+Between 4.9% and 9.0% across five projects, so this is uniform behaviour rather
+than one project's quirk, and any rule would have moved every record.
+
+Two corrections came out of measuring rather than assuming:
+
+- **They are paraphrases, not copies.** "removed from", "stripped from",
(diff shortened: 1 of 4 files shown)


realmarcin:
## Review

One finding, addressed in `a10e91b9`.

**The headline was far more threshold-dependent than "7.3%" suggests.** Same corpus, same code, varying only what counts as the same statement:

| threshold | restatements | rate |
|---|---|---|
| 0.5 | 191 | 12.0% |
| **0.6** | **116** | **7.3%** |
| 0.7 | 79 | 5.0% |
| 0.8 | 60 | 3.8% |
| 0.9 | 44 | 2.8% |
| 1.0 (exact) | 34 | 2.1% |

A six-fold range. So 7.3% is not a property of the corpus — it is a property of the corpus **at 0.6**, and 0.6 was chosen by inspection against one fact family in one record. Quoting it bare is the shape of every stale-number defect in this repo: a figure that reads as measured and is actually parameterised.

The command now prints the threshold beside the rate, and prints the *override* rather than the default when one is passed — reporting the default while having used an override would be worse than reporting nothing. The decision note carries the sensitivity table. Three tests hold it, including one asserting the parameter still changes the count, so the guard cannot go vacuous.

**The decision does not move.** Repetition is either the right product or it is not, and it is uniform across all five projects at every threshold. What changes is that any future comparison has to state the threshold to be a comparison at all.

### Checked and not filed

- **Is the 4.9–9.0% spread meaningful?** No. CM4AI is lowest on prose restatement and highest on structural (30 Dataverse URLs), so its low prose figure reflects a record dominated by component-crate URLs rather than less repetitive prose. Recorded in the note rather than filed, since the decision does not turn on the ranking.
- **Does `summarize` double-count a fact in three slots?** Yes — three slots produce three pairs, not one fact. That is intended for a *rate* over sentence pairs, and the per-project `slots_involved` list is what shows the fact-level picture. Documented, not changed.

### Not reviewed

Within-slot repetition. The comparison is deliberately cross-slot only, and the note says so under "what this does not settle".

Issue #520, opened 2026-08-12
Adding related_datasets to core makes 70 historical pairs report shared-slot-presence — and back-filling them would be wrong


realmarcin:
Consequence of #519, found reviewing it. Filed rather than fixed, because the obvious fix is the wrong one.

## What changes

`d4d_pair_consistency` treats a schema-identical slot as **present in both or neither**:

```
schema-identical slot is present only in full; it must be present in both or neither
```

Before #519, `related_datasets` and `data_governance` existed only on `Dataset`, so they were full-only slots and no pair could disagree about them. After #519 both are identity slots, and every historical pair whose full record populates one while its core does not now reports an error.

Measured across `data/d4d_concatenated`: **88 occurrences across 70 pairs**, all `related_datasets`. (`data_governance` is new enough that no full record on `main` carries it; the 10 records in #516 will add to this once merged.)

## Why this is not a regression

The error is **true**. Those core records genuinely lack a fact their full records state, and that was exactly the gap #510 was filed about. Making the gap visible is the point. Before, the exchange layer silently dropped the relation and nothing said so.

Nor is it a live gate: the pair check runs at generation time (`.claude/commands/d4d-full-core.md` phase 4), not over the historical corpus, and `--sync-core` populates both slots automatically — verified. Future runs are consistent by construction.

## Why back-filling would be wrong

The mechanical fix is to run `--sync-core` across all 70 pairs. **Do not.** Three reasons, each sufficient:

1. **It makes a historical record assert content it never generated.** A core record from 2026-07-28 would gain a `related_datasets` block that no run of that date produced. The record would then state a fact whose provenance is a migration script, indistinguishable from one the model actually produced.
2. **It invalidates the evidence attached to it.** Rewriting the bytes breaks `validation.artifacts.*.md5`, so every touched pair's verdict goes `STALE` — and unlike #519's own 7, these would be stale because *we* changed them, not because the schema moved.
3. **The reconciliation report would no longer describe the record it sits beside.** Each pair has a `*_reconciliation.md` recording what phase 4 actually decided. A back-filled slot appears in neither.

This is the same principle as `pre_registry` in #399 and the `canonical_superseded_by` tombstone in #516: historical evidence is annotated, never rewritten to match a rule that postdates it.

## Proposed

- Leave the 70 pairs alone.
- If the presence error becomes noisy, scope it by run date the way `requires_request` already does — silent before the slot existed in core, enforced after.
[...]

Issue #433, opened 2026-08-10
Existing verdicts keep no schema pin, and nothing reports how many


realmarcin:
Noticed verifying #426.

`validation_block()` now records the schema a verdict was reached against, and `validation_status()` returns STALE when it moves. But `d4d runs validate` skips a run whose status is already VALID:

```python
if not recheck and status not in (UNVERIFIED, STALE):
    continue
```

which is correct — re-validating 100+ records costs seconds each and the whole reason the verdict is cached is to avoid it. The consequence is that **every verdict currently in the corpus keeps no schema pin**, because none of them will be re-validated by an ordinary `d4d runs validate`.

That is deliberate and safe: an unpinned verdict is left alone, on the same reasoning as the live-provenance cutoff. Absent is not stale, and failing 55 records to enforce a rule that postdates them would discard evidence rather than check it.

But it does mean the protection only reaches records validated from now on, and there is currently no way to tell a pinned verdict from an unpinned one except by reading the YAML.

## Options

1. **Leave it.** New runs get pinned verdicts; old ones stay as they are. Zero work, and the corpus converges as runs are re-validated for other reasons.
2. **`d4d runs validate --recheck` over the corpus once**, so every verdict gains a pin against today's schema. Honest only if the schema has not changed since those verdicts were reached — and for the 2026-08-07 sweep it has not, which is checkable. For older records it has, and re-validating would produce a *new* verdict rather than pinning the old one, which is a different act and should not be disguised as bookkeeping.
3. **Report the count.** Have `d4d runs check` say how many verdicts carry no schema pin, so the gap is a number someone can watch shrink rather than a property nobody can see.

Recommend (3) now and (2) selectively, per label, where the schema demonstrably has not moved. (1) is acceptable but leaves the corpus in a state where "VALID" means two different things depending on when it was written.

## Related

#426 for the pin itself. #423 for the half already handled — a verdict *carried forward* across a re-record has compared the schema block since then.

PR #524, opened 2026-08-12
The American-English rule, and playbook drift detection (#502, #525)

### LINKED ISSUE #502: Generated prose uses British spellings, including inside an identifier — but the bundles do too, so this is not a find-and-replace

> Standardize on American English in the D4D generated docs ie no 'programme'.

## Confirmed

Counts in the fresh `2026-08-11_claude-opus-5-claudecode-generic_rep1` records:

```
programme     10        analyse         9        characteris    15
organisation   4        centre          3        recognis        3
organis       10        licence        46        prioritis       1
```

## The part that makes this non-trivial

**Some of these are quoted source text, and Americanising a quote would corrupt evidence.** The bundles themselves contain British spellings — `licence` 13 times, `programme` 6, `organisation` 1 — so a find-and-replace over the records would silently rewrite what a source said, which is the one thing the provenance guard exists to prevent.

The generated prose is separable, though, and one line shows both at once:

```
programme funding as "the Bridge2AI Program (NIH Common Fund; …"
```

The record's own prose says *programme* while the text it is quoting says *Program*. That is unambiguously a generation artifact.

## Worse than prose: it reaches identifiers

```
- id: aireadi:external-training-programme
```

A British spelling inside an `id` is structural, not stylistic — it is a token other records and mappings may key on, and it is not something a later copy-edit can safely change.

## Suggested

A uniform decision rule that generated prose uses American English, stated alongside the existing rules rather than as a post-hoc pass, plus an explicit carve-out that quoted source text, proper nouns and identifiers taken from sources keep their original spelling. A checker could then flag British spellings *outside* quoted spans; flagging them everywhere would produce mostly false positives given `licence` appears 13 times in the bundles.

Changes instructions, so it is a pre-sweep decision.

realmarcin:
Closes #502.

## The rule was easy; the placement was the decision

Camille Nebeker: *"Standardize on American English in the D4D generated docs ie no 'programme'."*

Generated prose — and identifiers the run **mints** — use American English. Three carve-outs, none optional:

- **Quoted source text keeps its original spelling.** Americanising a quotation corrupts evidence, which is exactly what the provenance guard exists to prevent.
- **Proper nouns keep theirs.** "Wellcome Trust Sanger Centre" is a name, not prose.
- **Identifiers copied from a source keep theirs.** An id from a crate or a DOI is a token.

## Why the playbook and not the condition prompts

Editing `d4d_generic_arm_prompt.md` rotates v1's pin — and that pin is what the fifteen records of the 2026-08-11 canonical arm hashed. Rotating it moves all fifteen from `canonical` to `superseded`. Reported rather than failed, but a real degradation of the baseline #516 was generated to establish.

Verified after the change:

```
d4d api prompts check --strict   exit 0
d4d runs check --strict          exit 0
canonical_prompt_status          canonical x5
```

House style is also orthogonal to what the conditions differ on. v1 against v3 measures structured-slot population and defect counts; spelling does not enter it, so the rule does not belong in the axis being compared.

## Measured

```
1345 occurrences in generated prose
  43 in text quoted from a bundle (left alone)
   8 inside an `id`
```

All 8 identifiers are **minted**, not copied — `d4d:VOICE-purpose-programme`, `chorus:related-bridge2ai-programme`, `#instance-programme-aggregate` — so every one violates the rule as stated.

The 43 quoted occurrences are why this cannot be a find-and-replace. The bundles contain `licence` 13 times and `programme` 6, and one line carries both at once: the record's own prose says *programme* while the text it quotes says *Program*.

## Two gaps, named rather than hidden

**The API path does not read the playbook**, so API-generated records are unaffected. Deliberate while §6 is pending — a prompt edit mid-comparison is the change #517 exists to warn about. The rule should move into the condition prompts at the next version rotation, once the arms are complete.

**`d4d evaluate spelling` cannot tell a minted id from a copied one**, and the rule turns on that distinction. It has not mattered because all 8 are minted, but a real identifier containing `licence` would be a false positive. One case is already mixed — a real DOI with a minted fragment — where the finding is still correct because the offending span is the part the run composed.

## Not done

**Existing records are not rewritten.** The 1345 occurrences stand. Editing them would make a record state prose no run produced — the same objection as #520. New runs comply; old records are evidence of what was generated.

Also replaces "the last rule is the load-bearing one" with the rule named explicitly, since inserting a rule silently repointed that sentence at a different one.

**1805 passed, 4 skipped.**

### DIFF
--- a/.claude/commands/d4d-full-core.md
+++ b/.claude/commands/d4d-full-core.md
@@ -216,9 +216,27 @@ to every project. Enforce them whether or not a prompt file was used to launch:
 - There is no target slot count, no expected density, and no expected
   relationship to any other arm or project. Apply your own judgment about what
   the evidence supports.
+- **Write generated prose in American English** — `program`, `organization`,
+  `analyze`, `license`, `center`, `labeling`, `enrollment`. This is house style
+  for the text *you* compose, and it applies to identifiers you mint as well as
+  to descriptions.
 
-The last rule is the load-bearing one: it is what makes a slot count an
-observation rather than a target.
+  Three carve-outs, and they are not optional:
+
+  - **Quoted source text keeps its original spelling.** Changing a quotation to
+    match house style corrupts evidence, which is the one thing the provenance
+    guard exists to prevent. The bundles contain `licence` 13 times and
+    `programme` 6.
+  - **Proper nouns keep their spelling** — "Wellcome Trust Sanger Centre",
+    "Medical Research Council Programme Grant". A name is not prose.
+  - **Identifiers copied from a source keep the source's spelling.** An id you
+    take from a crate or a DOI is a token, not a sentence. Only ids *you* mint
+    follow house style.
+
+The rule about there being no target slot count is the load-bearing one: it is
+what makes a slot count an observation rather than a target. Named rather than
+referred to by position, so inserting a rule cannot silently point this sentence
+at a different one.
 
 ### Recording the condition
 
--- a/notes/decision_502_american_english.md
+++ b/notes/decision_502_american_english.md
@@ -1,0 +1,81 @@
+# Decision: generated prose is American English (#502)
+
+**Decided 2026-08-12.** From Camille Nebeker's review: *"Standardize on American
+English in the D4D generated docs ie no 'programme'."*
+
+## The rule
+
+Generated prose uses American English — `program`, `organization`, `analyze`,
+`license`, `center`, `labeling`, `enrollment` — and so do identifiers the run
+mints. Three carve-outs:
+
+1. **Quoted source text keeps its original spelling.** Americanising a quote
+   corrupts evidence, which is the one thing the provenance guard exists to
+   prevent.
+2. **Proper nouns keep their spelling.** "Wellcome Trust Sanger Centre" is a
+   name, not prose.
+3. **Identifiers copied from a source keep the source's spelling.** An id taken
+   from a crate or a DOI is a token. Only minted ids follow house style.
+
+## Where it lives, and why not in the condition prompts
+
+**In the playbook** (`.claude/commands/d4d-full-core.md`, uniform decision
+rules), not in `d4d_generic_arm_prompt*.md`.
+
+Editing a condition prompt rotates its pin, and v1's pin is what the fifteen
+records of the 2026-08-11 canonical arm hashed. Rotating it would move all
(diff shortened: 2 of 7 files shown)


realmarcin:
## Review — one finding, and this PR is what makes it bite

**#525: playbooks are hashed per record and never checked for drift.** Fixed here in `dfa473b2`.

Every agentic record hashes the playbooks it read. Nothing ever compared those hashes to the files again. `d4d runs check` reports bundle drift (#452), render-gate status (#419), canonical prompt pins (#432), verdict schema pins (#433), reasoning-effort basis (#448) and schema straddle (#517) — and said nothing about the files the decision rules actually live in.

That is not a small omission. The playbook is where "prefer omission over inference", the disagreement rule, one-referent-per-`Dataset` and now the American-English rule are written. **A playbook edit is a change to the method exactly as a prompt edit is** — and prompt edits are guarded three ways while playbook edits were guarded by none.

Adding the rule in this PR left 17 records attesting to bytes no longer at that path, with `--strict` exiting 0 and nothing said. Now:

```
ⓘ  17 record(s) read a playbook whose bytes have since changed
     d4d-full-core.md (c1e03a0b -> a6b1d76a)    17
```

Both hashes are named, so a reader can tell *which* edit rather than merely that there was one. `absent` is a separate outcome from `drifted`, because a rename is a different diagnosis from an edit — the failure #431 was filed for.

Reported, never fatal, for the same reason as bundle drift: a drifted record is still valid evidence of what was generated, it just cannot be re-derived from the files it names. Playbooks are also *meant* to evolve, so a gate would turn every improvement to the method into a corpus-wide failure. That is also why this is drift detection rather than prompt-style pinning — a condition must mean one fixed thing; shared method must not.

## A defect I shipped and caught

The first version referenced `hashlib` without importing it — `runs.py` imports it inside two sibling functions rather than at module scope. `d4d runs check --strict` went from exit 0 to a `NameError` traceback. Caught by running the command rather than trusting the unit tests, which had not yet been written. Worth noting because the failure mode was a *crash in the gate itself*, which is the worst place for one.

## Verified

- Agentic records really do hash `.claude/commands/d4d-full-core.md` — checked against a real provenance record, not assumed.
- The rule's placement disturbs nothing: `prompts check --strict` 0, `runs check --strict` 0, all five projects still `canonical`.
- **1815 passed, 4 skipped.**

Closes #502 and #525.

PR #526, opened 2026-08-12
Give retention incentives their own slots (#504)

### LINKED ISSUE #504: Compensation vocabulary is payment-first; retention incentives have no slot and should not be inferred from a payment

> For compensation, the goal should be to represent as incentives to prevent dropout of participants. Deemphasize compensation itself and replace with incentives.

## Current state

The vocabulary is compensation-first. `D4D_Human.yaml`:

```yaml
  # class description
  Information about compensation or incentives provided to human research participants.
      compensation_provided:
        slot_uri: d4d:compensationProvided
      compensation_type:
        description: >
          What type of compensation was provided (e.g., monetary payment,
          gift cards, course credit, other incentives)?
```

So "incentives" appears only as a trailing alternative in both the class description and the `compensation_type` examples. The slot names, the `slot_uri`, and the leading framing are all compensation.

## What the review is asking for

Not a rename for its own sake. The distinction is substantive in human-subjects research: **compensation** is payment for time and burden, while **retention incentives** are structured to prevent attrition — and the ethical reading of the two differs, which is exactly why an Ethics and Privacy reviewer raised it. A datasheet that reports "$40/$80 gift cards" as compensation says something different from one that reports the same figures as a retention design.

## Care needed on three points

1. **Renaming a slot is not free.** `compensation_provided` and `compensation_type` are in the merged schema and in generated records; a rename changes the schema digest and orphans existing values unless aliased.
2. **The evidence usually says "compensation".** Sources describe payments; a record must still represent what the source states rather than re-framing it (the `target` failure the fitness axis names). So the schema can offer an incentive framing, but a run must not *assert* retention intent the bundle does not state.
3. **These may be two fields, not one renamed field.** "What was provided" and "what it was designed to do" are different claims, and collapsing them would lose the first.

## Suggested

Widen rather than rename: keep `compensation_provided` for what the source states, and add a slot for retention/incentive structure where the source describes one — with descriptions that say plainly which is which and that neither should be inferred from the other. That is additive, so it does not orphan values, and it keeps the record honest when a source only reports a payment.

Changes both schema and instructions, so it is a pre-sweep decision.

realmarcin:
Closes #504.

## The ask

Camille Nebeker: *"For compensation, the goal should be to represent as incentives to prevent dropout of participants. Deemphasize compensation itself and replace with incentives."*

## Widened, not renamed

This is what #504 itself recommends, and what the evidence requires.

Renaming `compensation_*` would move the schema digest and orphan values already in the corpus. Worse, it would make a record report a payment as a retention design when the source says only that a payment was made — the `target` failure the fitness axis names. **The sources overwhelmingly describe compensation**, and a record must represent what a source states.

So two new slots sit beside the four existing ones:

| slot | claim |
|---|---|
| `compensation_type` / `_amount` | what participants received |
| `compensation_rationale` | why that amount was appropriate |
| **`retention_incentives`** | how the structure was designed to keep them enrolled |
| **`retention_incentive_rationale`** | the stated reason for that design |

Each new slot has its own `slot_uri`, so the distinction survives into consumers that see a URI and not a slot name.

## Why this is substantive, not terminological

The ethical reading differs, which is why an Ethics and Privacy reviewer raised it. **"$40 then $80" reported as compensation says something different from the same figures reported as a retention design** — the second is a claim about study design and about who can afford to stay enrolled.

Structure and rationale are separate slots for the same reason `compensation_rationale` is separate from `compensation_amount`: what was done and why it was done are different claims, and sources routinely give one without the other.

## The load-bearing instruction

The descriptions state plainly that **an escalating payment schedule is not by itself evidence that retention was its purpose** — report the schedule under the compensation slots and leave the retention slots absent.

That second half matters. A rule that only said "do not infer" would leave a run with a true fact and nowhere to put it, which is how content gets dropped rather than corrected. A test holds both halves.

Nothing is required, so a source reporting only a payment still produces a valid record.

## Blast radius

17 verdicts pinned the schema and went `STALE` — the digest working as designed. All 17 revalidate unchanged, since adding optional slots is additive.

```
before: {'valid': 139, 'invalid': 19, 'stale': 17}
after:  {'valid': 156, 'invalid': 19}
```

The pair model is unaffected: `human_subject_research` remains a strict-identity slot and the identity count holds at 78.

**1808 passed, 4 skipped.** The 19 pre-existing lint problems in `D4D_Human.yaml` are unchanged and none mention these slots — verified against `main`.

## Timing

Between arms, deliberately. The API sweep is stopped and #517 established that a schema move during an arm is what must not happen.

### DIFF
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
@@ -304,8 +304,8 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_d4d_core.yaml
       md5: 19f4a17c60311d6bbc32fd06d266a76c
   schema:
-    full_sha256: f223e4bf11641452c706a28dd6dce3423d6c24d1f6e749a58bfa0726b0e7dc87
-    core_sha256: 3a660827ea17d6a8972899aeeb3843b0051ec5b0c0ef12456f1c0735c13ff43d
+    full_sha256: da5ad57b5db9e0b832ad2ae4c048550601b213b69f5d786b768adcbb7287cc88
+    core_sha256: 9b3fbe3bb8fdc01792ee41dd9db75cbef3c5ea40fdc1d90d5fce037951ad154f
   recorded_by: d4d runs validate
 intermediates:
 - path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/intermediate/AI_READI_audit.json
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
@@ -304,8 +304,8 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_d4d_core.yaml
       md5: 19f4a17c60311d6bbc32fd06d266a76c
   schema:
-    full_sha256: 060d9f542f817e291a5557e758715687b985a584df8d2fc74d2bfc99f857f9de
-    core_sha256: 4b725d5da4694a66326acf3f488c8c8c0042b2c5d5d81019ac5525f54bf511c0
+    full_sha256: f223e4bf11641452c706a28dd6dce3423d6c24d1f6e749a58bfa0726b0e7dc87
+    core_sha256: 3a660827ea17d6a8972899aeeb3843b0051ec5b0c0ef12456f1c0735c13ff43d
   recorded_by: d4d runs validate
 intermediates:
 - path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/intermediate/AI_READI_audit.json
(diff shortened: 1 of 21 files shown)


realmarcin:
## Review — one finding, fixed in `34a077af`

**The word "retention" collides with a slot that already exists**, and it would have bitten on the first run.

`retention_limit` means how long the *data* is kept. I checked what the bundles actually say before assuming the new slots would be populated sensibly:

| project | lines matching retention/attrition/drop-out |
|---|---|
| AI_READI | 8 |
| CHORUS | 0 |
| CM4AI | 32 |
| VOICE | 4 |
| VOICE_PEDIATRIC | 3 |

Reading them, **nearly every one is the data sense** — "limits on the retention of the data", `"bioSpecRetention"`, "records retention requirements", and in VOICE a citation about survey attrition plus "drop-out rates" as a feasibility measure. **None describes an incentive structure.**

Two consequences:

1. **A run meeting `retention_incentives` with those sources in context has an obvious wrong place to put them.** #403 already recorded what happens next: when two slots are near-synonyms and neither is constrained, records pick one.
2. **These slots will be empty in every current record, and that is correct.** Worth saying loudly, because otherwise a reviewer reads five empty slots as a generation failure and someone fills them from data-retention text — the exact error this change exists to prevent.

The description now says which sense is meant, **names `retention_limit`** as the slot for the other, and states that an empty slot is the correct outcome for these sources. Naming rather than implying: a reader resolving an ambiguity needs to be told where the fact *does* belong.

Four tests hold it, including one asserting `retention_limit` is named in the text.

## Verified, not assumed

- The 19 pre-existing lint problems in `D4D_Human.yaml` are byte-identical on `main` and none mention these slots — checked by stashing.
- The pair model is untouched: `human_subject_research` stays strict-identity, identity count holds at 78.
- 17 verdicts went `STALE` on each of the two schema moves; all 17 revalidate unchanged both times. Corpus at 156 valid, 0 stale.
- **1812 passed, 4 skipped.**

## Still true and worth stating plainly

This adds expressive capacity that **no current source exercises**. That is the right outcome for an ask about how a record *should* represent something, and it is not evidence the change works — the first source describing a retention design will be.

PR #527, opened 2026-08-12
Tell the fitness judge the ranges of nested attributes (#486)

### LINKED ISSUE #486: slot_spec renders nested attribute names without ranges, so the fitness judge cannot assess range conformance below the top level
Found while fixing #456, checking whether re-ranging `unit` would move the schema digest and invalidate cached judgements. It does not — and the reason is a gap worth naming.

## What the judge is told about a nested attribute

`slot_spec()` renders a slot's specification for the fitness judge. For a slot whose range is a class, it renders the class's obligations — but only as **names**:

```
Field: `variables`
Declared range: VariableMetadata (list — many values expected)
`VariableMetadata` requires: variable_name
`VariableMetadata` also accepts: categories, data_type, derivation, description, …, unit, used_software
```

No ranges. The judge learns that `unit` is an accepted field of `VariableMetadata` and nothing about what a `unit` value should look like. The schema digest — which the fitness and sub-type caches key on (#465) — renders the same names-only view, so the two are at least consistent: the cache cannot go stale on information the judge never had.

## Why it matters

`FORM_SUBTYPE_SYSTEM` asks the judge to distinguish two defects. One is judgeable from names alone:

> **hollow_object** — the content sits in a free-text field while the structured fields the range declares (`name`, `id`, `start_date`, …) are unused.

The other is not:

> **other** — a bare string where a list of objects is declared, **a wrong range**, a scalar where a structure belongs.

For a *top-level* slot, `slot_spec` supplies the range and the judgement is well founded. For an attribute nested one level down, it does not, so "wrong range" cannot be assessed and a value like `unit: mg/dL` under a `uriorcurie` declaration is invisible to the fitness axis. `d4d runs identifiers` found those 148 values precisely because it reads the schema directly rather than asking a judge.

So the `form` axis is measuring hollowness reliably and range conformance only at the top level. That is a narrower claim than "form failures", and the v1-vs-v3 comparison rests on it.

## Not urgent, and specifically not a cache problem

Nothing is stale: judge and key see the same thing. The fix is to render nested attribute ranges in `slot_spec`, which would deepen the judgement — and would legitimately invalidate the cached labels, since the judge would then be answering a better-informed question. That is a deliberate re-measurement, not a bug fix, and should be decided against the analysis plan rather than slipped in before a comparison.

Related: #465 (the key now carries the schema), #456 (the range change that surfaced this).

realmarcin:
Closes #486.

## The gap

`FORM_SUBTYPE_SYSTEM` asks the judge to separate two defects:

> **hollow_object** — the content sits in a free-text field while the structured fields the range declares are unused.
>
> **other** — a bare string where a list of objects is declared, **a wrong range**, a scalar where a structure belongs.

The first is judgeable from names alone. The second is not — and `slot_spec` rendered nested attributes as names only:

```
`VariableMetadata` also accepts: categories, data_type, …, unit, used_software
```

So below the top level, the form axis was measuring **hollowness reliably and range conformance not at all**. That is a narrower claim than "form failures", and the v1-vs-v3 comparison rests on it.

## What changed

`NestedClass` carries each attribute's declared range, rendered in both the digest and `slot_spec`:

```
`VariableMetadata` attribute ranges: data_type → VariableTypeEnum, id → uriorcurie,
  is_identifier → boolean, maximum_value → float, used_software → Software[]
A value of the wrong kind for its declared range is a form failure even when it reads well.
```

The rule is stated rather than left to be inferred — naming a range without saying what to do with it is what produced the gap.

**Only non-string ranges.** `string` is the schema default, so naming it costs prompt length and says nothing; `uriorcurie` is where a plausible-looking value is the wrong kind. **67 nested classes have a `uriorcurie` attribute**, almost all `id`, so this is largely the #402/#457 family becoming visible to the judge — where it matters most, since a bare token validates exactly as cleanly as a ROR IRI.

Multivalued ranges are marked (`Person` vs `Person[]`), because a judge told the first about the second cannot assess cardinality.

## The cache move is the point, not a side effect

Digest: **32599 → 36635 chars (+12.4%)**, fingerprint **`663e67a9` → `d2e9eaa0`**.

The fitness cache keys on that fingerprint, so every cached label is correctly skipped. The judge now answers a better-informed question and old answers are not comparable to new ones.

**Had `slot_spec` changed without the digest render changing, the key would have held while the question moved** — every cached label answering a worse-informed question with nothing to say so. That is the #465 failure exactly. A test strips the ranges from a built digest and asserts the fingerprint responds, so the two cannot silently diverge.

## Timing

#486 asks that this be "decided against the analysis plan rather than slipped in before a comparison". The API sweep is stopped, so it lands between arms — the same window #517 argues for. A future arm records digest `d2e9eaa0` and is cleanly separated from the 2026-08-11 arm at `659aae67`.

## Notes

- **`unit` correctly does not appear.** It is `string` since #456, even though it is the attribute #486 was filed about. The general capability is what this delivers, not that one case.
- **No schema YAML changed**, so no verdict is stale and nothing needs revalidating.
- **1838 passed, 4 skipped.**

### DIFF
--- a/src/data_sheets_schema/evidence_score.py
+++ b/src/data_sheets_schema/evidence_score.py
@@ -731,6 +731,23 @@ def slot_spec(slot: str, class_name: str = "Dataset",
         if nested.optional:
             lines.append(f"`{sd.range}` also accepts: "
                          f"{', '.join(nested.optional)}")
+        # The ranges of those attributes, without which "a wrong range" — one
+        # of the two defects FORM_SUBTYPE_SYSTEM asks the judge to separate —
+        # cannot be assessed below the top level (#486).
+        #
+        # Only non-string ranges: `string` is the default, so naming it adds
+        # prompt length and no information, while `uriorcurie` is where a
+        # plausible-looking value is the wrong kind. `unit: mg/dL` reads as a
+        # perfectly good value until you know `unit` is declared `uriorcurie`.
+        typed = {k: v for k, v in sorted(nested.ranges.items())
+                 if v not in ("string", "string[]")}
+        if typed:
+            lines.append(
+                f"`{sd.range}` attribute ranges: "
+                + ", ".join(f"{k} → {v}" for k, v in typed.items()))
+            lines.append(
+                "A value of the wrong kind for its declared range is a form "
+                "failure even when it reads well.")
     return "\n".join(lines)
 
 
--- a/src/data_sheets_schema/evidence_score.py
+++ b/src/data_sheets_schema/evidence_score.py
@@ -739,15 +739,19 @@ def slot_spec(slot: str, class_name: str = "Dataset",
         # prompt length and no information, while `uriorcurie` is where a
         # plausible-looking value is the wrong kind. `unit: mg/dL` reads as a
         # perfectly good value until you know `unit` is declared `uriorcurie`.
-        typed = {k: v for k, v in sorted(nested.ranges.items())
-                 if v not in ("string", "string[]")}
+        # Universal ranges once, then the ones specific to this class — the
+        # same split the digest render uses, from the same constants, so the
+        # judge's question and the cache key cannot drift apart.
+        lines.append(f"On every `{sd.range}` object: "
+                     f"{schema_digest.UNIVERSAL_RANGES}")
+        typed = schema_digest.shown_ranges(nested)
         if typed:
             lines.append(
                 f"`{sd.range}` attribute ranges: "
                 + ", ".join(f"{k} → {v}" for k, v in typed.items()))
-            lines.append(
-                "A value of the wrong kind for its declared range is a form "
-                "failure even when it reads well.")
+        lines.append(
+            "A value of the wrong kind for its declared range is a form "
+            "failure even when it reads well.")
     return "\n".join(lines)
 
 
--- a/src/data_sheets_schema/schema_digest.py
+++ b/src/data_sheets_schema/schema_digest.py
@@ -91,6 +91,14 @@ class NestedClass:
     # model cannot choose from a list it has never been shown.
     enums: dict[str, list[str]] = field(default_factory=dict)
     enums_truncated: dict[str, int] = field(default_factory=dict)
+    # Declared range of each nested attribute (#486). Names alone let a judge
+    # assess *hollowness* — content in prose while structured keys sit empty —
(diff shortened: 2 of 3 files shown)


realmarcin:
## Review — two findings, both fixed in `d0c2c528`

### 1. The first version violated an existing decision and blew the size budget

Two guards already in the repo caught it, which is the system working:

```
test_universal_attributes_are_not_repeated_on_every_class  FAILED
test_the_listing_did_not_double_the_digest                 FAILED (40090 not < 40000)
```

Putting every nested attribute's range on a per-class line reintroduced `id` and `used_software` on all 67 classes — precisely what the first guard forbids as bloat — and pushed the digest 90 characters past a budget the same file defends.

**Raising the ceiling would have been the wrong fix.** `id` is `uriorcurie` on all 67 nested classes and `used_software` is `Software[]` on 64, so saying it *once* is strictly better than saying it 67 times: identical information, no bloat. The digest lands at **36800** instead of 40090 and still carries what the judge lacked.

### 2. The consistency test gave false assurance — and the two renderings really had diverged

`test_the_fingerprint_responds_to_the_nested_ranges` strips *every* range, which moves the fingerprint whatever the filters are. So it passed while `slot_spec` and the digest render showed different sets:

```
digest shows: is_identifier, is_sensitive, maximum_value, minimum_value, precision
spec shows:   data_type, id, is_identifier, is_sensitive, maximum_value,
              minimum_value, precision, used_software
DIVERGENCE:   data_type, id, used_software
```

A change to any of those three would have moved the judge's question while the cache key held — **the #465 failure this PR exists to avoid**, reintroduced by the fix for it.

Replaced with set equality per nested class, and both renderings now derive from one `UNIVERSAL_RANGES` constant and one `shown_ranges` function rather than two filters that happened to agree.

### A hazard the debugging surfaced — filed as #528

`build()` memoises and returns the **shared** `ClassDigest`. The stripping test mutated it in place, corrupting the cached digest for every later test in the process; two of them then failed against empty data and looked exactly like real defects in the code under test.

Deep-copied here. Filed separately because nothing but convention stops production code doing the same, and `build()` feeds the generation prompt, `slot_spec` and the cache key.

## State

Digest **32599 → 36800** (+12.9%), fingerprint **`663e67a9` → `e164a95f`**. The fitness cache keys on that, so cached labels are correctly skipped — the judge now answers a better-informed question.

**1842 passed, 4 skipped.**

PR #529, opened 2026-08-13
Give data standards a vocabulary that reaches core (#403)

### LINKED ISSUE #403: No vocabulary for data standards: conforms_to is free text, conforms_to_schema is dead, and standards cannot project into core

realmarcin:
Closes #403.

## Two of the three items were already done

Confirmed rather than redone:

- **Item 1** — `conforms_to_schema` / `conforms_to_class` now have distinct, enforced meanings via #499: they describe *the record*, so they are per-record slots exempt from strict identity. They are no longer near-synonyms of `conforms_to`.
- **Item 3** — `CoreDistribution.conforms_to` already exists.

## What remained: the vocabulary

Every record of `2026-08-07_…-generic-v3_rep1` put a data standard into free-text `conforms_to`, in four shapes for one concept — long-name-plus-parenthetical, bare name, name-with-version, raw URL. None resolvable to a term, so the corpus cannot be queried for "what standards does this follow?".

`DataStandardEnum` covers the eight the corpus names (DICOM, BIDS, OMOP_CDM, WFDB, OPEN_MHEALTH, ESDS, CDS, RO_CRATE) plus FHIR and `OTHER`. `conforms_to_standard` carries it on `Dataset`, `CoreDataset` and `CoreDistribution`.

**That closes the projection loss.** A free string had no home in core, so the most interoperability-relevant fact about a distribution was dropped. AI_READI's phase 4 handled it correctly — it left `format` unset rather than coerce a wrong enum member — but the loss was created by the schema and took a paragraph of reconciliation prose to explain.

## Not merged into FormatEnum

The issue warns against this and a test now enforces it. CSV says how bytes are laid out; OMOP CDM says what the columns *mean*. Absorbing DICOM would conflate the two and make `format` unanswerable. The two enums are asserted disjoint.

## Added, not replaced

`conforms_to` stays free text recording what the sources say. The new slot records which standard that is. **Both are populated for the same standard** — verbatim and normalized — the ordinary label-plus-code pattern, and consistent with what #501 decided about repetition.

A term-only slot would drop "ASCII File Format Guidelines for Earth Science Data" rather than record it. `OTHER` exists so an uncovered standard is kept rather than forced onto a near-neighbour: **reporting BIDS because a dataset is neuroimaging, when the sources say otherwise, is an invention**, and the enum says so where a run will read it.

## The grounding gap is named, not papered over

**No `meaning:` is asserted on any value**, and the enum's own description explains why. Several of these standards have no stable ontology term this repository can verify, and a CURIE that resolves to nothing is exactly what `d4d runs identifiers` exists to find. `B2AI_STANDARD` is declared and is the natural home once the registry carries them; until then this is a controlled vocabulary without ontology grounding, and says so.

A test asserts every value's `meaning` is `None`, so grounding cannot arrive by accident — only by decision.

## State

Identity slots **78 → 79**. Identity is right here: which standard the *data* follows is a fact about the data and is the same in either record — unlike `conforms_to_class`, which describes the record.

17 verdicts went `STALE` on the schema move; all 17 revalidate unchanged. Corpus at 156 valid, 0 stale.

**1858 passed, 4 skipped.**

### DIFF
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
@@ -304,8 +304,8 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_d4d_core.yaml
       md5: 19f4a17c60311d6bbc32fd06d266a76c
   schema:
-    full_sha256: da5ad57b5db9e0b832ad2ae4c048550601b213b69f5d786b768adcbb7287cc88
-    core_sha256: 9b3fbe3bb8fdc01792ee41dd9db75cbef3c5ea40fdc1d90d5fce037951ad154f
+    full_sha256: 27992f75f8ee0c9905e9e4947d51298f7184b4bcd126cc45491e6616af127eb2
+    core_sha256: 3ed2ec45e974af811362988ac084e73a4ec9e026f60b4294db1cf66676244f92
   recorded_by: d4d runs validate
 intermediates:
 - path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/intermediate/AI_READI_audit.json
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/CHORUS_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/CHORUS_provenance.yaml
@@ -266,8 +266,8 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/CHORUS_d4d_core.yaml
       md5: 57d52aa67e15baf249e791cd1ce32481
   schema:
-    full_sha256: da5ad57b5db9e0b832ad2ae4c048550601b213b69f5d786b768adcbb7287cc88
-    core_sha256: 9b3fbe3bb8fdc01792ee41dd9db75cbef3c5ea40fdc1d90d5fce037951ad154f
+    full_sha256: 27992f75f8ee0c9905e9e4947d51298f7184b4bcd126cc45491e6616af127eb2
+    core_sha256: 3ed2ec45e974af811362988ac084e73a4ec9e026f60b4294db1cf66676244f92
   recorded_by: d4d runs validate
 intermediates:
 - path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/intermediate/CHORUS_audit.json
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-claudecode-generic_rep1/AI_READI_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-claudecode-generic_rep1/AI_READI_provenance.yaml
@@ -130,6 +130,6 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-claudecode-generic_rep1/AI_READI_d4d_core.yaml
       md5: 7fa94e1ccb88dcc9d7090ce6ff0b49d5
   schema:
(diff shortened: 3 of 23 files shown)


realmarcin:
## Review

The failure mode worth chasing here is a vocabulary value that came from model memory rather than from a source. **A permissible value is a claim, and once merged it is unfalsifiable** — the schema is not evidence, and nothing else would ever check it.

`CDS` was the one to check: "Clinical Dataset Structure v0.1.1" reaches the enum via a *generated record*, not directly from a document.

All nine are attested in the bundles:

| value | lines |
|---|---|
| RO-Crate | 206 |
| OMOP | 45 |
| DICOM | 23 |
| **CDS** | **19** |
| WFDB | 11 |
| FHIR | 8 |
| BIDS | 6 |
| Open mHealth | 5 |
| Earth Science Data | 2 |

Nothing invented. Pinned as a test, plus a second asserting every enum value has an attestation entry — so a value added later cannot skip the check by not being listed.

## Also verified

- **Both the judge and the generation prompt see the permitted values.** `slot_spec` renders `Permitted values: DICOM, BIDS, …` and the digest renders `accepts only:`. A vocabulary a run is never shown is one it cannot use — the failure #403 records for `DatasetRelationshipTypeEnum`, where runs filled a controlled slot with DataCite spellings and inventions because they had never seen the list.
- Digest 36800 → 37680, well inside the 40k budget.
- The two enums are asserted disjoint, so `FormatEnum` cannot quietly absorb a data standard later.

## Checked and not filed

`conforms_to_standard` renders on three nested classes as well as the top level, repeating the value list. That is the existing nested-enum behaviour — a controlled vocabulary on a nested slot has to be shown there or nowhere — not something this PR introduces.

**1858 passed, 4 skipped.**

Issue #457, opened 2026-08-11
Constrain uriorcurie with a pattern, and migrate the 7,415 values that would then fail


realmarcin:
The remaining half of #402. The audit landed in #455 (`48b26f4a`) and reports; nothing yet enforces.

## The change

`uriorcurie` declares no `pattern`, so LinkML renders it `{"type": ["string","null"]}` and every string is legal. Adding a pattern makes `funder_nih` a validation failure, which is correct and is why this is a migration rather than a flag flip.

## What it would break, today

`d4d runs identifiers` measures it exactly: **7,415 of 24,477 identifiers (30%) across 157 of 341 records**.

| slot | unresolvable |
|---|---|
| `id` | 7124 |
| `unit` | 148 |
| `publisher` | 68 |
| `latest_version_doi` | 31 |
| `data_substrate` | 22 |
| `data_topic` | 22 |

Of the 7,415: 3,891 are bare tokens and 3,524 are CURIEs on prefixes the schema does not declare. Those need different remedies — declare the prefix versus rewrite the value — and the second group may be much cheaper than it looks if a few prefixes account for most of it.

## Sequencing

1. **Settle #456 first.** `unit` is 148 of 148 unresolvable because it holds `%` and `years` — its range is wrong, and a migration that treated those as mis-entered identifiers would repair the wrong thing. Same question for `data_substrate` and `data_topic`.
2. **Count the undeclared prefixes.** If 3,524 values use a handful of prefixes, declaring them in the schema converts them to valid CURIEs with no record edits at all.
3. **Then decide the pattern and the fate of the remainder.** Options for existing records: rewrite in place (they are committed and hashed, so provenance implications — cf. #452), leave them failing and gate only new records, or version the constraint.

Note the interaction with #452: 64 records already reference input bundles whose bytes have changed. Rewriting record content would move their artifact hashes too, so whatever is decided there should be decided alongside this.

## Why not just add the pattern

`--strict` on `d4d runs identifiers` already gives anyone who wants the gate today a way to have it, without making 157 records fail their schema validation before there is a plan for them.

realmarcin:
**Precondition partly settled by #456** (merged as 23b3d8c5).

`unit` is re-ranged to `string`, so its 148 values leave this migration's scope entirely:

```
unresolvable: 7415 → 7267    (exactly the 148 unit values)
slots with range uriorcurie: data_substrate, data_topic, id, latest_version_doi, publisher
```

[...]

Issue #531, opened 2026-08-13
One namespace, five prefix spellings: declaring them all would enshrine the inconsistency


realmarcin:
Found doing #457's step 2, alongside #530.

#457's step 2 asks whether declaring a few prefixes converts many values cheaply. For the VOICE namespace the answer is that there is no *one* prefix to declare:

| prefix | count |
|---|---|
| `b2ai-voice` | 665 |
| `b2aivoice` | 172 |
| `voice` | 159 |
| `b2ai` | 66 |
| `b2ai.voice.org` | 32 |
| **total** | **1,094** |

Five spellings, one namespace, produced across runs by models improvising a CURIE prefix that the schema never gave them. There is also a second family that is not a namespace at all — `org` (58), `organization` (56), `creator` (36) — where the "prefix" names the *type of thing* rather than the namespace it belongs to.

## Why this is not just untidy

**Declaring all five would make every value valid and the corpus no more queryable than before.** `b2ai-voice:x` and `voice:x` would expand to different IRIs, so the same entity would carry several identities and nothing would join them. A CURIE's purpose is that two records naming one thing produce one IRI.

It is the same shape as #403 and #402: the schema offered no term, so records invented one, and they invented inconsistently. #403's remedy was a controlled vocabulary; the analogue here is declaring **one** prefix per namespace and treating the others as values to migrate.

## Sequencing against #457

This has to be settled before the migration, not during it:

- **Which spelling is canonical?** `b2ai-voice` is most used but that is not an argument; the registry or the project should decide.
- **The type-prefixes (`org:`, `creator:`) are a different remedy.** They are not mis-spelled namespaces, they are mis-modelled: an `Organization.id` of `org:ucsd` says nothing a ROR would not say better. #378 is about resolving organization identifiers against the B2AI registry and is the right home for those 150.
- **`urn:cm4ai:org:ucsd` is the same mistake in URN clothing** (#530) — well-formed, unregistered, and naming a type in the middle.

## Not proposed

Rewriting existing records. #457 already flags that they are committed and hashed, and #520 established the principle: historical records are annotated, not rewritten to satisfy a rule that postdates them. What is needed first is a decision about the canonical prefix so that *new* records stop adding to the pile.

realmarcin:
Researched the registry rather than picking a spelling, per the instruction to be consistent across all GCs and D4Ds. **The answer is that none of the five is correct**, and the registry has already settled this.

[...]

Issue #378, opened 2026-08-06
Resolve organization identifiers from the B2AI Standards registry (B2AI_ORG CURIEs); contribute missing universities upstream


realmarcin:
The B2AI Standards Registry maintains curated organizations with stable CURIEs (`B2AI_ORG:N`), plus `ror_id`, `wikidata_id` and URL per entry: https://github.com/bridge2ai/b2ai-standards-registry/blob/main/src/data/Organization.yaml

This is the right source for real organization identifiers — better than the name-as-id workaround repair used, and aligned with the enrichment-vs-extraction principle: a deterministic post-generation lookup, provenance-marked, never model-invented.

**Coverage census** (127 registry orgs vs 173 organization mentions across the four valid 1m-sweep records): **19 matched (11%)**. The registry covers standards bodies, agencies and consortium entities; it contains **no universities**, while our top mentions are UCSD (15×), Stanford (12×), Johns Hopkins (11×), UW (9×), WashU, Utah, UMass Lowell, OHSU… — the Bridge2AI-performing institutions themselves.

**Plan:**
1. **Local (this repo)**: vendor a pinned snapshot of the registry (hash + fetch date), a resolver matching on `name`/`description`/`ror_id` with strict normalization (no fuzzy guessing), and a `d4d enrich orgs` pass that fills `Organization.id` with the `B2AI_ORG` CURIE where a match exists, recording the enrichment in provenance. Schema 2.0.0's optional-id Organization is exactly shaped for this.
2. **Upstream (b2ai-standards-registry)**: the ~30 unmatched institutions with their RORs arguably belong in the registry — Bridge2AI's own performing sites. Contributing them is a cross-repo decision for a human; census list attached here when wanted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

realmarcin:
Measured, now that #535 declared `B2AI_ORG` and made the registry comparison possible.

## What the corpus actually holds

2,508 values across the three `Organization`-ranged slots (`affiliations`, `affiliation`, `accountable_organization`), 109 distinct organizations:

| id form | values | example |
|---|---|---|
| minted CURIE / URN | **1,308** | `urn:d4d:org:university-of-virginia` |
| ROR | 439 | `https://ror.org/0153tk833` |
| minted project IRI | 390 | `https://b2ai-voice.org/organization/usf` |
| no id | 348 | |

**Only 18% carry a ROR. 68% carry an identifier that resolves to nothing.**

[...]

PR #532, opened 2026-08-13
A URI without :// is still a URI (#530)

### LINKED ISSUE #530: The identifier audit counts scheme-based URIs (urn:, ark:, doi:) as undeclared CURIEs — 1,067 values misclassified
Found doing #457's step 2 (counting undeclared prefixes before deciding the migration).

`d4d runs identifiers` classifies a value as an absolute IRI only when it contains `://`, and as a CURIE otherwise. That misses **URI schemes whose syntax has no authority component**, which are URIs all the same:

| prefix | count | example | what it is |
|---|---|---|---|
| `urn` | 757 | `urn:cm4ai:org:ucsd` | RFC 8141 URN |
| `ark` | 286 | `ark:59853/b2ai-voice-dataset-feature-ppgs` | ARK identifier |
| `doi` | 24 | `doi:10.18130/V3/HIGT4C` | DOI in `doi:` form |
| | **1,067** | | |

All three are reported as "CURIE on an undeclared prefix" and counted toward the 8,317 unresolvable. **They are not CURIEs and declaring a prefix for them would be wrong** — `urn` is not a namespace this schema can bind to a base IRI, it is a scheme.

## Why it matters now

#457 sizes its migration on these counts. 1,067 of 3,817 undeclared-CURIE values — **28% of that bucket** — need no remedy at all beyond the classifier telling the truth. That materially changes the decision about what to rewrite.

It also changes the shape of the remedy: the issue's step 2 asks whether "declaring the prefix" converts values cheaply. For these, there is no prefix to declare, and any attempt to declare one would produce a binding that expands to nonsense.

## Correctness caveat, stated rather than hidden

Being *syntactically* a URI is not the same as resolving. `urn:cm4ai:org:ucsd` is a well-formed URN whose NID `cm4ai` is not IANA-registered, and `ark:59853/...` depends on a NAAN that may or may not be assigned. So these should move from "undeclared CURIE" to a category that says **well-formed URI, resolution not established** — not to the `uri` bucket alongside a resolvable `https://ror.org/...`, which would overstate their standing.

That is a third category the audit does not currently have, and inventing it is the substance of this fix.

## Suggested

1. Recognise the registered no-authority schemes (`urn:`, `doi:`, `ark:`, `mailto:`, `isbn:`, `info:`) as URIs rather than CURIEs.
2. Report them as a distinct outcome — well-formed, resolution unverified — so the headline "resolves to nothing" stops covering values that are correctly formed.
3. Re-run the counts and amend #457's sizing before any migration is planned against them.

## Separate finding, filed as its own issue

The genuinely-CURIE remainder has five spellings of one namespace — `b2ai-voice` (665), `b2aivoice` (172), `voice` (159), `b2ai` (66), `b2ai.voice.org` (32). Declaring prefixes for all five would enshrine that inconsistency rather than resolve it.

realmarcin:
Closes #530. Unblocks #457's sizing.

## The defect

`classify` tested for `://` and called everything else a CURIE. Schemes with no authority component have no `//`:

| prefix | count | example |
|---|---|---|
| `urn` | 757 | `urn:cm4ai:org:ucsd` |
| `ark` | 286 | `ark:59853/b2ai-voice-dataset-feature-ppgs` |
| `doi` | 24 | `doi:10.18130/V3/HIGT4C` |
| | **1,067** | |

All were reported as "CURIE on an undeclared prefix" and counted unresolvable.

**They are not CURIEs.** `urn` is a scheme, not a namespace this schema could bind to a base IRI — so "declare the prefix", the cheap remedy #457's step 2 goes looking for, does not apply to any of them, and declaring one would produce a meaningless expansion.

## Why a third category rather than moving them to `uri`

**Nor are they resolvable.** `urn:cm4ai:org:ucsd` is a well-formed URN whose NID is not IANA-registered; `ark:59853/…` depends on a NAAN that may not be assigned. Filing them beside a resolvable `https://ror.org/…` would overstate their standing as badly as calling them CURIEs understates it.

`URI_UNVERIFIED` says what is true: **well-formed under its scheme, resolution not established.**

## Order matters, and is pinned

The grammars overlap — `urn:cm4ai:org:ucsd` matches `prefix:reference` perfectly well. Scheme detection runs first, and a test asserts that a scheme which is *also* a declared prefix still reads as a URI, so classification cannot flip on which test ran.

## Effect

```
before   9923 uri                          8317 unresolvable (30%) / 171 records
after    9923 uri + 1067 unverified        7250 unresolvable (26%) / 144 records
```

That is the number #457 should be re-sized against, and it matches the estimate already posted there.

The exemption does not leak: `funder_nih` is still `bare_token`, `aireadi:public-dataset` is still `curie_undeclared`, and `URI_UNVERIFIED` is deliberately absent from `UNRESOLVABLE`.

**1869 passed, 4 skipped.**

### DIFF
--- a/src/data_sheets_schema/cli/runs.py
+++ b/src/data_sheets_schema/cli/runs.py
@@ -116,6 +116,12 @@ def identifiers_cmd(label, method, output, show, strict):
     click.echo(f"   slots with range uriorcurie: {', '.join(slots)}")
     c = report["counts"]
     click.echo(f"   {c[ident.URI]:>6}  absolute IRI")
+    # Named separately, never folded into either neighbour (#530). Folded into
+    # `uri` it would claim resolvability these do not have; left among the
+    # undeclared CURIEs it inflated the unresolvable headline by 1,067 values
+    # that need no remedy at all.
+    click.echo(f"   {c[ident.URI_UNVERIFIED]:>6}  URI on a no-authority scheme "
+               "(urn:, ark:, doi:) — well-formed, resolution not established")
     click.echo(f"   {c[ident.CURIE_DECLARED]:>6}  CURIE on a declared prefix")
     click.echo(f"   {c[ident.CURIE_UNDECLARED]:>6}  CURIE on an undeclared prefix")
     click.echo(f"   {c[ident.BARE]:>6}  bare token — neither IRI nor CURIE")
--- a/src/data_sheets_schema/identifiers.py
+++ b/src/data_sheets_schema/identifiers.py
@@ -55,8 +55,20 @@ _CURIE = re.compile(r"^([A-Za-z_][A-Za-z0-9._\-]*):(\S+)$")
 #: 1,067 corpus values sit here — `urn:` 757, `ark:` 286, `doi:` 24 — and every
 #: one was counted as an undeclared CURIE, inflating the migration #457 is sized
 #: against by 28% of that bucket.
+#:
+#: **Kept deliberately short, and to registered schemes only.** Every name here
+#: exempts a value from the audit, so a name that is *not* a scheme silently
+#: excuses a real defect. `isbn`, `issn` and `uuid` were dropped for exactly
+#: this reason: they are URN *namespaces* (`urn:isbn:…`), not schemes, and a
+#: record writing `uuid:abc` would be minting a prefix, not citing one.
+#:
+#: `file` is the sharpest case and is **excluded**. It is a registered scheme,
+#: but its 22 corpus occurrences are `file:torchaudio_spectrograms_parquet` —
+#: a minted type-prefix in the same family as `org:`, `creator:` and
+#: `software:` (#531), not `file:///path`. Adding it would erase 22 genuine
+#: findings to accommodate a name collision.
 NO_AUTHORITY_SCHEMES = frozenset({
-    "urn", "doi", "ark", "mailto", "isbn", "issn", "info", "tel", "uuid",
+    "urn", "doi", "ark", "mailto", "tel", "info",
 })
 
 URI = "uri"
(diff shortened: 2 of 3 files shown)


realmarcin:
## Review — one finding, fixed in the last commit

I checked what remained in the undeclared-CURIE bucket rather than trusting the scheme list I had just written, and found the list was too permissive in one direction and contained three names that are not schemes at all.

**`file` is the sharpest case, and it stays excluded.** It *is* a registered scheme, so a careless reading would add it. But its 22 corpus occurrences are:

```
file:torchaudio_spectrograms_parquet
file:sparc_ema_parquet
```

— a minted type-prefix in the same family as `org:` (58), `organization:` (56), `creator:` (36) and `software:` (30), all of which #531 covers. Not `file:///path`. **Exempting it would have erased 22 genuine findings to accommodate a name collision.**

**Dropped `isbn`, `issn`, `uuid`.** They are URN *namespaces* — `urn:isbn:…` — not schemes. A record writing `uuid:abc` is minting a prefix, not citing one. None appears in the corpus, so removing them changes no count and closes a hole before anything falls into it.

Counts are unchanged: 1,067 unverified, 7,250 unresolvable.

Three tests hold this, plus a cap on the list's length — **a growing exemption list is how an audit stops auditing**, and that is the failure mode worth guarding rather than the arithmetic.

## What this leaves for #457

The remaining 2,750 undeclared CURIEs are now cleanly two problems, neither of which is "add a pattern":

- **1,094** are five spellings of one namespace (#531)
- **~180** are type-prefixes that should be real identifiers (#531, #378)

Plus 4,500 bare tokens, which remain the largest group and the one with no cheap route.

**1872 passed, 4 skipped.**

PR #533, opened 2026-08-13
Hand out a copy of the cached digest, not the cache (#528)

### LINKED ISSUE #528: schema_digest.build returns the shared cached object, so a caller mutating it corrupts every later build
Found while writing tests for #486.

`build()` memoises into `_BUILD_CACHE` and returns the cached `ClassDigest` **by reference**. Any caller that mutates the returned object — or its `nested` entries, or their `ranges`/`enums` dicts — silently changes what every subsequent `build()` for that class returns, for the life of the process.

## How it showed up

A test built the digest, stripped `ranges` from each nested class to check the fingerprint responded, and passed. Two later tests in the same file then measured an **empty** digest and failed:

```
AssertionError: 0 not greater than 50
AssertionError: 0 not greater than 20 : too few nested classes exercised
```

The failures were real assertions failing against corrupted data, which is the dangerous shape: they looked like genuine defects in the code under test, and the first instinct is to "fix" the code.

## Why it matters beyond tests

`digest_text` → `render(build(...))` feeds:

- the generation prompt (the digest a run consumes),
- `slot_spec`, which is what the fitness judge reads,
- `schema_digest.fingerprint(...)`, which keys the fitness and sub-type caches.

A mutation anywhere in a long-lived process — a batch sweep, a notebook, a CLI command that builds twice — would change the prompt or the cache key for everything after it, with nothing recording that it happened. Nothing does this today; the exposure is that nothing prevents it either, and the failure would be silent and hard to attribute.

## Options

1. **Return a deep copy from `build()`.** Simplest and safest. Costs a copy per call; the digest is a few hundred small dataclasses, and `build` is already the expensive part it exists to avoid repeating.
2. **Freeze the dataclasses** (`frozen=True` on `SlotDigest`/`NestedClass`/`ClassDigest`). Catches mutation at the point it happens rather than downstream, but `ranges`/`enums` are dicts and would still be mutable — so it is a partial guard that reads as a total one.
3. **Document it and leave it.** The status quo, now written down.

Recommend (1). (2) is misleading unless the dicts become `MappingProxyType` too, which is a larger change for a hazard that is currently theoretical outside tests.

## Workaround in place

`tests/test_nested_ranges.py` deep-copies before mutating, with a comment saying why.

realmarcin:
Closes #528.

## The defect

`build()` memoised into `_BUILD_CACHE` and returned the stored `ClassDigest` **by reference**. Any caller mutating it — or its `nested` entries, or their `ranges`/`enums` dicts — silently changed what every later `build()` for that class returned, for the life of the process.

Found when a test stripped `ranges` to check the fingerprint responded, and two later tests in the same file measured an empty digest and failed. **Those failures read as real defects in the code under test**, which is the dangerous shape — the first instinct is to fix code that was correct.

## Why it matters outside tests

`build` feeds:

- the generation prompt (the digest a run consumes),
- `slot_spec`, which is what the fitness judge reads,
- `schema_digest.fingerprint(...)`, which keys the fitness and sub-type caches.

A mutation anywhere in a long-lived process — a batch sweep, a notebook, a CLI command that builds twice — would change the prompt or the cache key for everything after it, with nothing recording that it happened.

## Deep, not shallow

Replacing a dict is caught by a shallow copy; mutating one in place is not. A test covers both, because the second is the likelier accident.

## Cost, measured

```
cached build (with copy):   1.62 ms
cold build:               638 ms
copy costs 0.25% of a cold build
```

`SchemaView` construction and `class_induced_slots` are the expensive part the memo exists to avoid, and they are untouched.

## Two existing tests changed, deliberately

`test_repeated_builds_return_the_same_object` asserted `assertIs(a, b)`. That was a **proxy** for "it was memoised" and is no longer a valid one — identity was the means, not re-reading the schema is the end. It now asserts cost directly.

`test_different_classes_are_cached_separately` used `assertIsNot`, which would now pass **whatever the cache did**, since every call returns a distinct object. It compares rendered content instead. That one is worth noting: my change would otherwise have quietly turned a real test into a tautology.

The deep-copy workaround added to the #486 test is removed, and left mutating in place on purpose — if this isolation ever regresses, those tests fail, which is how it was found.

**1879 passed, 4 skipped.**

### DIFF
--- a/src/data_sheets_schema/schema_digest.py
+++ b/src/data_sheets_schema/schema_digest.py
@@ -19,6 +19,7 @@ truncated because their tail is usually curation notes rather than instruction.
 
 from __future__ import annotations
 
+import copy
 import hashlib
 from dataclasses import dataclass, field
 from pathlib import Path
@@ -161,11 +162,32 @@ _BUILD_CACHE: dict[tuple[str, str], "ClassDigest"] = {}
 
 
 def build(class_name: str, schema_path: Path | None = None) -> ClassDigest:
-    """Memoised on (class_name, schema_path); see _build_uncached."""
+    """Memoised on (class_name, schema_path); see _build_uncached.
+
+    Returns a **copy**. The cache used to hand out the stored object itself, so
+    any caller that mutated the result — or its `nested` entries, or their
+    `ranges`/`enums` dicts — silently changed what every later `build()` for
+    that class returned, for the life of the process (#528).
+
+    Found when a test stripped `ranges` to check the fingerprint responded, and
+    two later tests in the same file then measured an empty digest and failed.
+    Those failures read as real defects in the code under test, which is the
+    dangerous shape: the first instinct is to "fix" code that was correct.
+
+    The exposure is not confined to tests. `build` feeds the generation prompt,
+    `slot_spec` (what the fitness judge reads) and the digest fingerprint that
+    keys the fitness and sub-type caches. A mutation anywhere in a long-lived
+    process would change the prompt or the cache key for everything after it,
+    silently and with nothing recording that it happened.
+
+    A copy per call is cheap against what memoisation saves here: `SchemaView`
+    construction and `class_induced_slots` over the whole class, which is the
+    expensive part this cache exists to avoid repeating.
+    """
     key = (class_name, str(schema_path or ""))
     if key not in _BUILD_CACHE:
         _BUILD_CACHE[key] = _build_uncached(class_name, schema_path)
-    return _BUILD_CACHE[key]
+    return copy.deepcopy(_BUILD_CACHE[key])
 
 
 def _build_uncached(class_name: str, schema_path: Path | None = None) -> ClassDigest:
--- a/src/data_sheets_schema/schema_digest.py
+++ b/src/data_sheets_schema/schema_digest.py
@@ -355,8 +355,25 @@ def render(digest: ClassDigest) -> str:
     return "\n".join(lines)
 
 
+_TEXT_CACHE: dict[tuple[str, str], str] = {}
+
+
 def digest_text(class_name: str, schema_path: Path | None = None) -> str:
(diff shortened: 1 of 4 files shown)


realmarcin:
## Review — one finding, fixed in the last commit

**The copy landed on a hot path.** `LLMSlotFitnessScorer._context` calls `digest_text` once per fitness judgement to compute the cache key, so the deep copy was being paid 1,441 times in a sweep for a value that never changes:

```
render only (before #528)   0.47 ms
digest_text with copy       2.19 ms   +362%
```

2.5 s across a sweep of 1,441 API calls is negligible in context — but it is a 4× regression on a per-judgement function for **no benefit**: `digest_text` renders and discards, so it needs no copy at all.

Fixed by memoising the rendered string. **Caching the string is safe where caching the object was not** — strings are immutable, so there is nothing for a caller to mutate. `digest_text` is now effectively free, cheaper than before the copy existed.

## And the fix made one of my own tests trivial

`test_the_rendered_digest_is_unchanged_after_a_mutation` calls `digest_text` twice. With the text cache, the second call returns the string memoised *before* the mutation — so it passed without exercising anything, and would have held even if `build` went back to sharing the object.

It now clears `_TEXT_CACHE` between the two reads. Three further tests cover the text cache itself, including one asserting it cannot serve something an uncached `render` would not produce.

## Checked and clean

Only three `build()` call sites exist in `src/` and **none mutates** the result, so the hazard was theoretical in production — as stated in #528 rather than overclaimed. The value of the change is that nothing prevents it, and `build` feeds the generation prompt, the judge's spec and the cache key.

`test_different_classes_are_cached_separately` is worth flagging: it used `assertIsNot`, which would now pass **whatever the cache did**. Left unchanged, my PR would have quietly turned a real test into a tautology. It compares rendered content now.

**1882 passed, 4 skipped.**

PR #536, opened 2026-08-13
Bind the B2AI_* prefixes to the registry's namespaces (#535)

### LINKED ISSUE #535: Our B2AI_* prefixes expand to different IRIs than the registry's, and B2AI_DATA is not declared at all
Found researching #531 against `bridge2ai/standards-schemas`.

`data_sheets_schema.yaml` declares three `B2AI_*` prefixes. Compared with `standards_dataset_schema.yaml`, which is where the registry defines them:

| prefix | ours | registry |
|---|---|---|
| `B2AI_DATA` | **not declared** | `https://w3id.org/bridge2ai/standards-dataset-schema/` |
| `B2AI_ORG` | **not declared** | `https://w3id.org/bridge2ai/standards-organization-schema/` |
| `B2AI_SUBSTRATE` | `https://w3id.org/bridge2ai/b2ai-standards-registry/` | `https://w3id.org/bridge2ai/standards-datasubstrate-schema/` |
| `B2AI_TOPIC` | `https://w3id.org/bridge2ai/b2ai-standards-registry/` | `https://w3id.org/bridge2ai/standards-datatopic-schema/` |
| `B2AI_STANDARD` | `https://w3id.org/bridge2ai/b2ai-standards-registry/` | no such prefix — the registry class is `DataStandardOrTool` |

## Why this is worse than an undeclared prefix

`d4d runs identifiers` reports an undeclared prefix as unresolvable, which is honest. **A prefix declared to the wrong namespace resolves — to the wrong thing.** `B2AI_TOPIC:x` in one of our records and `B2AI_TOPIC:x` in the registry are different IRIs, and nothing downstream can tell: the CURIE is well-formed, the prefix is declared, and the audit passes it.

That is the same shape as #402's central finding one level up. The audit checks that a prefix is *declared*, never that it is declared *correctly*, because there was nothing to check it against. There is now.

## Scope

`B2AI_SUBSTRATE` and `B2AI_TOPIC` are the live wrong ones. `data_substrate` and `data_topic` hold 22 unresolvable values each (#457), so usage is small — but the wrongness is in the schema, not the records, and it affects every future value.

`B2AI_STANDARD` is this repository's invention. #403 named it as the natural home for grounding `DataStandardEnum`; that should point at whatever the registry actually uses for `DataStandardOrTool` rather than at a namespace only we define.

## Suggested

1. Correct `B2AI_SUBSTRATE` and `B2AI_TOPIC` to the registry's namespaces.
2. Declare `B2AI_DATA` and `B2AI_ORG`, which the registry defines and we do not — `B2AI_ORG` is also what #378 needs for organization identifiers.
3. Decide whether `B2AI_STANDARD` should be retired in favour of the registry's own term for `DataStandardOrTool`, or kept with a note that it is local.
4. Add a check that our `B2AI_*` declarations match the registry's, so this cannot drift again. The registry is a public YAML file; comparing against it is cheap and is the only thing that would have caught this.

Item 4 is the durable one. Items 1–2 fix today's values; only 4 stops the next divergence.

realmarcin:
Closes #535. Found researching #531.

## The defect

All three `B2AI_*` prefixes pointed at the registry's **homepage** URL while the registry declares per-schema namespaces:

| prefix | was | now |
|---|---|---|
| `B2AI_SUBSTRATE` | `…/b2ai-standards-registry/` | `…/standards-datasubstrate-schema/` |
| `B2AI_TOPIC` | `…/b2ai-standards-registry/` | `…/standards-datatopic-schema/` |
| `B2AI_DATA` | *not declared* | `…/standards-dataset-schema/` |
| `B2AI_ORG` | *not declared* | `…/standards-organization-schema/` |
| `B2AI_STANDARD` | `…/b2ai-standards-registry/` | unchanged — **local, and now recorded as such** |

So `B2AI_TOPIC:x` in one of our records and `B2AI_TOPIC:x` in the registry were **different IRIs**.

## Why this is worse than an undeclared prefix

`d4d runs identifiers` reports an undeclared prefix as unresolvable, which is honest. **A prefix bound to the wrong namespace resolves — somewhere else — and the audit passes it**, because the CURIE is well-formed and the prefix *is* declared. Nothing downstream can tell.

It is #402's finding one level up: the audit checked that a prefix was declared, never that it was declared *correctly*, because there was nothing to check it against.

## The durable half is the check

`b2ai_registry_prefixes.yaml` pins the registry's declarations with the commit they were taken at (`76122808`, 2026-01-12), and a test compares the schema against it.

**Deliberately offline.** A refresh is an edit to the pin with a diff a reviewer reads, not a network dependency that turns a registry-side change into a surprise CI failure. Same reasoning as the canonical prompt pins.

A second test requires every `B2AI_` prefix to be either the registry's or recorded in `local_only` **with a reason**: a prefix we invent is fine, a prefix we invent *silently* is how this started.

## Bearing on #531

`B2AI_DATA` is the registry's own `default_prefix` for datasets, and the four Grand Challenge datasets already have identifiers there (`B2AI_DATA:1`–`4`). Declaring it is the precondition for adopting them; this PR does not adopt them, since that needs the release-version question settled first.

`B2AI_ORG` is also what #378 needs for organization identifiers.

## Blast radius

Declared prefixes **33 → 35**. **No value's classification changes** — nothing yet uses `B2AI_DATA` or `B2AI_ORG`, and the two corrected prefixes bind values that were already counted as declared. The fix is to the schema and affects every future value rather than any existing one.

17 verdicts went `STALE` on the schema move; all 17 revalidate unchanged. Corpus at 156 valid, 0 stale.

**1898 passed, 4 skipped.**

### DIFF
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_provenance.yaml
@@ -304,7 +304,7 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/AI_READI_d4d_core.yaml
       md5: 19f4a17c60311d6bbc32fd06d266a76c
   schema:
-    full_sha256: 27992f75f8ee0c9905e9e4947d51298f7184b4bcd126cc45491e6616af127eb2
+    full_sha256: 992cbb48d6c8fbcfce68a88f6992de81851c070037baec5db07db8c594514e47
     core_sha256: 3ed2ec45e974af811362988ac084e73a4ec9e026f60b4294db1cf66676244f92
   recorded_by: d4d runs validate
 intermediates:
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/CHORUS_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/CHORUS_provenance.yaml
@@ -266,7 +266,7 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-api-generic_rep1/CHORUS_d4d_core.yaml
       md5: 57d52aa67e15baf249e791cd1ce32481
   schema:
-    full_sha256: 27992f75f8ee0c9905e9e4947d51298f7184b4bcd126cc45491e6616af127eb2
+    full_sha256: 992cbb48d6c8fbcfce68a88f6992de81851c070037baec5db07db8c594514e47
     core_sha256: 3ed2ec45e974af811362988ac084e73a4ec9e026f60b4294db1cf66676244f92
   recorded_by: d4d runs validate
 intermediates:
--- a/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-claudecode-generic_rep1/AI_READI_provenance.yaml
+++ b/data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-claudecode-generic_rep1/AI_READI_provenance.yaml
@@ -130,6 +130,6 @@ validation:
       path: data/d4d_concatenated/claudecode_agent_core/2026-08-11_claude-opus-5-claudecode-generic_rep1/AI_READI_d4d_core.yaml
       md5: 7fa94e1ccb88dcc9d7090ce6ff0b49d5
   schema:
-    full_sha256: 27992f75f8ee0c9905e9e4947d51298f7184b4bcd126cc45491e6616af127eb2
+    full_sha256: 992cbb48d6c8fbcfce68a88f6992de81851c070037baec5db07db8c594514e47
     core_sha256: 3ed2ec45e974af811362988ac084e73a4ec9e026f60b4294db1cf66676244f92
   recorded_by: d4d runs validate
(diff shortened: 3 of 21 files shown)


realmarcin:
## Review — one correction to my own wording, and one finding

**Correction.** The PR says the two corrected prefixes "bind values that were already counted as declared". That is wrong, and understates the result: **zero `B2AI_*` CURIEs exist anywhere in the corpus** — not `B2AI_SUBSTRATE`, not `B2AI_TOPIC`, not the two newly declared. Verified by grep across `data/d4d_concatenated/`.

So the claim "no value's classification changes" holds, and holds more strongly than stated: the prefixes were declared, pointed at the wrong namespace, and were never used by anything. The fix is entirely forward-looking.

**That absence is itself informative**, and led to the finding below.

## Finding — filed as #537

`data_substrate` and `data_topic` are ranged `uriorcurie` and hold **prose paragraphs**:

> `data_topic:` Type 2 diabetes mellitus and associated health domains, including vitals and physical measurements, blood and urine biomarkers, cardiac electrophysiology…

That is the `unit` problem again — ranged `uriorcurie` while holding `%` and `years` until #456 re-ranged it. It answers the question #457's step 1 explicitly left open about these two slots, and the answer is that **their 44 values are a range problem, not an identifier problem**, so they should leave the migration count the way #530's 1,067 did.

The reason the registry prefixes were never used is the same one #403 found for data standards: **nothing told a run which registry terms exist.** A prefix with no visible vocabulary behind it cannot be populated, so runs described the concept instead of naming it.

## Verified

- Declared prefixes 33 → 35; audit counts unchanged (7,250 unresolvable, 1,067 unverified).
- 17 verdicts went `STALE` on the schema move; all 17 revalidate unchanged. Corpus 156 valid, 0 stale.
- **1898 passed, 4 skipped.**

PR #540, opened 2026-08-13
Add the AI-READI release-3.0.0 RO-Crate and the v2.0 licence (#539)

### LINKED ISSUE #539: Updated AI-READI RO-Crate (v3.0.0) is not in the manifest, and the manifest's license source is two versions stale
The AI-READI RO-Crate has been updated and shared as a Drive folder (`ro-crate-metadata.json`, 110 KB, plus croissant/datasheet/preview variants and `ai_ready_score.json`). Retrieved and examined; not yet added to the repository.

## It describes release 3.0.0

Root entity `ark:59853/rocrate-b2ai-aireadi-release-3-0-0`, `version: 3.0.0`, `identifier: https://doi.org/10.60775/fairhub.3`, `datePublished: 11/17/25`, `contentSize: 3.82 TB`, with 9 component crates (ECG, OMOP, FLIO, retinal OCT/OCTA/photography, wearable blood glucose, wearable activity, environmental sensor).

**The canonical D4D record already says 3.0.0**, so the crate and the record agree on the release. This is not a stale-record problem.

## 1. The bundle contains no crate content at all

`grep -c "ark:59853\|ro-crate\|fairscape" AI_READI_preprocessed.txt` → **0**. The manifest's ten sources are two publications, two documentation sites, a license PDF, three fairhub pages and the IRB protocol. No RO-Crate.

So this is genuinely new evidence, and adding it is a manifest change that rehashes the bundle — every AI_READI record then drifts (#452) and the arm would need regenerating to consume it. That is a decision, not a cleanup.

## 2. The manifest's license source is two versions behind

```yaml
- id: dataset_license
  url: https://zenodo.org/records/10642459/files/AI-READI-LICENSE-v1.0.pdf?download=1
```

- manifest fetches **v1.0**
- the canonical record states **"AI-READI custom license v2.0"**
- the v3.0.0 crate points to **`https://doi.org/10.5281/zenodo.17555036`** — a different Zenodo record again

This one is independent of the crate decision and worth fixing regardless: the bundle is feeding runs a licence two releases old for a dataset documented at 3.0.0.

## 3. The file is not valid UTF-8

`ro-crate-metadata.json` fails `json.load` under UTF-8:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xa9 in position 5096
b'"copyrightNotice": "Copyright \xa9 2026 AI-READI"'
```

It is cp1252 — `©` as a single `0xA9` byte. RFC 8259 requires UTF-8 for interchanged JSON, so strict parsers reject it outright. Worth reporting upstream; any ingestion here needs an explicit decode until it is fixed.

## 4. Facts the crate carries that the record does not

| crate field | value |
|---|---|
| `contentSize` | 3.82 TB |
| `deidentified` | true |
| `fdaRegulated` | false |
| `humanSubjectResearch` | Yes |
| `confidentialityLevel` | HL7:2N (normal) |
| `conformsTo` | fairscape profile |
| `hasPart` | 9 component crates |

Several map onto slots the record leaves empty — including `is_deidentified`, and the governance slots added in #503/#510.

## 5. One disagreement to preserve, not resolve

- crate `dataGovernanceCommittee`: **"AI-READI Consortium"**
- record `data_governance.committee_name`: **"Data Access Committee"** (from the BMJ protocol)

Both are plausibly right at different granularity. Under the uniform decision rules this is represented, not silently reconciled.

Also note the crate's `author` is a single string, `"AI-READI Consortium"`, where the record carries 17 named creators from the publications. The record is **richer** here; a crate-derived regeneration must not lose that.

## Decision needed

Whether to add the crate to `source_manifest.yaml` as an AI_READI source. It would improve coverage of exactly the slots the corpus is weakest on, and it rehashes the bundle and invalidates the current arm's inputs. Fixing the license URL (item 2) is worth doing either way.

realmarcin:
Closes #539. Both changes authorised on the basis that D4Ds are being regenerated shortly, so bundle drift is acceptable.

## The RO-Crate is new evidence, not a refresh

The AI_READI bundle contained **no crate content at all** — `grep "ark:59853\|ro-crate\|fairscape"` returned **0**, now returns **51**.

It carries facts no other AI_READI source states:

| field | value |
|---|---|
| `contentSize` | 3.82 TB |
| `deidentified` | true |
| `fdaRegulated` | false |
| `humanSubjectResearch` | Yes |
| `confidentialityLevel` | HL7:2N (normal) |
| `dataGovernanceCommittee` | AI-READI Consortium |
| `hasPart` | 9 component crates |

Several map onto slots the corpus is weakest on, including the governance slots added in #503/#510.

## The licence was one version stale

The manifest fetched `AI-READI-LICENSE-v1.0` while the dataset is at release 3.0.0 and both the crate and the canonical record state v2.0.

**Verified before switching**, not inferred: Zenodo 17555036 is *"License terms for reusing the AI-READI dataset"*, version 2.0, 7 November 2025 — and the downloaded PDF's own first page reads `AI-READI DATA LICENSE AGREEMENT (Version 2.0)`. The bundle now carries v2.0 and no longer mentions v1.0. `verification_url` moved with it; it still pointed at the superseded record.

Worth noting: the record already said v2.0, having learned it from the docs site rather than from the source designated to carry it. A correct answer by an unreliable route, now by the intended one.

## Preprocessing now decodes non-UTF-8 sources

The crate is **cp1252** — `©` as a single `0xA9` byte — so `read_text(encoding="utf-8")` raised at byte 5096. The JSON path crashed outright; the byte-copy path would have put bytes in the bundle the concatenator cannot read.

UTF-8 is tried first, so nothing well-formed is reinterpreted, and the fallback prints rather than passing silently. `errors="ignore"` was rejected: it would delete the copyright symbol and leave no trace that it had.

The copy path re-encodes **only** files that fail to decode, so every currently-valid source is still copied byte-for-byte and **no other project's bundle moves**.

The raw file is stored exactly as retrieved, defect included — `raw/` is what was fetched, and normalising it there would hide an upstream bug this repository should be reporting.

## Consequences, all expected and all self-reported

- bundle md5 `20150c10` → `0f3abb51`
- **bundle drift 64 → 68 records.** The four AI_READI records that still matched now do not. The pinned figure in `test_bundle_drift` is updated *with the reason* — that pin exists precisely so the corpus cannot absorb an input change silently, and updating it is how the event gets acknowledged.
- `test_schema_straddle` no longer asserts the arm's bundles are `current`. Those records correctly state the bytes they read; the path no longer resolves to them. `absent` would still be a defect and is still rejected.

## Kept deliberately

Old v1.0 raw and preprocessed files. Records generated before today declare them, and deleting them breaks the trail for the same reason #520 refuses to rewrite historical records.

## Weakest link, flagged in the manifest

The crate URL is a **Google Drive direct-download link**. fairhub returns 404 for `/api/datasets/3/ro-crate-metadata.json` and the two obvious variants, so no citable URL exists. This is the least durable provenance in the manifest and should be replaced when the crate is published somewhere stable.

`runs check --strict` 0, `scope --check --strict` 0, **1898 tests pass**.


realmarcin:
## Review

Checked the two claims most likely to be wrong.

**"No other project's bundle moves."** Confirmed by hash, not by inspection of the code path:

```
CHORUS           9b2ef4b65d67957f79362266cab0bc7a   (unchanged — matches what its records pin)
CM4AI            1dfd34e5610fed7c22bea1f09c0bc60c
VOICE            dcd717170da6762569c0b4eeafc1c3d2
VOICE_PEDIATRIC  008212accc7ec95bf3f0121566b588c2
```

`git diff --stat` over `concatenated/` shows **one file changed**. The selective re-encode did what it was scoped to do — had it re-encoded unconditionally, all four would have moved and drifted every record in the corpus.

**The bundle is clean end to end:**

```
valid UTF-8:                                    533,995 chars
© present:                                      True
Version 2.0:                                    True
ark:59853/rocrate-b2ai-aireadi-release-3-0-0:   True
3.82 TB:                                        True
dataGovernanceCommittee:                        True
11 files in the table of contents
```

So the cp1252 defect is resolved in the bundle while the character it hinged on survives — which is the outcome `errors="ignore"` would have quietly failed.

## One thing I got wrong on the way, worth recording

I first set `processed_file` to `.json`. The pipeline rejected it:

```
❌ Error: Canonical processed output must be .txt: aireadi_ro_crate_metadata_2026-08-12.json
```

That is the manifest validation working — the existing `fairhub_api_dataset_3_2026-07-27.json` already establishes `.json → .txt`, and I had not matched the convention. Fixed before anything downstream ran.

## Not done, deliberately

**No records regenerated.** The bundle now differs from what every AI_READI record consumed, and all 20 correctly report `drifted`. Regeneration is yours to schedule; nothing here pre-empts it.

PR #541, opened 2026-08-13
Render the registry vocabulary the slots already declared (#538)

### LINKED ISSUE #538: values_from: B2AI_TOPIC is never rendered, contradicts its own description, and no correct value obeys it
Found chasing #537's surviving question — why runs write prose into `data_topic` when its description says to prefer omission.

`D4D_Composition.yaml` declares, on both `data_topic` and `data_substrate`:

```yaml
values_from:
  - B2AI_TOPIC        # and B2AI_SUBSTRATE
```

Three things are wrong with that at once.

## 1. It is never rendered, so no run sees it

`schema_digest` does not emit `values_from` for any slot. `d4d api render-prompt` and the agentic digest both carry the range, the description, enum values and nested obligations — and not this. A constraint a run cannot see cannot be followed.

## 2. No correct value obeys it

The 78 well-formed values in these two slots, by source:

| vocabulary | values |
|---|---|
| MeSH (`meshb.nlm.nih.gov`) | 22 |
| Cellosaurus | 20 |
| EDAM (`edamontology.org`) | 19 |
| OBO purl | 12 |
| EBI | 5 |
| **B2AI_TOPIC / B2AI_SUBSTRATE** | **0** |

Zero. Every value anyone would call correct violates the declared constraint.

## 3. It contradicts the description directly above it

```yaml
description: >
  General topic of each instance, as an ontology term — records use
  GO, MeSH, EFO and NCIT IRIs as well as Bridge2AI standards terms.
values_from:
  - B2AI_TOPIC
```

The prose names four external ontologies and treats Bridge2AI terms as one option among them; `values_from` admits only the last. A reader following the description produces values the declaration rejects, which is what happened 78 times.

## Why rendering it would be the wrong fix

The obvious remedy — make the digest emit `values_from`, as #486 did for nested ranges and #403 for enum values — is **actively harmful here**. It would push runs toward a vocabulary nothing uses and away from the five they actually use correctly.

So the ordering matters: the declaration has to be right before it is made visible.

## Options

1. **Widen it** to the vocabularies in use — MeSH, Cellosaurus, EDAM, OBO, EBI — plus B2AI_TOPIC. Matches the description and the corpus. Needs those to be expressible as `values_from` targets, which for external ontologies may mean dropping it in favour of description guidance.
2. **Remove it.** It constrains nothing today (unenforced and unrendered) and misstates practice. Honest, and loses a stated intention.
3. **Keep it and mean it** — decide B2AI_TOPIC is the target vocabulary, render it, and treat the 78 external-ontology values as migration. Highest cost, and it contradicts #456's finding that those values are correct.

Recommend (1) or (2) over (3). This is the same shape as #403: a slot whose vocabulary was never shown to the runs, filled by improvisation — except here the improvisation produced *good* values and the declaration is what is wrong.

## Only two slots

`values_from` appears twice in the whole schema, both in `D4D_Composition.yaml`. So this is contained, and whichever option is chosen, the digest should then render it — a declared constraint that no run can see is worth nothing whichever vocabulary it names.

realmarcin:
Closes #538. Implements option 3 — **the reverse of what I first recommended on the issue**, after resolving every term in both slots.

## What the corpus showed

All **23** well-formed `data_substrate` IRIs are the wrong *kind* of thing:

| value | × | is |
|---|---|---|
| `cellosaurus.org/CVCL_0419` | 11 | MDA-MB-468, a breast cancer cell line |
| `cellosaurus.org/CVCL_B5P3` | 9 | KOLF2.1J, an iPSC line |
| `meshb…D005453` | 1 | Fluorescence |
| `meshb…D013058` | 1 | Mass Spectrometry |
| `meshb…D064113` | 1 | CRISPR-Cas Systems |

The slot means *"type of data"*. A cell line is biological material; fluorescence and mass spectrometry are assay techniques.

Meanwhile the **11 prose values** — tabular survey data, DICOM imaging, waveform signals — are correct in substance and had no term available to be written as.

**`B2AI_SUBSTRATE` has DICOM, Comma-separated values, Waveform Data, JSON, Image, Text and Parquet — and always did.** `B2AI_TOPIC` has Diabetes, Voice, Voice Disorders, Clinical Observations and Cell. The declaration was right; it was invisible.

## Why option 3 and not the two I first preferred

- **Widening** `values_from` to MeSH/EDAM/OBO would have *blessed the cell lines*.
- **Removing** it would have discarded a correct declaration because it had never been enforced.

## What changed

Rendered in the digest **and** in `slot_spec`, from one function, so the run and the fitness judge see the same list.

Both go through the **nested** class — `data_topic` and `data_substrate` are attributes of `Instance`, not slots of `Dataset`, so a top-level rendering reaches neither. Same trap #486 hit with nested ranges.

Pinned in `b2ai_registry_vocabularies.yaml` with the commit it came from, for the reason the prompt pins exist. A `values_from` naming no pinned vocabulary renders **nothing** rather than guessing.

## Cost, and the guard I raised

```
digest       37,680 → 40,230   (+2,550, 6.8%)
fingerprint  5c26b444 → bfa6db52
```

The prefix is stated once and terms listed `id=name`, halving what repeating `B2AI_SUBSTRATE:` 81 times would cost.

The size guard is raised **40k → 44k deliberately**. It exists to catch a regression that *doubles* the digest; this is 6.8% and buys the thing whose absence produced 23 wrong values.

**The fingerprint move is intended** — the judge's question changed, so cached fitness labels must not be reused (#465). A test strips the vocabulary and asserts the fingerprint responds.

## A guard I had to narrow, carefully

`test_digest_carries_no_dataset_facts` forbade the bare token `b2ai`, which now matches the *vocabulary namespace*. Narrowed to the dataset-specific forms — and **widened** to also catch `aireadi` and `fairhub`, which it had been missing.

A companion test asserts the vocabulary names **are** present and that no dataset is ever named on the same line, so the exemption cannot quietly become one that lets a real fact through. A controlled vocabulary is structure: it says what kinds of value a slot admits, never what a dataset contains.

## Not done

Existing records are not rewritten (#520). The 23 mis-typed values stand as evidence of what was generated without the vocabulary.

**1913 passed, 4 skipped.** `runs check --strict` 0, `prompts check --strict` 0.

### DIFF
--- a/src/data_sheets_schema/b2ai_registry_vocabularies.yaml
+++ b/src/data_sheets_schema/b2ai_registry_vocabularies.yaml
@@ -1,0 +1,166 @@
+# Bridge2AI registry vocabularies for `values_from` slots, pinned (#538).
+#
+# `data_topic` and `data_substrate` declare `values_from: B2AI_TOPIC` /
+# `B2AI_SUBSTRATE`. Nothing rendered that declaration, so no run has ever been
+# shown the vocabulary — and the corpus reflects it: all 23 well-formed
+# `data_substrate` IRIs name cell lines (Cellosaurus MDA-MB-468, KOLF2.1J) or
+# assay techniques (Fluorescence, Mass Spectrometry, CRISPR-Cas Systems), none
+# of which is a type of data. The 11 prose values are correct in substance and
+# had no term to be expressed as.
+#
+# Pinned rather than fetched, for the same reason as the prompt pins: a
+# refresh is an edit with a diff a reviewer reads, not a network dependency
+# that turns a registry-side change into a surprise CI failure.
+#
+# Refresh with:
+#   curl -sSL https://raw.githubusercontent.com/bridge2ai/b2ai-standards-registry/main/project/data/DataSubstrate.tsv
+#   curl -sSL https://raw.githubusercontent.com/bridge2ai/b2ai-standards-registry/main/project/data/DataTopic.tsv
+source:
+  repository: bridge2ai/b2ai-standards-registry
+  commit: d38bcf2408c960a75dfadc295b47990a3890c26f
+  committed: '2026-06-04T19:20:27Z'
+  retrieved: '2026-08-13'
+  paths:
+    - project/data/DataSubstrate.tsv
+    - project/data/DataTopic.tsv
+
+vocabularies:
+  B2AI_SUBSTRATE:
+    B2AI_SUBSTRATE:1: Array
+    B2AI_SUBSTRATE:2: Associative Array
+    B2AI_SUBSTRATE:3: BIDS
+    B2AI_SUBSTRATE:4: BigQuery
+    B2AI_SUBSTRATE:5: Column Store
+    B2AI_SUBSTRATE:6: Comma-separated values
+    B2AI_SUBSTRATE:7: Data
+    B2AI_SUBSTRATE:8: Data Frame
+    B2AI_SUBSTRATE:9: Database
+    B2AI_SUBSTRATE:10: Delimited Text
+    B2AI_SUBSTRATE:11: DICOM
+    B2AI_SUBSTRATE:12: Directed acyclic graph
+    B2AI_SUBSTRATE:13: Document Database
+    B2AI_SUBSTRATE:14: Graph
+    B2AI_SUBSTRATE:15: Graph Database
+    B2AI_SUBSTRATE:16: HDF5
+    B2AI_SUBSTRATE:17: Heap
+    B2AI_SUBSTRATE:18: Hierarchical Array
+    B2AI_SUBSTRATE:19: Image
+    B2AI_SUBSTRATE:20: JSON
+    B2AI_SUBSTRATE:21: KGX TSV
+    B2AI_SUBSTRATE:22: MongoDB
+    B2AI_SUBSTRATE:23: MySQL
+    B2AI_SUBSTRATE:24: N-Dimensional Array
+    B2AI_SUBSTRATE:25: Neo4j
+    B2AI_SUBSTRATE:26: Neural Network Model
+    B2AI_SUBSTRATE:27: NNEF
+    B2AI_SUBSTRATE:28: ONNX
+    B2AI_SUBSTRATE:29: Pandas DataFrame
+    B2AI_SUBSTRATE:30: Parquet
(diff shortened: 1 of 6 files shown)


realmarcin:
## Review — one finding, and it would have hurt one project

**The vocabulary is not exhaustive, and the slot's own guidance does not reach the run.**

A nested attribute's *description* is not rendered in the digest, so `data_topic`'s "Where no term is found, prefer omission over a prose topic" is invisible. That did not matter before, because the vocabulary was invisible too. **After this change a run sees 56 terms and no instruction to decline.**

CHORUS is the live case. Its subject is *acute and critical care*, and none of the 56 `B2AI_TOPIC` terms names it:

```
Activity Monitoring, Biology, Cardiomyocyte, Cell, Cheminformatics,
Clinical Observations, Data, ... Voice, Voice Disorders, Waveform, mHealth
```

No `Critical Care`, no `Radiology`, no `Intensive Care`. So as first committed, this would have put 56 near-neighbours in front of a run with nothing saying none of them is right — **picking the closest is precisely the invention `OTHER` was added to prevent in #403.** I would have replaced 23 wrong-kind values with a different wrong value and called it a fix.

The rendering now ends: *"If no term fits, omit the slot rather than approximate, and never restate the subject as prose here."* Both failure modes, because the corpus shows both — 23 of the wrong kind and 11 in prose.

## Coverage, pinned rather than described

| | |
|---|---|
| substrate concepts the corpus needs | **all 8 present** — DICOM, CSV, Waveform Data, JSON, Text, Image, Parquet, Data Frame |
| AI_READI / CM4AI / VOICE topics | well covered — Diabetes, Glucose Monitoring, Ophthalmic Imaging, Cell, Proteome, Voice, Voice Disorders |
| **CHORUS topic** | **no term** |

The CHORUS gap is asserted *as a gap*, so if `Critical Care` is ever added upstream the test fails and the gap gets closed rather than forgotten.

Two concepts that look missing and are not: `OMOP` and `Electronic health record` have no substrate term, but OMOP CDM is a data *standard* and belongs in `conforms_to_standard`, where #403's enum already has `OMOP_CDM`. Right slot, not a gap.

Digest 40,230 → 40,434; fingerprint `bfa6db52` → `622e6d03`.

**1919 passed, 4 skipped.**

PR #542, opened 2026-08-13
State the core header block instead of implying it (#515)

### LINKED ISSUE #515: The rendered spec says use the header block 'exactly'; the playbook requires two more lines on core files

realmarcin:
Closes #515.

## The contradiction

The prompts gave a full-record header under **"use exactly"**, then one line:

> The core file uses "phase 2" and the core schema path in the corresponding lines.

But the playbook requires two *further* lines on core files — `# Sources:` naming the same-run full record, and `# Phase 4 reconciliation: completed`. Both instructions are authoritative in their own terms, so every run had to resolve the conflict alone.

**Fifteen did, four ways.** Measured across the 2026-08-11 arm, the line naming the projected-from full record appears as:

| spelling | records |
|---|---|
| `# Sources:` | 8 |
| `# Full D4D input:` | 5 |
| `# Full record input:` | 1 |
| wrapped onto a continuation line, no key | 1 |

One record also opened `# D4D Datasheet` rather than `# D4D Core Datasheet`.

## The fix

The core block is now written out in full — the four lines that differ from the full-record block, and the two that have no counterpart there — and it says *why* `# Sources:` is required rather than leaving it to be inferred.

`HEADER_FIELDS` gains `Sources` and `Phase 4 reconciliation`. Every core record carries the link in substance and `parse_header` discarded it in **all** of them, because the key was never listed — the same silent drop `Prompt` suffered until #399.

## All four conditions, and why

v4 has no records. v1 and v3 are being re-baselined for the comparison, which is the cheapest moment to absorb a pin rotation. **v2 has 16 records** that become `superseded` — but the analysis plan's open mechanism question needs a v2 arm, so leaving the defect there would only make it recur.

## What the rotation does to the corpus

Two commits, deliberately: the edit, then the declaration that it is now canonical. Each pin keeps its predecessor as `superseded`, so runs made under the old text still verify.

- **The canonical arm reads `superseded`.** Pinned once and since rotated is ordinary evolution; `uncanonical` — never published at all — is the failure, and remains one.
- **18 runs read `unverifiable`** on the render gate, which recognises the file changed and says *the recorded hash still stands, it just cannot be re-derived here*. It does **not** report `mismatch`, the instruction-substitution the gate exists to catch.
- `runs check --strict` and `prompts check --strict` both stay **0**.

## Five tests updated, not weakened

They asserted `canonical`/`match` on the arm. Those were true when written and are properties of a **fresh** record rather than of a record — the same lesson #497's digest test already carries. They now assert the durable property: never `mismatch`, never `uncanonical`. Accepting anything would have thrown the checks away; this preserves exactly the distinction that matters.

**1919 passed, 4 skipped.**

## Consequence for the sweep

Today's CHORUS canary was generated under the **pre-rotation** text, so it must be re-run before any fan-out. Fixing and fanning out in the same step is what the canary rule forbids.

### DIFF
--- a/src/data_sheets_schema/provenance.py
+++ b/src/data_sheets_schema/provenance.py
@@ -75,7 +75,15 @@ HEADER_FIELDS = ("Generation Method", "Agent runtime", "Provider", "Model",
                  # `Prompt` was absent, so `parse_header` silently discarded
                  # the one field a historical prompt could be recovered from
                  # (#399). 15 records carry it and read as carrying nothing.
-                 "Prompt", "Prompt components")
+                 "Prompt", "Prompt components",
+                 # The core record's link to the full record it was projected
+                 # from (#515). Present in substance in every core record and
+                 # discarded by `parse_header` in all of them, because the key
+                 # was never listed — the same silent drop `Prompt` suffered
+                 # until #399. It was also spelled four ways across one arm,
+                 # which is what "use exactly, except add two lines" produces
+                 # when fifteen runs each resolve the contradiction alone.
+                 "Sources", "Phase 4 reconciliation")
 
 
 def _md5(path: Path) -> str | None:
--- a/src/download/prompts/canonical_hashes.yaml
+++ b/src/download/prompts/canonical_hashes.yaml
@@ -53,37 +53,69 @@ files:
       \ is left to the manifest scope declaration rather than restated here (#422)."
     sha256: b7db067c2852b57287cb68bd1fd4f3bfc229496cc2a4bd33065bb9558a35b215
   src/download/prompts/d4d_generic_arm_prompt.md:
-    bytes: 5754
-    pinned_at_commit: 225f224544aa8f9ca8ac7f61f57aab9813536866
-    pinned_on: '2026-08-10'
-    reason: 'initial pin (#432): the file as it stood when the registry was created. All 77
-      recorded prompt hashes in the corpus match these bytes, so pinning here fails nothing
-      retroactively.'
-    sha256: 0fbc626bc3a51d74a8ba3d8e2813752f3bc5ce9366ee5f19c99089afdfd9e8dc
+    bytes: 6737
+    pinned_at_commit: 7141fdc63e43f5049bf98289a142c85bd63107ed
+    pinned_on: '2026-08-13'
+    reason: '#515: the core header block is stated in full instead of implied by ''the core
+      file uses phase 2 and the core schema path''. That one line contradicted the playbook''s
+      requirement for ''# Sources:'' and ''# Phase 4 reconciliation: completed'', and fifteen
+      runs of the 2026-08-11 arm resolved it four different ways.'
+    sha256: f5873222343632a97e0d9200e283924751cddfc70e1c879bbeb9f6416cf95d86
+    superseded:
+    - pinned_on: '2026-08-10'
+      reason: 'initial pin (#432): the file as it stood when the registry was created. All
+        77 recorded prompt hashes in the corpus match these bytes, so pinning here fails nothing
+        retroactively.'
+      retired_on: '2026-08-13'
+      sha256: 0fbc626bc3a51d74a8ba3d8e2813752f3bc5ce9366ee5f19c99089afdfd9e8dc
   src/download/prompts/d4d_generic_arm_prompt_v2.md:
-    bytes: 9110
-    pinned_at_commit: 225f224544aa8f9ca8ac7f61f57aab9813536866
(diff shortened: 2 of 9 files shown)


realmarcin:
## Review

Checked the two things most likely to be wrong.

**Does the rendered core block actually satisfy the playbook?** Yes — 16 lines, every placeholder substituted, nothing left literal:

```
# D4D Core Datasheet for CHORUS Dataset
# Generation Method: schema-grounded agentic, phase 2
…
# Sources: data/preprocessed/concatenated/CHORUS_preprocessed.txt + data/d4d_concatenated/…/CHORUS_d4d.yaml
…
# Schema: src/data_sheets_schema/schema/data_sheets_schema_core_all.yaml
# Generated: 2026-08-13
# Phase 4 reconciliation: completed
```

The `Phase 4` line is **byte-identical to `PHASE4_HEADER`** in `d4d_pair_consistency.py`, which is what `--sync-core` appends and what the completion criterion checks. So the prompt and the tooling now agree on one string rather than three.

*(My first check reported the playbook as not requiring that line — a literal-string search that missed the backtick form at `d4d-full-core.md:557`. It does require it.)*

**Is `# Generated:` right?** It renders today's date, not the label's — `run_date` defaults to now, and a run labelled `2026-08-14` executed today writes `2026-08-13`. That is correct and matches existing semantics: the label is the *arm's* date, each record's Generated line is when it was actually produced. Same distinction already noted for the agentic arm's 2026-08-12 headers under a 2026-08-11 label.

## What the rotation cost, stated plainly

Nothing silently. The corpus now reads:

- canonical arm → `superseded` (was `canonical`)
- 18 runs → `unverifiable` on the render gate, with the reason given
- `runs check --strict` and `prompts check --strict` → still **0**

`unverifiable` here is the honest answer, not a downgrade: the gate says *the recorded hash still stands, it just cannot be re-derived here*. It reserves `mismatch` for the instruction-substitution it exists to catch, and that is still fatal.

## The five tests

They asserted `canonical`/`match` — true when written, and properties of a *fresh* record rather than of a record. I replaced them with the durable claim (never `mismatch`, never `uncanonical`) rather than deleting them. That keeps the distinction between "pinned once, since rotated" and "never published at all", which is the whole point of #436's bypass check.

## Blocking consequence for the sweep

**Today's CHORUS canary was generated under the pre-rotation text and must be re-run.** The canary rule is explicit that a fix means re-canary, not fix-and-fan-out. I have not touched it — it stays on disk as evidence of the pre-rotation behaviour, and the sweep should start from a fresh canary once this merges.

PR #549, opened 2026-08-15
Run the pair checker the API path claims to have run (#544)

### LINKED ISSUE #544: 11 of 12 v4 pairs fail full/core consistency, and phase 4 reports them reconciled
Found reviewing the `2026-08-13_claude-opus-5-api-generic-v4` arm. Every record validates individually and `runs check --strict` exits 0, which is why nothing surfaced this: both checks read one file at a time.

## The measurement

**70 errors across 12 records; 11 of 12 pairs fail.**

| project | rep1 | rep2 | rep3 |
|---|---|---|---|
| AI_READI | 10 | 8 | 6 |
| CHORUS | 6 | **0** | 3 |
| CM4AI | 6 | 9 | 3 |
| VOICE | 9 | 4 | 6 |

By code: `shared-slot-content` 67, `resource-projection-presence` 3.

By slot, the top offenders are the *evidence-channel* slots — `source_caveats` (9), `notes` (7) — then `data_governance` (5), `is_deidentified` (5), `informed_consent` (4).

## The direction is the surprising part

Of the list-length disagreements, **5 have core larger than full** and 3 the reverse. Concretely, CHORUS rep1:

```
$.acquisition_methods                 full=5, core=6
$.data_governance.stewardship_roles   full=1, core=3
$.existing_uses                       full=2, core=3
$.sampling_strategies                 full=1, core=2
$.is_deidentified.deidentification_details
    full:  'De-identification is applied before data enter the shared enclave…'
    core:  same sentence + three further sentences on tokenization, imaging
           de-identification, privacy scanning and OMOP geocoding
```

**Core is meant to be a projection of full.** A core record holding facts the full record does not is not a projection — it is a second, divergent record. And the extra content is not noise: the deidentification detail in core is materially better than full's.

## Why it is plausible, and why it is still a defect

Phase 2 generates core from *the bundle plus the completed full record*, so it can legitimately find things phase 1 missed. That explains the asymmetry. What it does not explain is phase 4: reconciliation exists to make the pair consistent, and every record carries `# Phase 4 reconciliation: completed`.

So either reconciliation ran and did not converge, or it converged on a definition of consistency that is not the one `d4d_pair_consistency` implements. Both are worth knowing; the header currently asserts a property the records do not have.

## Why nothing caught it

- `linkml-validate` validates each artifact alone.
- `runs check --strict` aggregates per-record status.
- `d4d_pair_consistency` is invoked by the *agentic* playbook at phase 4. The API path has `reconcile_full` and `reconcile_core` phases but **nothing runs the pair check afterwards as a gate**.

That last point is the actionable one: the API runner should run the pair check after reconciliation and record the result in provenance, so a divergent pair is a reported outcome rather than something a reviewer stumbles on three days later.

## Suggested

1. **Run `validate_pair_data` at the end of an API run** and record the result. Report, do not fail — a divergent pair is still usable evidence, and the point is that it stops being invisible.
2. **Decide what phase 4 should do when core is richer.** Copying core's value into full is one answer, dropping it is another, and they are different products. The `# Phase 4 reconciliation: completed` header should not be written until whichever rule is chosen actually holds.
3. Consider whether `source_caveats` and `notes` should be identity slots at all. They are commentary about the record rather than dataset facts, and 16 of 70 errors are there — this may be the same category question #499 settled for `conforms_to_class`.

## Not proposed

Rewriting the 12 records. The divergence is evidence of how the pipeline behaves, and the records validate.

realmarcin:
Closes #544.

The API path writes `# Phase 4 reconciliation: completed` into every core
header and never ran the checker that would substantiate it. `validate_pair_data`
has been in the tree with its own tests since long before, and had no caller
anywhere in the pipeline.

| arm | pairs checked | divergent |
|---|---|---|
| agentic 2026-08-11 (playbook runs the checker) | 15 | **0** |
| API v4 2026-08-13 | 12 | **11** |

Every one of those 12 records validates individually, and `d4d runs check
--strict` exits 0 on the arm. `linkml-validate` and every gate in `runs check`
read one file at a time; this is a property of two files together, so no
single-file gate could have seen it.

## What lands

- **`pair_consistency()`** in `api_runner`, called after repair — repair
  rewrites both records, so a pair checked before it describes bytes that no
  longer exist. Returns `None` when the checker cannot run, keeping "not
  established" distinct from "consistent".
- **The verdict is falsifiable.** Both artifacts pinned by md5, for the reason
  `validation_block` pins its own (#426, #433).
- **`pair_status()` + a `runs check` reporter.** Reported, never fatal: the
  records are individually valid, and the whole corpus predates the check.
  122 records report `unrecorded` — which is not `consistent`.

## A defect this PR's own review caught

The first version derived the full record's path from the provenance file's
location. The two records are not in one directory — full under
`{method}/{label}/`, core under `{method}_core/{label}/` — so it looked for a
file that was never there and would have called every pair stale. Paths are
now recorded alongside the hashes, and `test_pins_name_the_paths_rather_than_rebuilding_them`
names the mistake.

## Not in scope

Backfilling the 12 v4 records, and fixing the divergences themselves. This PR
makes the property visible and records it going forward.

Suite: 1936 passed, 4 skipped.

### DIFF
--- a/src/data_sheets_schema/api_runner.py
+++ b/src/data_sheets_schema/api_runner.py
@@ -1165,8 +1165,13 @@ def pair_consistency(spec: RunSpec) -> dict[str, Any] | None:
         from data_sheets_schema.d4d_pair_consistency import (load_pair_schema,
                                                              validate_pair_data)
         import yaml as _yaml
-        if not (spec.full_path.exists() and spec.core_path.exists()):
-            return None
+        missing = [str(f) for f in (spec.full_path, spec.core_path)
+                   if not f.exists()]
+        if missing:
+            # Not None. None means the block was never written; this run wrote
+            # one and the file it needed was absent, which `validate_outputs`
+            # will also have flagged. Two different states, two answers.
+            return {"ran": False, "reason": f"missing: {', '.join(missing)}"}
         pair = load_pair_schema(FULL_SCHEMA_PATH, CORE_SCHEMA_PATH)
         full = _yaml.safe_load(spec.full_path.read_text(encoding="utf-8")) or {}
         core = _yaml.safe_load(spec.core_path.read_text(encoding="utf-8")) or {}
@@ -1195,8 +1200,11 @@ def pair_consistency(spec: RunSpec) -> dict[str, Any] | None:
         "errors": len(report.errors),
         "warnings": len(report.warnings),
         "identity_slots": len(report.identity_slots),
+        # Bounded, and says so. A list silently cut at 20 reads as a complete
+        # one; `errors` above is the true count either way.
         "findings": [{"code": i.code, "path": i.path, "message": i.message[:200]}
                      for i in report.errors[:20]],
+        "findings_truncated": max(0, len(report.errors) - 20) or None,
     }
 
 
--- a/src/data_sheets_schema/api_runner.py
+++ b/src/data_sheets_schema/api_runner.py
@@ -1142,6 +1142,64 @@ def validate_outputs(spec: RunSpec) -> list[dict[str, str]]:
     return problems
 
 
+def pair_consistency(spec: RunSpec) -> dict[str, Any] | None:
+    """Does the full/core pair this run produced actually agree? (#544)
+
+    The API path has `reconcile_full` and `reconcile_core` phases and writes
+    `# Phase 4 reconciliation: completed` into every core header, but nothing
+    ever ran the pair checker afterwards. The agentic playbook does, at its own
+    phase 4, and the difference is the whole difference: across the
+    2026-08-13 v4 arm 11 of 12 pairs failed, against 0 of 15 in the
+    2026-08-11 agentic arm.
+
+    Every individual record validated in both arms, which is why neither
+    `linkml-validate` nor `runs check --strict` noticed — both read one file at
+    a time, and this is a property of two files together.
+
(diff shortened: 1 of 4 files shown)


realmarcin:
Review round complete. Findings filed as #550, #551, #552.

- **#551** — three defects in this PR as first written, all fixed here: the
  staleness check rebuilt a path that never existed (would have reported STALE
  for every pair); two silent caps; a missing artifact returning the same value
  as "never checked".
- **#550** — the 11 divergent v4 pairs this check makes visible. Open; not a
  code change.
- **#552** — the check is inert for the 122 existing records, and the agentic
  arm runs the checker but records it nowhere. Deliberately out of scope:
  backfilling means writing 122 records that each carry a header comment a
  `yaml.safe_dump` round-trip strips, so it needs its own canary.

Also verified `check_provenance` accepts the new top-level key — `execute()`
runs it on its own output and raises if unusable, so a rejection there would
fail every live run after the model spend. Pinned by `RecordGateTest`.

Suite: 1938 passed, 4 skipped.

realmarcin:
The arm comparison in this PR's description is wrong. Backfilling the check across the corpus (#552) gives roughly four in five records divergent in **both** arms — not agentic 0/15 against API 11/12. The agentic playbook does run the checker, and its records still diverge. Details and the schema-pin check that rules out a stale-schema artifact are in #550.

Issue #547, opened 2026-08-14
A record supplied correct RORs that appear nowhere in its bundle — right answer, no evidence


realmarcin:
Found doing content review of the `2026-08-13_claude-opus-5-api-generic-v4` arm. Every structural check passes on this record; only checking values against the *bundle* reveals it.

## What happened

VOICE rep1 carries 19 external identifiers that appear nowhere in its input bundle. They are not wrong:

| identifier | resolves to | in bundle? |
|---|---|---|
| `ror.org/032db5x82` | University of South Florida | **no** (the *name* appears 16×) |
| `ror.org/02r109517` | Weill Cornell Medicine | **no** |
| `ror.org/042nb2s44` | Massachusetts Institute of Technology | **no** |

The run learned the institution names from the evidence and supplied their ROR identifiers from model memory.

**This is the fabrication class the provenance guard exists to prevent, in its hardest form: right answer, no evidence.** Checking the value cannot detect it — every ROR is correct. Only checking the source can.

## Scope: one record in twelve

```
VOICE rep1    5 grounded ·  0 minted-on-real-base · 19 ABSENT
VOICE rep2    2 grounded ·  0 minted              ·  0 ABSENT
VOICE rep3    2 grounded · 14 minted              ·  0 ABSENT
AI_READI     27 grounded · 17 minted              ·  0 ABSENT
CM4AI         9 grounded · 12 minted              ·  0 ABSENT
CHORUS        0 external identifiers at all
```

"Minted-on-real-base" is legitimate and separated deliberately: `doi:10.60775/fairhub.3#split-train` is a DOI the bundle *does* contain with a local fragment appended. Conflating the two would have made this look systematic when it is not.

## A second defect in the same values

`https://ror.org/02r109517#rameau`, `#sean-parker`, `#sigaras`, `#bahr`, `#csd`, `#watts` — a person fragment appended to an **institutional** ROR. That identifier asserts "Weill Cornell Medicine, fragment rameau" and is being used to identify a person. A ROR identifies an organisation; a person needs an ORCID or a minted id in a namespace of our own.

## Why nothing caught it

- `linkml-validate` — a well-formed `uriorcurie` is valid whatever it names.
- `d4d runs identifiers` — a resolvable IRI on a real host is the *best* outcome it recognises.
- the audit phase — it caught a false creator identity in AI_READI rep1, so it can catch this class, but it did not here.
- `d4d_pair_consistency` — compares the pair, not either against the evidence.

[...]

PR #555, opened 2026-08-15
Ground external identifiers against the bundle they were read from (#547)

realmarcin:
Closes the tooling half of #547. The rule half (say in the prompt that an
identifier must come from the evidence) is v5 prompt content and stays open.

## The gap

Nothing in the pipeline compared an identifier to its evidence. VOICE rep1
supplies RORs that appear nowhere in its bundle, and **every one is correct** —
`ror.org/032db5x82` really is the University of South Florida, whose name the
bundle states 16 times. The run learned the institution from the evidence and
the identifier from memory.

No existing check can see this. `linkml-validate` accepts any well-formed
`uriorcurie`; `d4d runs identifiers` treats a resolvable IRI on a real host as
the *best* outcome; the pair checker compares the two records to each other.
Only the bundle distinguishes a correct value from an attested one.

## Three outcomes, kept apart

| status | meaning |
|---|---|
| `grounded` | the bare identifier occurs in the bundle |
| `minted_fragment` | a bundle identifier with a local fragment appended — `doi:10.60775/fairhub.3#split-train`. Attested base, our fragment. |
| `absent` | neither |

Collapsing the middle one would make a narrow defect look systematic: AI_READI
rep1 has 17 minted fragments and 0 absent. Notation is normalised first, so
`https://ror.org/x`, `ror.org/x` and `ROR:x` are one identifier — without that,
the CURIE conversion this arm went through would have made every grounded value
read as absent.

## Measuring found more than the issue reported

#547 sampled one replicate per project. Distinct-identifier counts reproduce
its figures **exactly** for the replicates it sampled:

| | grounded | minted | absent |
|---|---|---|---|
| AI_READI rep1 | 27 | 17 | 0 |
| CM4AI rep1 | 9 | 12 | 0 |
| VOICE rep3 | 2 | 14 | 0 |
| VOICE rep1 | 5 | 0 | **12** |
| **CM4AI rep3** | 40 | 0 | **10** ← not in #547 |

The CM4AI bundle contains exactly one ROR (`0153tk833`, verified by grep) and
rep3's records carry ten others. So the defect is in two records, not one.

The issue's second defect is also larger than reported: a person fragment on an
institutional ROR — `ROR:02r109517#rameau`, asserting "Weill Cornell Medicine,
fragment rameau" about a person — appears **14 times** in VOICE rep1, not the 6
listed.

## Shape

Same as #549 and #553: computed by the runner, pinned to the bundle's md5,
reported by `d4d runs check`, never fatal (#520 — annotate rather than
rewrite).

16 tests. Suite: 1974 passed, 4 skipped.

### DIFF
--- a/src/data_sheets_schema/api_runner.py
+++ b/src/data_sheets_schema/api_runner.py
@@ -1241,6 +1241,30 @@ def report_claims_block(spec: RunSpec) -> dict[str, Any] | None:
     return out
 
 
+def grounding_block(spec: RunSpec) -> dict[str, Any] | None:
+    """Are this run's external identifiers in the bundle it read? (#547)
+
+    VOICE rep1 supplied 19 RORs that appear nowhere in its bundle. Every one is
+    correct — the run learned the institutions from the evidence and the
+    identifiers from memory. Right answer, no evidence, and no existing check
+    can see it: `linkml-validate` accepts any well-formed `uriorcurie`, and
+    `d4d runs identifiers` treats a resolvable IRI as the best possible
+    outcome.
+
+    Reported, never fatal, following #520.
+    """
+    try:
+        from data_sheets_schema.grounding import check_run
+        out = check_run(spec.full_path, spec.core_path, spec.bundle)
+    except Exception as exc:                                       # noqa: BLE001
+        return {"checked": False, "reason": str(exc)[:200]}
+    if out.get("checked"):
+        from data_sheets_schema.provenance import _md5
+        out["artifacts"] = {"bundle": {"path": str(spec.bundle),
+                                       "md5": _md5(spec.bundle)}}
+    return out
+
+
 REPAIR_SYSTEM = (
     "You repair the shape of Datasheets-for-Datasets records. The schema "
     "digest defines the required structure. The validator findings are the "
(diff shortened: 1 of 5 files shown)


realmarcin:
Review round found that this PR's own table was wrong twice, both from counting occurrences where identifiers were meant — filed as #556 and fixed here.

Corrected, distinct identifiers, pooled across each pair:

| | grounded | minted | absent | org-fragment |
|---|---|---|---|---|
| AI_READI rep1 | 27 | 17 | 0 | 0 |
| AI_READI rep2 | 10 | 14 | 0 | 0 |
| AI_READI rep3 | 10 | 13 | 0 | 0 |
| CHORUS rep1–3 | 0 | 0 | 0 | 0 |
| CM4AI rep1 | 9 | 12 | 0 | 0 |
| CM4AI rep2 | 2 | 0 | 0 | 0 |
| **CM4AI rep3** | 40 | 0 | **10** | 0 |
| **VOICE rep1** | 5 | 0 | **19** | **7** |
| VOICE rep2 | 2 | 0 | 0 | 0 |
| VOICE rep3 | 2 | 14 | 0 | 0 |

VOICE rep1 now reproduces #547's figures exactly (5 / 0 / 19). The organisational-fragment count is 7 distinct identifiers against the 6 #547 listed, not the 14 I claimed. CM4AI rep3 remains the record #547 did not sample.

Suite: 1976 passed, 4 skipped.

Issue #545, opened 2026-08-14
The API path never receives the playbook's uniform decision rules — 135 British spellings in the v4 arm prove it


realmarcin:
Predicted in the #502 decision note as a known gap; now measured.

`d4d evaluate spelling` on `2026-08-13_claude-opus-5-api-generic-v4_rep1`:

```
135 occurrence(s) in generated prose, 1 in text quoted from a bundle (left alone)
   prohibited_uses    organisations -> organizations
   source_caveats     Organisation  -> organization
```

The American-English rule was placed in `.claude/commands/d4d-full-core.md` (#502) rather than in the condition prompts, deliberately: editing a prompt rotates its pin and would have demoted the canonical arm. The note recorded the cost plainly —

> **The API path does not read the playbook.** `d4d api run` renders a condition prompt and nothing else, so API-generated records are unaffected by this rule.

— and this is that cost, at 135 occurrences in one replicate set.

## The general problem is bigger than spelling

The playbook's **uniform decision rules** are the ones described as applying "whether or not a prompt file was used to launch":

- prefer omission over inference
- represent what disagreeing sources state rather than silently selecting one
- `Dataset` admits one referent
- no target slot count
- American English (#502)
- identifiers as CURIEs where a prefix is declared (this session)

**An API run receives none of them.** They reach the agentic path because the launch instruction tells the agent to open the playbook; the API path renders a condition prompt and calls the model. So the two paths are following different rule sets while both being described as "the method", which is the same class of problem #420 was filed for — instructions reaching the model through files nobody tracks.

Two of these rules were added in the last two days *specifically* to fix defects, and neither reaches half the pipeline.

## Options

1. **Render the uniform rules into the API instruction**, from the playbook file, so one source serves both paths. Changes the rendered text, so it re-baselines every condition and needs a pin rotation — but it is the only option that makes the rules actually uniform.
2. **Duplicate them into the condition prompts.** Same pin cost, plus two copies to drift apart, which is what #518/#521 kept finding in the build.
3. **Accept the divergence and document it** — the paths are different methods and the playbook is agentic-only. Cheapest, and it means "uniform decision rules" is a misnomer.

Recommend (1), sequenced with the next deliberate re-baselining rather than mid-arm. Whichever is chosen, the playbook's claim that these apply "whether or not a prompt file was used to launch" is currently false for the API path and should not stay written that way.

## Scope

[...]

Issue #340, opened 2026-08-05
condition_of() returns None for every v3 and v4 run


realmarcin:
Found reviewing #339. Pre-existing for v3; #339 adds v4 with the same gap.

`runs.condition_of()` infers a run's prompt condition from the prompt paths in
its provenance, using a hardcoded chain:

```python
if "d4d_generic_arm_prompt_v2.md" in joined:  return "generic_v2"
if "d4d_tuned_arm_prompt.md" in joined:       return "tuned"
if "d4d_generic_arm_prompt.md" in joined:     return "generic"
return None
```

Neither v3 nor v4 is listed, and `d4d_generic_arm_prompt_v3.md` does not contain
`d4d_generic_arm_prompt.md` as a substring — the `_v3` sits between `prompt` and
`.md`. So a v3 run falls through every branch and the function returns **None**.

**Silent, and directly on the rerun path.** The planned rerun is v3. Anything
that asks which condition produced a record gets `None` and cannot tell a v3 run
from an unrecognisable one. `arm_delta` and the analysis that follows sit
downstream of this.

**Fix:** derive from `CONDITION_PROMPTS` rather than restating the list — the
same lesson `cli/api.py` already carries in a comment, where writing the
condition list out three times left `generic_v2` unreachable from every entry
point. `tuned` needs checking first because it deliberately shares v1's file and
is distinguished by `d4d_tuned_arm_prompt.md` appearing alongside.

realmarcin:
Fixed in #339.

PR #559, opened 2026-08-15
Add the generic-v5 condition: four rules, and why it is four (#547, #531, #545)

realmarcin:
Closes the rule half of #547 and settles #531. Completes #545 by emptying
`PLAYBOOK_ONLY`.

v5 = v4 plus one marked block. **The block holds four rules**, which breaks the
one-rule-per-version convention v2, v3 and v4 kept.

## Why four, and what it costs

Two of them are not new instructions. American English (#502) and CURIE form
were added to the playbook mid-arm, deliberately, to avoid rotating a pin while
runs were in flight — so they have reached the agentic path ever since and the
API path never. Bringing them into the condition prompt is a **parity fix**, and
their expected effect on the agentic arm is nil by construction.

The other two are new and inseparable:

- **an identifier is a fact and comes from the evidence** (#547). VOICE rep1
  supplied 19 RORs absent from its bundle, CM4AI rep3 another ten. Every one
  correct, none attested.
- **where the evidence supplies no identifier, hang one off an identifier it
  does** — a fragment, not a new prefix (#531, decided). A fragment on an
  organisation's ROR does not name a person.

Stated apart they would be incoherent: forbidding unattested identifiers without
saying what to write instead pushes a model toward inventing a local namespace,
which is the ~12,000-value defect #531 records.

**The cost is stated, not hidden.** A v4-against-v5 comparison measures the
block and cannot attribute any effect to a rule inside it.
`notes/generic_v5_analysis_plan.md` says so, is registered before any run, and
lists what would falsify the block rather than only what would confirm it.

## Both directions of the divergence guard

The #545 guard looked only for rules reaching the agentic path and not the API
path. Two of v5's four arrived the other way round — written for the prompt —
and a rule only the API arm follows is the same defect wearing the other shoe.
Both new rules go into the playbook too, and the guard now checks both ways.

It also stops probing by text prefix. Until v5 the copies were byte-identical so
a prefix matched; they cannot be now, because the playbook cites issue numbers
and real identifiers while the generic prompt may contain neither. The
correspondence is a declared table instead — weaker than byte-identity, and the
honest weaker thing, since a table someone must edit beats a probe that silently
stops matching.

## The genericity tests earned their keep

A first draft of the block used a real dataset DOI as its example of a minted
fragment. `test_no_dataset_identifiers` refused it. A generic arm that
illustrates a rule with one project's identifier is priming every other project.
The block now contains no digits at all, names no project and names no slot.

## Two guards that assumed v5 could not exist

- the prompt-registry CLI fixture built its temp prompts from a hardcoded
  `("_v2", "_v3", "_v4")`. Derived from `CONDITION_PROMPTS` now — same shape as
  #467.
- `test_american_english_rule` asserted no condition prompt carries the rule.
  #502's real constraint is that a prompt *already run* may not acquire it. v5
  carries it from birth and no record names v5. The test now says that, and
  asserts the premise instead of assuming it.

Pin taken at the commit the bytes were committed in; `d4d api prompts check
--strict` reports 11 files, 0 off-pin.

Suite: 2016 passed, 4 skipped.

### DIFF
--- a/.claude/commands/d4d-full-core.md
+++ b/.claude/commands/d4d-full-core.md
@@ -232,6 +232,23 @@ to every project. Enforce them whether or not a prompt file was used to launch:
   `uriorcurie` — and never to prose. A URL inside a sentence or a citation is
   text, not an identifier, and must be left exactly as written.
 
+- **An identifier is a fact and comes from the evidence.** Take it from the
+  declared bundle or omit it; do not supply an identifier you recognise but the
+  bundle does not state. A correct identifier the evidence does not contain is
+  still an unsupported claim, and to a reader who was not present it is
+  indistinguishable from a wrong one. Naming an organisation the bundle names is
+  grounded; adding that organisation's ROR from your own knowledge is not — the
+  2026-08-13 arm did exactly this, supplying RORs for institutions the bundle
+  names only in prose (#547). `grounding.absent` in the provenance record counts
+  them.
+
+- **Where something needs an identifier and the bundle supplies none, hang it
+  off one the bundle does supply** — a fragment on the identifier of the thing
+  it is part of, `doi:10.60775/fairhub.3#split-train`, rather than a new
+  namespace (#531). A person is identified by an ORCID and an organisation by a
+  ROR; **a fragment appended to an organisation's ROR does not identify a
+  person**, it asserts something false about that organisation.
+
 - **Write generated prose in American English** — `program`, `organization`,
   `analyze`, `license`, `center`, `labeling`, `enrollment`. This is house style
   for the text *you* compose, and it applies to identifiers you mint as well as
--- a/notes/generic_v5_analysis_plan.md
+++ b/notes/generic_v5_analysis_plan.md
@@ -1,0 +1,64 @@
+# generic v5 — what the block predicts, registered before any run
+
+Written 2026-08-14, before any v5 generation. Its purpose is to hold the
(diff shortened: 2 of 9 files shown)


realmarcin:
Review round complete.

Verified v5 is reachable end to end rather than merely present: `d4d api run --condition` offers it, `prompt_files()` makes it pinnable, `condition_of` maps a v5 prompt hash back to the condition, and `condition_delta('generic_v4','generic_v5')` reports a difference on `base` alone. All of those derive from `CONDITION_PROMPTS` rather than restating a list — the #340 lesson — so none needed editing.

Rendered the v5 request spec offline as the free check before any billed call: 7,804-character body, the marked block present in what the model receives, and the rationale sections correctly excluded from it.

Two things came out of the round:

- **An attribution corrected in the rationale.** It said #531 found ~12,000 minted ids; #531 established 1,094 VOICE values across five spellings, and the corpus-wide figure is mine from implementing this. Fixed, and the pin rotated for it.
- **#560 filed.** That rotation is the finding. The pin hashes the whole file including rationale prose `prompt_body` never sends, so correcting a sentence of documentation costs exactly what changing a decision rule costs. Free here because no record names generic_v5 yet; mid-arm it would mean rationale errors go uncorrected because correcting them looks like tampering. Not fixed here — it changes the meaning of every existing pin and deserves its own commit.

Suite: 2016 passed, 4 skipped.

Issue #546, opened 2026-08-14
Reconciliation reports assert removals that did not happen, from a false premise about the core schema


realmarcin:
Found reviewing the v4 arm, by reading the reports against the records rather than trusting either alone.

## The claim, and the record

CM4AI rep3's reconciliation report:

> **Action:** the `distributions` block was removed from the core record in its entirety.

The record has **10 `distributions` entries**.

Same pattern in AI_READI rep1 (10 entries retained) and VOICE rep2 (3 retained). Three of twelve records assert a removal that did not occur.

*(A fourth match — VOICE rep1 `errata` — is a false positive from my scan: the sentence is a summary count, not a removal claim. Not counted.)*

## The premise is also false

All three reports justify the removal the same way. AI_READI rep1:

> The core record carried a ten-entry `distributions` block with keys `path`, `bytes`, `format`, `media_type`, `conforms_to`, `conforms_to_standard`. **No such slot exists in the `CoreDataset` inventory.** It appears to have been invented as a core-side substitute for the full record's `file_collections`.

`distributions` **is** a declared `CoreDataset` slot, with range `CoreDistribution`, and every key listed is a declared `CoreDistribution` slot — `path`, `bytes`, `format`, `media_type`, `conforms_to`, and `conforms_to_standard` since #403. Verified against `data_sheets_schema_core_all.yaml`.

So the audit reasoned from a false statement about the schema toward a removal, then did not perform it. **The records are correct by accident, not by design** — had the resolution been applied, three records would have lost a valid ten-entry block, and the redistribution described (into `distribution_formats`, `conforms_to_standard`, `total_size_bytes`) would have been a lossy re-encoding of content that already had a proper home.

## Why this is worse than a wrong record

A reconciliation report is the human-readable audit trail — the artifact a reviewer reads *instead of* diffing YAML. Two failures compound:

1. **It states schema facts that are false.** Nothing checks an audit's claims against the schema, though the schema is machine-readable and right there.
2. **It states outcomes that did not occur.** Nothing checks the report against the record it describes.

Either alone is recoverable. Together they mean the report cannot be relied on for what changed, and `# Phase 4 reconciliation: completed` in the header asserts a process whose own account of itself is wrong.

## Related, and probably the same root cause

#544 — 11 of 12 pairs fail `d4d_pair_consistency` while every report claims reconciliation completed. A phase that misreports its own actions is a plausible explanation for a phase that reports success without achieving consistency.

## Suggested

[...]

Issue #521, opened 2026-08-12
make full-schema and gen-core-schema have wrong prerequisites, so editing a module is a silent no-op


realmarcin:
Found landing #499, which edits `D4D_Base_import.yaml`. Sibling of #518 and arguably worse, because the two combine into a loop with no exit.

## full-schema had no prerequisites at all

```make
$(SOURCE_SCHEMA_ALL):
	@echo "Generating D4D-Full schema with merged imports..."
```

No prerequisite list. Make therefore rebuilds it only when the file is **missing**, so on any existing checkout:

```
$ make full-schema
make: Nothing to be done for `full-schema'.
```

regardless of what changed. That is why `regen-all` has to `rm -f $(SOURCE_SCHEMA_ALL)` before calling it — the workaround is in the repo and the cause was not.

## Combined with #518 it is a closed loop

Before #518's fix, editing a module produced:

1. `make check-sync` → *"Merged schema is out of date … Fix: make full-schema"*
2. `make full-schema` → *"Nothing to be done"*
3. `make check-sync` → same message

The tool names a remedy that cannot work. Anyone following it either discovers `regen-all` by reading the Makefile or concludes check-sync is noise — and the second is the likelier outcome, which is how a real staleness gets ignored.

## gen-core-schema had the same shape, narrower

```make
$(D4D_CORE_SCHEMA_ALL): $(D4D_CORE_SCHEMA) src/data_sheets_schema/schema/D4D_Core.yaml
```

Two prerequisites and not the twelve modules `D4D_Core` imports. Editing `D4D_Base_import.yaml` left the merged core schema stale with make reporting nothing to be done. Observed directly: the `d4d:perRecord` annotation added for #499 was in the module and absent from `data_sheets_schema_core_all.yaml`, and `load_pair_schema` read it as unset.

## Fix

Both rules derive their module list from the corresponding schema's own `imports:` block, the same derivation #518 introduced for `check-schema-sync.sh`:

```make
SCHEMA_IMPORTS = $(addprefix $(SOURCE_SCHEMA_DIR),$(addsuffix .yaml,$(shell \
	sed -n '/^imports:/,/^[a-z\#]/p' $(SOURCE_SCHEMA_PATH) | \
	sed -n 's/^[[:space:]]*-[[:space:]]*\(D4D_[A-Za-z_]*\)[[:space:]]*$$/\1/p')))
```

[...]

Issue #525, opened 2026-08-12
Playbooks are hashed per record but never checked for drift, so editing one is invisible


realmarcin:
Found reviewing #524, which edits `.claude/commands/d4d-full-core.md` to add the American-English rule.

Every agentic record hashes the playbooks it read:

```yaml
playbooks:
  hash_algorithm: sha256
  files:
    - path: .claude/commands/d4d-full-core.md
      sha256: c1e03a0b…
    - path: .claude/commands/d4d-agent.md
    - path: .claude/agents/d4d-provenance-guard.md
```

**Nothing ever compares those hashes to the files again.** `d4d runs check` reports bundle drift (#452), render-gate status (#419), canonical prompt pins (#432), verdict schema pins (#433), reasoning-effort basis (#448) and now schema straddle (#517). It says nothing about playbooks.

## Why this matters more than it looks

The playbook is where the *uniform decision rules* live — "prefer omission over inference", "represent what the evidence states rather than silently selecting one", `Dataset` admits one referent, and as of #524 the American-English rule. These are instructions that shape every slot in every record.

So a playbook edit is a change to the method, exactly like a prompt edit. Prompt edits are guarded three ways: the render gate catches text edited after rendering, the canonical pin catches text edited before it (#432), and `runs check --strict` is fatal on `uncanonical`. Playbook edits are guarded by nothing.

Concretely, right now: the 15 records of the 2026-08-11 canonical arm hash a version of `d4d-full-core.md` that #524 replaces. After that merges, every one of them attests to reading bytes that no longer exist at that path, and `runs check --strict` still exits 0.

## This is the #452 shape, one level up

`bundle_drift` asks whether a record's declared *input* still hashes to what the record consumed. The same question about its declared *instructions* is not asked. #431 already established that playbooks are "instructions reaching the model through files nobody tracks" and fixed *which* files are recorded; it did not make the recorded hashes falsifiable.

## Proposed

`playbook_drift(method, label, project)` mirroring `bundle_drift`, with the same four outcomes — `current`, `drifted`, `absent`, `unrecorded` — reported by `runs check` and **never fatal**, for the same reason bundle drift is not: a drifted record is still valid evidence of what was generated, it simply can no longer be re-derived from the files it names.

Non-fatal also matters because playbooks are edited deliberately and often, and a gate would make every improvement to the method a corpus-wide failure.

## Not proposed

[...]

PR #564, opened 2026-08-15
Give both agentic entry points one copy of the uniform rules (#563)

### LINKED ISSUE #563: /d4d-agent carries none of the uniform decision rules, and the #545 guard does not look at it
Found answering a question about whether the skills need updating after the v5
work.

## Three playbooks are hashed into every agentic record

```
.claude/commands/d4d-full-core.md
.claude/commands/d4d-agent.md
.claude/agents/d4d-provenance-guard.md
```

The uniform decision rules live in **`d4d-full-core.md` only**. Grepping
`d4d-agent.md` for every one of them returns zero: no "prefer omission", no
"sources that disagree", no "one referent", no "no target slot count", no CURIE
rule, no American English rule, and neither of the two identifier rules added
for v5.

## Why this is usually harmless, and when it is not

`d4d-full-core.md` delegates extraction to `d4d-agent.md` — *"Follow the method
in `.claude/commands/d4d-agent.md` (read it first)"* — while keeping the rules
itself. So a paired run entered through `/d4d-full-core` gets both, and every
record in the corpus was produced that way.

But `d4d-agent.md` is also a standalone entry point. Its own header says it
produces full records only and to use `/d4d-full-core` when core is wanted too.
**A `/d4d-agent` run therefore generates a full D4D under none of the uniform
decision rules** — no omission-over-inference, no identifier-from-evidence, no
CURIE form.

## The guard has the same blind spot

`tests/test_playbook_reach.py` compares the rules in `d4d-full-core.md` against
the current condition prompt, in both directions since #559. It never opens
`d4d-agent.md`. So the file that would silently diverge is the one nothing
checks — which is the #518/#521 pattern one level further out than #545 caught
it.

## Also: the agentic path does not run the two new checks

`d4d-full-core.md`'s phase 4 runs `d4d_pair_consistency` twice. It does not run
the report check (#546) or the grounding check (#547), both of which now run
automatically on the API path. So the arm that reads the rules is the arm that
does not verify them, and vice versa.

## Suggested

1. `d4d-agent.md` either states the uniform rules or refuses to run standalone
   and points at `/d4d-full-core`. Restating them creates a fourth copy to keep
   in sync, so pointing is probably right — but it has to be explicit, because
   right now the file reads as a complete method.
2. Extend the guard to every hashed playbook rather than the one.
3. Add the report and grounding checks to phase 4, so both arms verify what
   both arms are told.

realmarcin:
Closes #563.

## The gap

The uniform decision rules lived in `d4d-full-core.md` and nowhere else.
`/d4d-agent` is a **standalone entry point that produces full records**, and it
carried none of them — grepping it for all eight returns zero.

Nothing on disk is affected: every record in the corpus was generated through
`/d4d-full-core`, which holds the rules and delegates extraction to
`d4d-agent.md`. The gap was in what that file permitted, not in what it
produced. It would have mattered the first time someone ran `/d4d-agent`
directly for a v5 record.

## One copy, named by everything that needs it

Extracted to `.claude/commands/d4d-uniform-rules.md`. Both entry points and the
provenance guard now point at it and none restates it — a fourth hand-kept copy
is the defect, and this is #518 / #521 / #545 one level further out.

Hashed like the other playbooks, because it is now where the rules live and a
record that did not hash it could not show which rules it followed. Records made
after this hash four files rather than three.

## The guard had the same blind spot

`test_playbook_reach` only ever opened `d4d-full-core.md` — so the file that
could silently diverge was precisely the one nothing checked. It now reads
`AGENT_PLAYBOOKS` and requires every hashed playbook to state the rules or name
the file that does, so a fifth playbook arrives already covered.

A second test asserts the rules are restated nowhere else. A pointer sitting
beside a stale duplicate is worse than either alone: two sources of truth and no
signal about which is current.

## Phase 4 now runs the checks the API path already runs

The agentic phase 4 ran `d4d_pair_consistency` and nothing else. The grounding
check (#547) and report check (#546) run automatically on every API run and not
at all here — so the arm that reads the rules was the arm that did not verify
them. Both are now steps 6 and 7, with what each finding means and what to do
about it.

## One unrelated fixture

`test_prompt_condition_consistency` asserted a literal `3` playbooks, so a test
about *fabricated* playbooks failed for a reason having nothing to do with
fabrication. Derived from `AGENT_PLAYBOOKS` now.

Suite: 2022 passed, 4 skipped.

### DIFF
--- a/.claude/agents/d4d-provenance-guard.md
+++ b/.claude/agents/d4d-provenance-guard.md
@@ -8,6 +8,13 @@ model: inherit
 color: yellow
 ---
 
+
+> **The uniform decision rules are in `.claude/commands/d4d-uniform-rules.md`.**
+> Read them with this file. The evidence boundary and the rule that an
+> identifier must come from the evidence are the same principle applied to two
+> kinds of value, and enforcing one without the other is how a run supplied
+> correct RORs its bundle never stated (#547, #563).
+
 # D4D Provenance Guard
 
 Apply this policy to every D4D generation run, regardless of whether the runtime
--- a/.claude/commands/d4d-agent.md
+++ b/.claude/commands/d4d-agent.md
@@ -8,8 +8,19 @@ approach.
 > full D4D, audits both against current sources in Phase 3, and performs strict
 > schema-derived full/core reconciliation in Phase 4.
 
-Before generation, read and enforce
-`.claude/agents/d4d-provenance-guard.md`.
+Before generation, read and enforce **both**:
+
+- `.claude/agents/d4d-provenance-guard.md` — the evidence boundary;
+- `.claude/commands/d4d-uniform-rules.md` — the uniform decision rules that
+  apply to every condition, every project and both runtimes.
+
+The rules are not restated here. Until #563 they lived only in
+`d4d-full-core.md`, so a run entered through this command — which is a complete
+standalone method for full records — followed none of them: not
+omission-over-inference, not identifier-from-evidence, not CURIE form. Every
+record in the corpus happens to have been generated through `/d4d-full-core`,
+which reads them and delegates extraction here, so nothing on disk is affected.
+The gap was in what this file permitted, not in what it produced.
 
 ## Task Overview
 
--- a/.claude/commands/d4d-full-core.md
+++ b/.claude/commands/d4d-full-core.md
@@ -443,7 +443,12 @@ identical slot and proves consistency across the pair.
    from data_sheets_schema.identifiers import uriorcurie_slots
    r = check_run(Path('<full_file>'), Path('<core_file>'), Path('<bundle>'),
                  uriorcurie_slots())
-   print(r['distinct']); [print(f) for f in r['findings']]"
+   if not r.get('checked'):
+       print('NOT CHECKED:', r['reason'])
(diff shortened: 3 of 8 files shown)


realmarcin:
Review round complete; #565 filed and fixed here.

The finding is that I verified the two new phase-4 snippets against CHORUS, which has **zero external identifiers** — so the grounding command printed three zeroes and looked correct. Run against a record with findings it fails twice: `KeyError` when `check_run` declines (bundle absent, which 59 records are already exposed to), and 78 lines of output for VOICE rep1's 19 ungrounded identifiers. Both fixed; the snippet now branches on `checked` and prints one line per distinct identifier.

A command written into a playbook is code that runs unattended, and it was tested on the one project where it could not fail.

Also verified on the corpus rather than assumed:

- `d4d runs check --strict` still exits 0.
- Playbook drift now reports 30 records reading changed bytes, 0 still matching — expected and honest, since the extraction rewrote two playbooks and added a third. Non-fatal by design (#525).
- Records made after this hash four playbooks; the count is derived from `AGENT_PLAYBOOKS` in the one test that had it as a literal.

Suite: 2022 passed, 4 skipped.

PR #570, opened 2026-08-16
Pin the prompt body separately from its rationale (#560)

### LINKED ISSUE #560: The prompt pin hashes rationale prose the model never sees, so a typo fix looks like a re-baseline
Observed reviewing #559, and deliberately not fixed there.

## What happened

I corrected an attribution in the v5 prompt's **rationale** — the sections above
`## Prompt body`, which `prompt_body()` does not send to the model. The
condition's actual text did not change by a byte.

`d4d api prompts check --strict` failed anyway, correctly by its own rules, and
resolving it rotated the pin. The registry now records:

```
✓ pinned at 606e7382cd4e…
  previous e4d0570921f8… kept as superseded, so runs made under it still verify
```

A reader of that history sees a condition re-baselined between two pins. It was
not: the instruction is identical and only a sentence of documentation moved.

## Why it did not matter this time, and when it will

No record names `generic_v5` yet, so the rotation was free. Mid-arm it would not
be. The rule that a prompt edit must rotate its pin exists precisely so a
mid-arm change is visible and expensive (#432) — and under the current hashing,
**fixing a typo in a rationale paragraph carries exactly the same cost as
changing a decision rule.**

The predictable consequence is the one the rule was written against: rationale
errors go uncorrected during an arm, because correcting them looks like tampering.

## Options

1. **Leave it.** Hashing the whole file is simple and impossible to game by
   moving text across the boundary. The cost is that documentation ossifies for
   the length of an arm.
2. **Pin the body separately.** Record both `sha256` (whole file, for the audit
   trail) and `body_sha256` (what `prompt_body` returns). `check --strict` fails
   on a body change; a whole-file-only change reports as `annotated` — visible,
   not fatal. The render gate already compares the rendered spec against the
   recorded instruction, so the body is the thing that has to be immovable.
3. **Split the files**: `d4d_generic_arm_prompt_v5.md` for the instruction and a
   sibling rationale document. Cleanest boundary, most churn, and it separates
   the reasoning from the rule it explains — which is why the rationale is in
   the file to begin with.

(2) preserves the guarantee that matters while letting prose be corrected, and
it costs one extra field. I have not implemented it because it changes the
meaning of every existing pin and belongs in its own reviewable commit rather
than riding along with a new condition.

realmarcin:
Closes #560.

The pin hashed the whole prompt file, including the rationale sections that
`prompt_body` never sends to the model. Correcting a sentence of documentation
failed `check --strict` exactly as changing a decision rule does, and resolving
it rotated the pin — leaving a history that reads as a condition re-baselined
when the instruction had not moved by a byte. **That happened once already**,
while adding v5.

The predictable consequence is the one the rule exists to prevent: rationale
errors go uncorrected during an arm, because correcting them looks like
tampering.

## Three statuses

| status | meaning | `--strict` |
|---|---|---|
| `canonical` | file and body both at their pin | passes |
| `annotated` | body at its pin; only rationale differs | **passes**, reported |
| `uncanonical` | the body moved, or the pin cannot tell | fails |

Demonstrated on the real v5 prompt:

```
rationale-only edit → strict exit 0
  ⓘ  annotated  …_v5.md: the prompt body is at its pin; only text outside
     `## Prompt body` differs, which is rationale the model is never sent.

body edit → strict exit 1
  ❌ uncanonical  …_v5.md hashed 52f6838017…, which is neither the pinned
     606e7382cd… nor any hash it was pinned at before
```

## Existing pins gain a body hash without rotating

The bytes are unchanged, so deriving a second hash from them adds no claim —
the same ground as `backfill-effort`. A pin that still lacks one **fails
closed**: guessing would turn an unchecked file into a passing one. Files with
no `## Prompt body` section (components, the tuned block) get no body hash and
keep whole-file checking, which is right because every byte of those is sent.

## The record gate is deliberately not softened

`status_of_hash` judges a hash a *record* wrote, for bytes that need not exist
any more. Those cannot be decomposed into body and rationale, and comparing
today's on-disk body against the pin would let a record whose prompt matches
nothing pass as `annotated` because an unrelated file happens to agree.

The first version of this did exactly that, and the existing
`test_a_hash_that_was_never_pinned_is_the_finding` caught it. The check now
lives only in `disk_status`, and a new test asserts the separation so it cannot
drift back.

9 tests. Suite: 2035 passed, 4 skipped.

### DIFF
--- a/src/data_sheets_schema/cli/api.py
+++ b/src/data_sheets_schema/cli/api.py
@@ -440,15 +440,20 @@ def prompts_check_cmd(strict):
     # record may legitimately name a prompt the registry does not cover; the
     # working tree has no such excuse. A condition prompt nobody pinned is
     # text that was never declared, which is the hole (#432), not a gap in it.
-    bad = [r for r in rows if r["status"] != pr.CANONICAL]
+    # `annotated` is reported and not counted against the gate: the
+    # instruction is at its pin and only rationale moved (#560).
+    bad = [r for r in rows if r["status"] not in (pr.CANONICAL, pr.ANNOTATED)]
+    annotated = [r for r in rows if r["status"] == pr.ANNOTATED]
     for r in rows:
-        mark = {"canonical": "✓", "superseded": "⚠️ ",
+        mark = {"canonical": "✓", "superseded": "⚠️ ", "annotated": "ⓘ ",
                 "uncanonical": "❌", "unpinned": "❌",
                 "missing": "❌"}[r["status"]]
         click.echo(f" {mark} {r['status']:12} {r['path']}")
         if r["reason"]:
             click.echo(f"       {r['reason']}")
-    click.echo(f"\n{len(rows)} prompt file(s), {len(bad)} not at their pin")
+    click.echo(f"\n{len(rows)} prompt file(s), {len(bad)} not at their pin"
+               + (f", {len(annotated)} annotated (body unchanged)"
+                  if annotated else ""))
     if bad:
         click.echo("Declare them with `d4d api prompts pin --file <path> "
                    "--reason '<why this is the condition's text>'`. Editing a "
--- a/src/data_sheets_schema/prompt_registry.py
+++ b/src/data_sheets_schema/prompt_registry.py
@@ -57,6 +57,30 @@ UNPINNED = "unpinned"
 # record named it and hashed nothing. Distinct from `unpinned`, which is an
 # absence of evidence — this is evidence of an absence (#437).
 MISSING = "missing"
+# The instruction is at its pin; something else in the file is not. A prompt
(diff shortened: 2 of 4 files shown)


realmarcin:
Review round complete; #572 filed and the finding fixed here.

`body_sha256_of` and `api_runner.prompt_body` are two implementations of the same split, in separate modules because `prompt_registry` cannot import `api_runner` without a cycle. If they drift the pin would vouch for text the model never receives — the inverse of what this PR is for, and one piece of content in two places is this repository's recurring defect (#518, #521, #545, #563).

Merging them would mean threading a callback through to dodge the cycle, which is worse than the duplication. `TheTwoBodyExtractorsAgreeTest` asserts they hash identical text for every condition prompt instead, so drift fails a test.

Suite: 2036 passed, 4 skipped.

Issue #400, opened 2026-08-07
The agentic path captures no reasoning measure, so Claude Code runs cannot be compared with API runs on reasoning spend


realmarcin:
`{PROJECT}_reasoning.jsonl` is written in-process by `d4d api run`, which can read thinking blocks and token counts off the response (`src/data_sheets_schema/reasoning.py`). A Claude Code subagent has no access to its own token accounting, so the agentic path writes no reasoning log at all.

Consequence: for an agentic run we can record the reasoning-effort *setting* — `CLAUDE_EFFORT` is exposed in the environment, so it is genuinely observed rather than asserted — but there is no `reasoning_tokens_estimate` and no per-phase record of whether a thinking block was returned. Any comparison of reasoning spend between the API arm and the Claude Code arm is therefore one-sided.

This is a real limitation of the runtime, not a bug to fix in this repo, and the right response is probably to state it rather than to synthesize a substitute. Writing a `reasoning.jsonl` for agentic runs that carried the effort level but no measurements would be worse than writing none: `d4d provenance reasoning` would then report a log that looks comparable to the API path's and is not.

What is worth doing:

1. Have `d4d provenance reasoning` distinguish "no log because this runtime cannot produce one" from "no log because reasoning capture postdates this run". It currently prints a single message covering both, and the CLAUDE.md note already makes the analogous distinction for `reasoning_present` vs `reasoning_available` under CBORG.
2. Record the limitation wherever arms are compared, so a missing reasoning figure for the agentic arm is not read as zero.

Noted while adding reasoning-effort capture to the agentic path for label `2026-08-07_claude-opus-5-claudecode-generic-v3_rep1`; see #397.

PR #571, opened 2026-08-16
Let an agentic run record which phases it performed (#562)

### LINKED ISSUE #562: The provenance record describes the API arm's steps and the agentic arm's not at all
Asked directly: does the generation record say what steps or rounds a run went
through, for both arms? For the API arm, in detail. For the agentic arm, nothing.

## What an API record carries

`api_usage` is a list with one entry per phase:

```yaml
- phase: full            attempt: 1  seconds: 390.3  output_tokens: 34151  stop_reason: end_turn
- phase: core            attempt: 1  seconds: 146.3  output_tokens: 16624
- phase: audit           attempt: 1  seconds:  55.0  output_tokens:  4324
- phase: reconcile_full  attempt: 1  seconds: 172.7  output_tokens: 18127
- phase: reconcile_core  attempt: 1  seconds: 140.3  output_tokens: 15788
- phase: report          attempt: 1  seconds:  85.6  output_tokens:  6854
- phase: repair_full     attempt: 1  seconds: 152.1  output_tokens: 17151
- phase: repair_core     attempt: 1  seconds: 149.1  output_tokens: 16681
```

plus `repair` (round count and what each round was given), `phases_skipped`,
and `intermediates`.

## What an agentic record carries

None of those keys exist. The agentic playbook has four named phases and a
non-skippable validate-and-iterate step, and the record says nothing about
which ran, in what order, how many iterations the validate step took, or
whether any phase was skipped. That information exists only as prose in
`{PROJECT}_reconciliation.md`, which #546 has just shown cannot be trusted as
an account of what happened.

## This is not the #400 limitation

#400 established that a Claude Code subagent cannot report its own token
accounting, so a reasoning log for the agentic path would be worse than none.
That is a real runtime limit and it covers `seconds`, `output_tokens` and
`stop_reason`.

**It does not cover phase structure.** Which phase is running, whether it
completed, and how many validate-and-iterate rounds it took are all things the
agent knows and could write. The playbook already tells it to write them into
the reconciliation report in prose.

## Why it matters now

Every arm comparison made this week has been one-sided in a way that is easy to
miss. #544's whole finding was about a phase — reconciliation — whose API
implementation reports success without achieving it. The equivalent question
cannot be asked of the agentic arm from the record, only from prose.

## Suggested

A `phases` block the agentic path writes, with the fields it can honestly
supply: phase name, ordinal, completed yes/no, iterations where a phase loops,
and the artifacts it produced. Omit timing and tokens, and say why they are
omitted rather than leaving them absent — the same three-way distinction
`d4d provenance reasoning` already draws between "cannot capture", "predates
capture" and "missing".

realmarcin:
Closes #562.

The API path writes eight `api_usage` entries per run — phase, attempt,
seconds, tokens, stop reason. The agentic path wrote **nothing**, so its phase
structure existed only as prose in the reconciliation report, and #546 showed a
report is not a reliable account of what happened.

That made every arm comparison one-sided in a way easy to miss. #544's entire
finding was about a phase — reconciliation reporting success without achieving
it — and the same question could not be asked of the other arm from the record.

## This is not the #400 limitation

A Claude Code subagent cannot report its own token accounting, which is why no
reasoning log is written for it. But **which phase ran, whether it completed,
and how many times a loop iterated** are things the agent knows — the playbook
already tells it to write them into the report in prose.

So timing and tokens are **named as unavailable** rather than left absent. A
named limit reads as a fact about the instrument; an absent field reads as an
oversight. Same three-way distinction `d4d provenance reasoning` already draws.

## Shape

```bash
d4d provenance record ... \
  --phase '{"name":"generate_full","completed":true,"artifacts":["CHORUS_d4d.yaml"]}' \
  --phase source_audit \
  --phase '{"name":"reconcile","completed":true,"iterations":3}'
```

- A bare name, or JSON where completion, iterations or artifacts matter.
- **Malformed values raise** rather than being dropped — a log missing one
  phase is worse than no log, because it reads as a run that skipped a step.
- `iterations` appears only when given; writing `1` on a phase that cannot loop
  would imply the number was measured.
- Phases are never inferred from artifacts on disk. A phase that ran and a
  phase whose output happens to exist are different claims.

## `runs check` reports three states, not two

| state | count today |
|---|---|
| nothing recorded | 72 |
| `api_usage` | 50 |
| `phase_log` | **0** |

Collapsing the first two would suggest the arms are comparable in detail, which
they are not. Calling the agentic path's absent would be false once it starts
writing one. The last figure is what v5 changes.

10 tests. Suite: 2044 passed, 4 skipped.

### DIFF
--- a/.claude/commands/d4d-full-core.md
+++ b/.claude/commands/d4d-full-core.md
@@ -495,8 +495,27 @@ poetry run d4d provenance record \
   --input-bundle {EXACT INPUT BUNDLE PATH} \
   --prompt {EACH PROMPT FILE THE RUN CONSUMED} \
   --prompt-text {THE INSTRUCTION AS SENT} \
-  --reasoning-effort {EFFORT, ONLY IF YOU KNOW IT}
+  --reasoning-effort {EFFORT, ONLY IF YOU KNOW IT} \
+  --phase '{"name":"generate_full","completed":true,"artifacts":["{PROJECT}_d4d.yaml"]}' \
+  --phase '{"name":"generate_core","completed":true,"artifacts":["{PROJECT}_d4d_core.yaml"]}' \
+  --phase '{"name":"source_audit","completed":true}' \
+  --phase '{"name":"reconcile","completed":true,"iterations":{HOW MANY TIMES YOU RAN VALIDATE-AND-ITERATE}}'
 ```
+
+**Pass one `--phase` per phase you actually performed, in order.** The API path
+records eight `api_usage` entries per run; this path recorded nothing, so its
+phase structure existed only as prose in the reconciliation report — and #546
+showed a report is not a reliable account of what happened. Every comparison
+between the two arms was one-sided as a result (#562).
+
+Record what you did, not what this file describes. A phase you skipped is
+omitted; a phase that did not complete gets `"completed": false`. `iterations`
+belongs only on a phase that actually loops — writing `1` on a phase that
+cannot iterate implies the number was measured.
+
+Timing and token counts are not asked for, and must not be estimated: this
+runtime has no access to its own accounting (#400). The record names them as
+unavailable so the gap reads as a limit rather than an oversight.
 
 Writes `{METHOD}_core/{VERSION}/{PROJECT}_provenance.yaml` capturing schema
 md5s, model and runtime identity, input bundle hash, repo commit, software
--- a/src/data_sheets_schema/cli/provenance.py
+++ b/src/data_sheets_schema/cli/provenance.py
@@ -18,6 +18,38 @@ def provenance():
     pass
 
 
+def _parse_phases(specs) -> list[dict]:
+    """`--phase` values into phase dicts, in the order given (#562).
+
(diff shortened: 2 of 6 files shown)


realmarcin:
Review round complete; #572 records the one gap, deliberately not closed here.

Nothing verifies that an agentic run actually passed `--phase`. A run that omits it reports `absent`, indistinguishable from every record written before the feature existed. That is the same class as every other playbook instruction, and I have not made it fatal for the reason none of the other reporters are: a record without a phase log is still usable evidence, and a gate would fail the whole history.

What would close it is a check that a `live` record whose runtime reads playbooks — the flag #545 added — also carries a `phase_log`. That has a real edge case (a run that legitimately performed one phase) and deserves a real `phase_log` to test against, which the first v5 agentic run will produce. Filed rather than built speculatively.

Verified end to end rather than by unit test alone: wrote a live record with three `--phase` values through the real CLI, confirmed the block on disk, and removed the test label afterwards.

Suite: 2044 passed, 4 skipped.

Issue #578, opened 2026-08-16
[v5-P1] Two of the three new checks report a pass where nothing was checked


realmarcin:
From the Codex readiness review, 2026-08-15 (findings 7 and 8). **Both verified
before filing.**

Both are the distinction this corpus has enforced repeatedly — "not established"
is not "fine" — missed inside the checks written to enforce it.

## 1. Grounding passes when there is nothing to ground

```python
check_run(Path('/nope/a.yaml'), Path('/nope/b.yaml'), <real bundle>)
→ {'checked': True, 'counts': {'grounded': 0, 'minted_fragment': 0, 'absent': 0},
   'distinct': {...all zero...}}
```

`check_run` skips a record path that does not exist and returns `checked: True`
with zero counts — which `runs check` reports as fully grounded. The pair check
gets this right (`ran: false` with the missing file named); the grounding check,
written days later, does not.

Also: distinct sets preserve case, so `doi:10.1234/ABC` and `doi:10.1234/abc`
count as two identifiers. DOIs are case-insensitive in the local part by
convention and case-varying in practice.

## 2. Report claims about both records are checked against one

`_target` returns `"either"` for a claim naming no record **and** for a claim
naming both. `"either"` is then read against core alone (`runs.py`), so a report
saying "removed from the full and core records" passes when core removed it and
full did not — which is exactly the direction #566 is about.

Separately, a schema claim about a dotted path is checked using its root slot
only, so "`distributions.bogus` is not declared" is reported as a false claim
because `distributions` exists. That is the checker contradicting a report that
was right.

## Fix

- `check_run`: return `checked: false` with the missing paths named, as
  `pair_consistency` does.
- Normalise identifier case before the distinct set.
- Give `_target` a third value for "both", checked against both records.
- Resolve dotted schema claims against the declared slot at that path, not the
  root.

**Should fix before the run.** These do not corrupt records, but they are the
instruments the v5 result will be read through, and each currently fails toward
"looks fine".

realmarcin:
Fixed in #586 — closed by keyword only for #579, so this stayed open.

- `grounding.check_run` returns `checked: false` naming the missing paths when both records are absent, instead of `checked: true` with three zeroes.
- Identifier case is normalised before counting, so `doi:…/ABC` and `doi:…/abc` are one identifier.
- `_target` has a `both` case, checked against both records. "Removed from the full and core records" no longer passes while the full record holds it.
[...]

Issue #577, opened 2026-08-16
[v5-P1] The v5 plan's measured baselines are wrong in one place and undefined in two


realmarcin:
From the Codex readiness review, 2026-08-15 (finding 10). **Verified before
filing — the review is right and I was wrong.** These are baselines I recorded
in `notes/generic_v5_analysis_plan.md` in #567.

## 1. A plain error

The plan says minted fragments are

> 17 / 12 / 14 in AI_READI rep1, CM4AI rep1, VOICE rep3; **0 elsewhere**

Measured now:

```
AI_READI  rep1: 17      CM4AI rep1: 12      VOICE rep3: 14
AI_READI  rep2: 14
AI_READI  rep3: 13
```

**"0 elsewhere" is false** — AI-READI rep2 and rep3 have 14 and 13. The figures
were in my own earlier output and I transcribed only three of five. Prediction 2
of the plan ("minted-fragment counts rise, or hold") is measured against this
baseline, so it would have been evaluated against a number that is 27 too low.

## 2 and 3. Two counts whose definition is unstated

- **Undeclared CURIE prefixes: 370 (mine) vs 320 (Codex).** The difference is
  `urn:` (41) and `ark:` (9). I counted them as undeclared prefixes; Codex
  classifies them as no-authority URI schemes, which is what
  `identifiers.NO_AUTHORITY_SCHEMES` does. Both are defensible; the plan states
  neither, so the v5 figure can be compared against the wrong one.
- **British spellings: 627 (mine) vs 618 in generated prose / 626 including
  quoted (Codex).** Mine counted across full and core text without separating
  quoted material, which the American-English rule explicitly exempts.

## Fix

Restate all three with the counting rule written down, and derive them from a
committed script rather than an ad-hoc pass so the v5 figure is produced the
same way. The other baselines reproduced exactly — absent identifiers 19/10,
URL-valued slots 397, pair divergence 11/12 — which is why the three that did
not are worth fixing rather than distrusting the lot.

**Not blocking for generation; blocking for the plan being usable afterwards.**
A prediction measured against a wrong baseline is worse than no prediction.

realmarcin:
Fixed in #586. `scripts/v5_baselines.py` is committed so the v5 figures are produced the same way as v4's, and prints the counting rules beside the numbers.

[...]

Issue #566, opened 2026-08-15
Facts discovered in the core phase cannot reach the full record: reconcile_full never sees core


realmarcin:
Found analysing the v4 arm before v5. This is what #550's content divergences
turn out to be, and it is worse than a consistency defect.

## The full record is missing facts the core record has

AI-READI rep3's **core** record states, and its **full** record does not:

> Version 3.0.0 ships a recommended split into training, validation and testing
> partitions in proportions of 70%, 15% and 15% (1,576 / 352 / 352
> participants), distributed in the participants.tsv file.

> …the project team prepared the split itself so that the validation and test
> sets are balanced as far as possible for sex, race/ethnicity and diabetes
> status, with mean age also matched across partitions (train 61.1 +/- 11.3
> years; validation 60.3 +/- 10.8; test 60.5 +/- 11.4).

> Participants received a 200 USD stipend for the study visit, paid through NIH
> award OT2OD032644 once the final data were received and the study devices were
> returned…

> …peripheral blood mononuclear cells for future genomics, proteomics,
> metabolomics, gene expression and immunological studies. Biospecimens are not
> part of this dataset…

CHORUS rep1's core record carries a `sampling_strategies` entry for the holdout
test set — *the dataset's most distinctive structural feature*, called out in
the NIH abstract — that its full record does not have.

**The full record is the comprehensive one.** A reader of it never sees the
recommended split.

## Scale

**10 of the 12 v4 records, 25,417 characters** of core-only prose:

| record | core-only chars |
|---|---|
| CM4AI rep2 | 5,571 |
| AI_READI rep3 | 4,386 |
| AI_READI rep1 | 3,707 |
| VOICE rep1 | 2,532 |
| CM4AI rep1 | 2,435 |
| VOICE rep3 | 1,841 |
| VOICE rep2 | 1,547 |
| AI_READI rep2 | 1,430 |
| CHORUS rep1 | 1,159 |
| CM4AI rep3 | 809 |

Classifying all 66 content divergences: 28 paraphrase, **19 where full is a
truncation of core**, 17 materially different, 2 where core truncates full, 1
whitespace.

## The mechanism, from `PHASE_INPUTS`

```python
"core":            ("Completed full record",),
"audit":           ("Completed full record", "Completed core record"),
"reconcile_full":  ("Completed full record", "Audit findings"),
"reconcile_core":  ("Reconciled full record", "Completed core record", "Audit findings"),
```

[...]

Issue #565, opened 2026-08-15
Review of #563: the playbook's new check snippets crashed on the path they will most often hit


realmarcin:
Filed from the review round on #564; **fixed there**.

The point of #563's third part is that the agentic phase 4 should run the same
checks the API path runs. I wrote the commands into the playbook and verified
them against CHORUS — which has **zero external identifiers**, so the grounding
snippet printed `{grounded: 0, minted_fragment: 0, absent: 0}` and looked fine.

Running it against a record that actually has findings showed two problems.

## 1. `KeyError` when the check declines to run

```python
r = check_run(...)
print(r['distinct'])          # KeyError: 'distinct'
```

`check_run` returns `{"checked": False, "reason": ...}` when the bundle is
absent — which is not an edge case here. It is what happens for a
project whose bundle path moved, and 59 records already name a bundle whose
bytes have changed. An agent following a playbook instruction that raises will
improvise, and the improvisation is unreviewable.

Now branches on `checked` and prints the reason.

## 2. 78 lines for 19 facts

The snippet printed every finding. VOICE rep1 has 19 ungrounded identifiers
across 78 occurrences, because each appears in both records and in several
slots. A phase-4 step that emits 78 lines invites exactly the response #546
documents — an audit whose output nobody reads carefully.

Now prints one line per distinct `(kind, identifier)`: 26 for VOICE rep1, and
the counts line already carries the occurrence figure.

## The general point

A command written into a playbook is code that runs unattended, and it was
tested on the one project where it could not fail. Verifying it against CHORUS
proved only that it parsed.

realmarcin:
Fixed in #564.

PR #586, opened 2026-08-16
Make the canary able to stop the sweep, and fix what it reads (#579, #578, #577)

### LINKED ISSUE #579: [v5-P1] A batch reports success while the checks v5 exists to measure are failing
From the Codex readiness review, 2026-08-15 (finding 13). **Verified.**

`d4d api batch` counts a run as succeeded on individual schema validation alone.
A run whose full/core pair diverges, whose reconciliation report contradicts its
own record, or which carries identifiers absent from its bundle is counted as a
success.

That is how the v4 arm passed: 12 records, all valid, `runs check --strict`
exit 0, and 11 divergent pairs plus 29 ungrounded identifiers inside them.

## Why it matters more for v5 than it did for v4

The canary rule exists so one unit is verified before the batch fans out. But
the canary's verdict is the batch's verdict, and the batch does not look at the
three checks. **So the canary can pass while exhibiting exactly the defects v5
was built to fix**, and the sweep proceeds to spend the remaining runs.

This is the same shape as #565 — verifying against the one project that could
not fail — except here the instrument is blind rather than the sample.

## Fix

Not "fail the batch": the checks are deliberately non-fatal, and a divergent
pair is still usable evidence. What is needed is that the **canary** is held to
a higher bar than the batch:

- report the three check outcomes per run in the batch summary, so a sweep
  cannot look clean while they fail;
- have the canary step assert them explicitly, and refuse to fan out on a
  regression against the v4 baseline.

The baselines for that comparison are in `notes/generic_v5_analysis_plan.md`
(and #577 corrects three of them).

**Should fix before the run** — this is the difference between a canary and a
first run.

realmarcin:
Closes #579, #578, #577 — the three remaining findings from the Codex readiness
review, in the order they matter for a v5 run.

## #579 — the canary could not stop anything

`d4d api batch` counted success on schema validation alone. A run whose pair
diverges, whose report contradicts its own record, or which carries identifiers
absent from its bundle entered the "succeeded" column. **That is how the v4 arm
swept clean**: twelve valid records, `runs check --strict` exit 0, eleven
divergent pairs and twenty-nine ungrounded identifiers inside them.

The canary rule verifies one unit before fanning out — but the canary's verdict
*was* the batch's verdict, and the batch never looked at the three checks.

- `execute()` returns the check blocks; the batch prints them for every run.
- `--canary-baseline <prefix>` holds the first run against a previous arm and
  refuses to fan out on a regression.

**A regression gate, not a perfection gate.** Requiring zero pair errors would
refuse to start, since eleven of twelve v4 records have some. The bar is the
*worst* value each metric took across the baseline arm for that project:

```
AI_READI   pair errors 10   report findings 2   ungrounded 0
CHORUS     pair errors  6   report findings 0   ungrounded 0
VOICE      pair errors  9   report findings 2   ungrounded 19
```

VOICE's 19 ungrounded identifiers are in its own baseline, so a v5 VOICE run is
not failed for inheriting them — only for exceeding. A v4 record replayed
against its own arm returns `ok`, which is the test that the gate is satisfiable
by the arm that defined it rather than a wish.

A check that **could not run** also stops the sweep: a canary whose instruments
were blind has verified nothing — #565 one level down.

## #578 — two checks passed where nothing was checked

| defect | now |
|---|---|
| `check_run` returned `checked: true` + three zeroes when both records were absent | `checked: false`, naming them |
| identifier case preserved, so `doi:…/ABC` and `doi:…/abc` were two identifiers | normalised before counting |
| `_target` returned `either` for a claim naming *both* records, read against core alone | `both` is its own case, checked against both |
| a dotted schema claim resolved to its root — "`distributions.bogus` is not declared" reported false because `distributions` exists | skipped rather than guessed; checking it needs the parent's range class |

The third mattered most: "removed from the full and core records" passed while
the full record still held it — the direction #566 is about.

## #577 — three baselines wrong or undefined

`scripts/v5_baselines.py` is committed so v5's figures are produced the same way
as v4's, and it prints the counting rules beside the numbers.

The plan claimed minted fragments were "0 elsewhere" outside three records.
**AI-READI rep2 and rep3 have 14 and 13** — prediction 2 would have been
evaluated against a baseline 27 too low. The undeclared-prefix count (370 with
`urn:`/`ark:`, 320 without) and the British-spelling count (613 in generated
prose, ~626 including quoted) each had a defensible rule that was never stated.

23 new tests. Suite: 2092 passed, 4 skipped.

### DIFF
--- a/notes/generic_v5_analysis_plan.md
+++ b/notes/generic_v5_analysis_plan.md
@@ -94,37 +94,58 @@ Stated so they can be wrong.
 
 ## Measured v4 baselines
 
-Taken 2026-08-15 from the 12 records of `2026-08-13_claude-opus-5-api-generic-v4`,
-before any v5 run. Predictions above are unfalsifiable without them.
+Produced by `poetry run python scripts/v5_baselines.py`, committed so the v5
+figures are produced the same way. **Three of these were wrong or undefined in
+the first version** (#577), which is why the script exists rather than an ad-hoc
+pass.
 
-| quantity | v4 baseline | which prediction |
-|---|---|---|
-| ungrounded external identifiers (distinct) | 19 in VOICE rep1, 10 in CM4AI rep3, 0 in the other ten | 1 |
-| minted fragments on an attested base (distinct) | 17 / 12 / 14 in AI_READI rep1, CM4AI rep1, VOICE rep3; 0 elsewhere | 2 |
-| person fragments on an organisational ROR (distinct) | 7 in VOICE rep1, 0 elsewhere | 3 |
-| undeclared CURIE prefixes (occurrences) | **370** — `chorus:` 226, `cm4ai:` 86, `urn:` 41, `ark:` 9, `nih:` 8 | 4 |
-| identifier slots holding a URL | **397** — `b2ai-voice.org` 302, `chorus4ai.org` 38, `reporter.nih.gov` 18, `fairhub.io` 16 | 4 |
-| British spellings in record prose | **627** — `licence` 199, `analyse` 85, `organisation` 81, `enrolment` 71, `programme` 67 | 5 |
-| full/core pairs diverging on content | 11 of 12 | regression watch |
+| project | rep | grounded | minted | absent | org-frag |
+|---|---|---|---|---|---|
+| AI_READI | 1 | 27 | 17 | 0 | 0 |
+| AI_READI | 2 | 10 | **14** | 0 | 0 |
+| AI_READI | 3 | 10 | **13** | 0 | 0 |
+| CHORUS | 1–3 | 0 | 0 | 0 | 0 |
+| CM4AI | 1 | 9 | 12 | 0 | 0 |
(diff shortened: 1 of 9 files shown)


realmarcin:
Review round complete; **#587 filed and fixed here** — and it was the gate not working at all.

The refusal was raised inside the per-run `try`, whose handler is a bare `except Exception`. So the sweep caught its own canary refusal: under `--continue-on-error` it logged the refusal as a failed run and **fanned out anyway**, which is the single thing this PR exists to prevent. Without that flag it stopped, but reported a failed run rather than a refusal, and released the run lock twice.

A gate that exists and does nothing is worse than no gate, because the summary then reports a canary that passed.

Now recorded in `canary_stop`, loop breaks, lock released once, raised after the loop. `--continue-on-error` governs individual run failures and no longer overrides the decision to fan out at all. `GateControlFlowTest` asserts all three, the last because a lock outliving its process turns a stopped sweep into a permanently blocked label prefix (#513).

Suite: 2096 passed, 4 skipped.

Issue #572, opened 2026-08-16
Review of #560 and #562: two small gaps, both about a second copy


realmarcin:
Filed from the review rounds on #570 and #571. The first is fixed in #570; the
second is stated rather than fixed, with reasoning.

## 1. The prompt body is extracted by two implementations (#570) — fixed

`prompt_registry.body_sha256_of` splits on `## Prompt body` and strips, and so
does `api_runner.prompt_body`. They live in separate modules because
`prompt_registry` cannot import `api_runner` without a cycle.

If they ever drift, **the pin would vouch for text the model never receives** —
the exact inverse of what #560 is for. And this repository's recurring defect
is one piece of content maintained in two places: #518, #521, #545, #563.

Not merged into one function, because the cycle is real and threading a
callback through would be worse than the duplication. Instead
`TheTwoBodyExtractorsAgreeTest` asserts they hash identical text for every
condition prompt, so drift fails a test rather than silently weakening a gate.

## 2. Nothing verifies an agentic run passed `--phase` (#562) — not fixed

The playbook now instructs the agent to record its phases, and `d4d runs check`
reports the three states. But a run that simply omits `--phase` reports
`absent`, indistinguishable from every record written before the feature
existed.

That is the same class as every other playbook instruction, and I have not made
it fatal, for the reason the other reporters are not fatal: a record without a
phase log is still usable evidence, and a gate would fail the entire history.

**What would close it** is a check that a `live` record whose runtime reads
playbooks — the flag #545 added — also carries a `phase_log`, reported the way
bundle drift is. That is a real check with a real edge (a run that legitimately
performed one phase), and it belongs after the first v5 agentic run has
produced a `phase_log` to test it against. Filing so it is not forgotten rather
than building it speculatively.

realmarcin:
Both halves are now closed.

The duplicated body extractor was pinned by `TheTwoBodyExtractorsAgreeTest` in #570 — merging the two would need a callback threaded through a real import cycle, so drift fails a test instead.

The unverified phase log is fixed in #588: `phase_log_status` now distinguishes `missing` — an agentic `live` record written after the feature landed with no `phase_log` — from `absent`, a record that predates it. Scoped by `record_mode`, the #545 `playbooks.consumed` flag and `PHASE_LOG_SINCE`, so honest history is not filed as a defect. Reported and not fatal, like the other record-level checks.

Issue #568, opened 2026-08-15
The v5 canary must be AI-READI, not CHORUS: #566 pushes the largest request to ~279k tokens


realmarcin:
Found reviewing #567.

## The numbers

`reconcile_full` request size in the v4 arm, input plus cache_read (cached
tokens still occupy the context window):

| project | effective context |
|---|---|
| **AI_READI** | **249,015** |
| VOICE | 200,652 |
| CM4AI | 181,661 |
| CHORUS | 52,080 |

#566 adds the core record to that phase. Core records are ~30k tokens for
AI-READI, ~21k for VOICE, ~19k for CM4AI, ~11k for CHORUS — so the largest
request goes to roughly **279k**.

## Why this matters for the canary

The v4 canary was CHORUS, and CHORUS is a 52k request against AI-READI's 249k —
a factor of five. A CHORUS canary exercises credentials, persistence, the run
lock and every gate, and tells you **nothing** about whether the largest project
fits.

It is the same shape as the #565 finding one step up: I verified the grounding
snippet against the one project that had no identifiers, so it could not fail.
Canarying the cheapest unit tests everything except the thing most likely to
break.

## Also unresolved: what the actual ceiling is

The v4 records name the model as `claude-opus-5` with no `[1m]` suffix, yet
AI-READI's requests ran at 249k and succeeded. So either the route CBORG
resolves has a larger window than the name implies, or cache_read is not
counted the way I am counting it. **The corpus does not record the context
limit that was in force**, which means "will 279k fit" cannot be answered from
the record — only by trying.

## Proposed

1. **The v5 canary is AI-READI**, and it must complete `reconcile_full`
   specifically, not merely start.
2. If it fails on context, the fallback is to give `reconcile_full` the core
   record's *slots that full lacks* rather than the whole record — a diff
   rather than a document. That is more code and should not be built
   speculatively.
3. Record the model's context limit in provenance alongside `max_tokens_by_phase`,
   so the next person can answer this from the record. Small, and separable.

PR #588, opened 2026-08-17
Close the five remaining review findings (#580, #572)

### LINKED ISSUE #580: [v5-P2] Four smaller defects from the Codex review, none blocking
From the Codex readiness review, 2026-08-15 (findings 5, 6, 9, 12, 14). Grouped
because each is small and none blocks a run; split out if any turns out to be
larger than it looks.

## 1. The dry-run planner does not describe the requests that will run

`plan()` carries only the phase-1 placeholder, so its `audit` lacks core, its
`reconcile_full` lacks core and audit, and later phases lack their declared
inputs. `approx_tokens()` also counts cached blocks separately from `messages`,
which already contain them.

So the free check that exists to size a run **cannot see the change #566 made**,
which is the one thing about v5 whose size is in question (#568). The canary
rule says exhaust free checks first; this one currently answers a different
question than it appears to.

## 2. The report phase cannot observe what it reports

`report` receives the audit findings and neither reconciled record, yet is asked
for "what was changed in each record and why". It must reconstruct actions it
never saw. That is a plausible root cause for #546 — reports asserting removals
that did not happen — and #546 fixed the checking rather than the cause.

Also, `report_claims_block` pins only the report's md5, not the records or the
schema the claims were checked against. #426's lesson applied to two of the
three checks and not this one.

## 3. `schema_moved` suppresses more than it should

It is true whenever the full-schema digest differs from the run's, and then
downgrades **every** shared-slot presence mismatch to a warning. It does not
establish that the particular slot post-dates the run. So an unrelated schema
edit suppresses real presence defects in historical pairs.

This is the guard that corrected the corpus-wide figure in #561, so it is doing
real work — it is simply broader than its justification. Narrowing it needs
per-slot introduction history, which nothing records today; that may be the
reason to leave it as is and say so.

## 4. `--project VOICE` also matches VOICE_PEDIATRIC

`backfill-checks` filters with `name.startswith(f"{project}_")`, and
`VOICE_PEDIATRIC_provenance.yaml` starts with `VOICE_`. A scoped `--overwrite`
therefore writes records outside the requested project. Verified.

One-line fix; filed because a scoped write that is not scoped is the kind of
thing that is only noticed after it matters.

realmarcin:
Closes #580 and #572 — everything left from the Codex readiness review.

## The planner did not describe the requests it costs

`plan()` carried a nine-character placeholder after phase 1 and nothing else
ever. So `audit` was costed without core, `reconcile_full` without core or
audit, and `reconcile_core` and `report` with nothing at all — **the free check
that exists to size a run could not see the largest thing in it**, and after
#566 could not see the change whose size was the open question (#568).

Now every phase is costed with the inputs it declares, sized from the most
recent existing records for that project:

```
full             146,100   carried {}
core             171,163   carried {full 111,828}
audit            193,855   carried {full 111,828, core 79,216}
reconcile_full   196,446   carried {full, core, audit 9,902}
reconcile_core   193,445   carried {reconciled full, core, audit}
report           148,456   carried {audit, reconciled full, core}
```

`estimate_basis` names where the sizes came from, and says the totals are a
lower bound when no prior artifacts exist rather than costing carried inputs as
empty.

`approx_tokens` also double-counted: `build_phase` starts the message parts with
`list(cached)`, so the bundle was counted in both `cached_blocks` and `messages`.

## The report phase could not observe what it reports

Asked "what was changed in each record and why" and given only the audit
findings. #546 found **every** record that emitted a `distributions` block
reporting a removal that did not happen — a phase asked to narrate a diff it
cannot see is a plausible cause, and #546 fixed the checking rather than the
cause.

It now receives the reconciled records and is told to check each statement
against them before writing it. The verdict also pins the records and the
schema, not just the report — a claim is checked *against* a record and a slot
inventory (#426, #544).

## `schema_moved` excused every presence mismatch

True whenever the digest differs at all, so an unrelated schema edit suppressed
real defects. Narrowing needs per-slot introduction history, which nothing
recorded — **so this records it.**

`digest_inventory.yaml` maps digest → slot names, keyed by the `Dataset` digest
because that is what a provenance record stores. A mismatch is then excused only
for slots the ledger shows did not exist at the run's digest:

| run digest | slot existed then | verdict |
|---|---|---|
| today's | yes | **error** |
| unrecorded | unknown | warning (previous behaviour) |
| none supplied | — | warning (previous behaviour) |

An unknown digest falls back rather than guessing: inventing an answer would
fail history retroactively. Exact from v5 onward, which is when it matters.

**The two headline results are unchanged** — the 2026-08-11 agentic arm stays
consistent, v4 CHORUS stays at six errors.

## Two smaller ones

- `--project VOICE` matched `VOICE_PEDIATRIC_provenance.yaml` via
  `startswith`, so a scoped `--overwrite` rewrote another project's records.
- **#572**: `phase_log_status` returned `absent` both for a record predating the
  feature and for an agentic run that skipped it. `missing` is now its own
  status, scoped by `record_mode`, the #545 `playbooks.consumed` flag and the
  date the feature landed — so honest history is not filed as a defect, the
  distinction #400 draws for reasoning logs.

17 new tests. Suite: 2113 passed, 4 skipped.


realmarcin:
Review round complete; **#589 filed and fixed here**.

`_carry_sizes` guessed the audit findings at an eighth of the core record, with a comment claiming the audit is not separately archived. It is — `_snapshot` writes `{project}_audit.json` under `intermediate/`, seven of them for AI-READI alone. Measured median is 18,627 against a core median of 83,327, so the real ratio is nearer a quarter and the guess understated every phase carrying it. A planner whose job is sizing a run before spending on it was guessing at a number on disk.

Also checked the context consequence of giving `report` the reconciled records, since #568 made that live:

| | estimate before | after |
|---|---|---|
| AI_READI report | 148,456 | 196,309 |
| AI_READI peak phase | reconcile_full | reconcile_full, 197,929 |

Against a recorded actual peak of 285,113 for that project and a demonstrated 363,261 across the corpus, so no new context risk — the report phase now costs about what `reconcile_full` does.

And confirmed `digest_inventory.yaml` is not swept by anything: `lint-modules` globs `D4D_*.yaml`, `MERGED_SCHEMAS` names two explicit paths, no Python glob targets that directory.

Suite: 2113 passed, 4 skipped.

Issue #573, opened 2026-08-16
[v5-P0] The v5 prompt and the uniform rules contradict each other on URLs


realmarcin:
From the Codex readiness review, 2026-08-15
(`notes/codex_v5_readiness_review_2026-08-15.md`, finding 1 and 2).
**Verified before filing.**

## The contradiction

`src/download/prompts/d4d_generic_arm_prompt_v5.md`, the v5 block:

> Write every identifier as a CURIE — a declared prefix, a colon, and the local
> part — **never as a URL**, a resolver link or a bare IRI.

`.claude/commands/d4d-uniform-rules.md`:

> **Where no prefix is declared, a resolvable URL is the correct answer** and an
> invented prefix is not.

For the no-declared-prefix case these are opposite instructions. **The API path
reads only the prompt body; the agentic path reads both.** So the two runtimes
implement different conditions under one condition name.

This is not hypothetical: the v4 arm has **397 URL-valued identifier slots**,
302 of them `b2ai-voice.org`, precisely the no-declared-prefix case. Under the
playbook they are correct; under the v5 prompt they are defects.

## Why the guard did not catch it

`tests/test_playbook_reach.py` checks that a *phrase* appears in both files —
`"as a CURIE"` is in both, so the test passes. It was built to catch a rule
present in one file and absent from the other, and cannot see two rules that
are both present and disagree.

The same weakness hides a second thing: the playbook illustrates its rule with
`b2ai-voice:` and a FAIRhub DOI, project-specific examples the generic prompt is
forbidden to contain. The agentic arm therefore reads examples drawn from the
projects it is generating for.

## The fix is a decision, not just an edit

v5's rule 3 already gives the answer the playbook predates: where the evidence
supplies no identifier, hang one off an identifier it does supply. That
supersedes "a resolvable URL is the correct answer". So:

1. Rewrite the playbook clause to match v5 — fragment on an attested
   identifier, not a URL.
2. Decide what this means for the 397 existing URL values. They are not
   rewritten (#520), but the v5 baseline in the analysis plan describes them as
   "a baseline for v5's rule 3", which only holds once the two texts agree.
3. Strengthen the parity guard so that agreement is checked, not co-occurrence.
   Semantic equivalence is not mechanically checkable in general, but *this*
   case is: the playbook must not tell the model to write a URL in an
   identifier slot.

**Blocking.** Generating both arms under contradictory rules makes the arm
comparison meaningless for every identifier slot.

Issue #556, opened 2026-08-15
Review of #547: the grounding check counted occurrences and reported them as facts


realmarcin:
Filed from the review round on #555. All three are **fixed in #555**, and the
first one had already produced two wrong numbers in that PR's own description.

## 1. Occurrences reported where distinct identifiers were meant

`counts` tallied every *occurrence* of an identifier, but the PR body, the
`runs check` reporter and a test all read those numbers as counts of
identifiers. Because every identifier appears in both the full and the core
record, everything doubled:

- "a person fragment on an institutional ROR appears **14 times**" — it is
  **7 distinct identifiers**, each in both records. #547 listed 6 examples, so
  the real increment over the issue is one, not eight.
- the per-record table in the PR read `VOICE rep1 … 12 absent`, taken from a
  run made before ROR fragments were captured. With fragments captured it is
  **19**, which is #547's own figure exactly.

Both counts are now recorded and named: occurrences say how much of a record
rests on unattested values, distinct says how many facts are at issue. The
reporter uses distinct.

The corrected table changes nothing about the finding itself — CM4AI rep3 still
carries 10 identifiers absent from a bundle containing exactly one ROR.

## 2. Inconsistent deduplication

`identifier_not_in_bundle` findings were deduplicated per (path, identifier);
`fragment_on_org_identifier` findings were not, so the same identifier at one
path could be reported repeatedly. Now both use the same key.

## 3. Dead parameter

`person_fragment_on_org(path, value)` never used `path`. Removed.

realmarcin:
Fixed in #555.

PR #592, opened 2026-08-17
Restore the CURIE rule's force, and give the canary the metric that missed it (#591)

### LINKED ISSUE #591: [v5] The canary regressed on CURIE form: 45 resolver URLs where a prefix is declared
Found by the v5 canary (AI-READI, `2026-08-16_claude-opus-5-api-generic-v5_rep1`).
The run passed its gate and should not have.

## The regression

`ROR`, `doi` and `ORCID` are declared prefixes. The v5 record puts resolver URLs
in `uriorcurie` `id` slots:

```
https://doi.org/10.60775/fairhub.3
https://ror.org/01yc7t268
https://ror.org/00cvxb145
```

| AI-READI | URL-valued identifier slots |
|---|---|
| v4 rep1 | 12 |
| **v5 canary** | **45** |

All 45 are `id`, `publisher` or `latest_version_doi` — every one `uriorcurie`,
every one on a host whose prefix the schema declares. v5's own first rule says a
CURIE wherever the schema declares a prefix.

## Probable cause: #573 softened the rule while fixing its contradiction

The rule as first written said an identifier is

> **never** as a URL, a resolver link or a bare IRI

#573 rewrote it to resolve a genuine contradiction with the playbook — the
playbook said a URL is correct where no prefix is declared, and `download_url`
and `access_urls` are `uri`-ranged so an unscoped ban was wrong. The replacement
reads

> a CURIE wherever the schema declares a prefix for it … **rather than** as a
> resolver URL … A slot whose declared range is a URL takes a URL, and a URL
> inside prose or a citation is text: leave both exactly as written.

The carve-outs are correct and must stay. The framing is weaker — "rather than"
against "never" — and arrives alongside two exemptions, which plausibly licensed
leaving resolver URLs in identifier slots.

## The canary gate could not see it

The gate compares pair errors, report findings and ungrounded identifiers.
These URLs are **grounded** — `doi.org/10.60775/fairhub.3` is in the bundle — so
the grounding check is right to pass them. Nothing measures "an identifier slot
holding a URL where a CURIE was available", which is what v5's headline rule is
about.

## Fixes

1. Restore the rule's force, keeping #573's scope carve-outs, and re-pin. Free:
   one record names `generic_v5` and it is the canary.
2. Add URL-in-identifier-slot as a fourth canary metric with its v4 baseline.

## Not a regression

British spellings on the same record fell **144 → 8**, and grounding shows 13
minted fragments with 0 absent — the American English rule and rule 3 are both
working.

realmarcin:
Closes #591. Both fixes the canary called for.

## What the canary found

The v5 canary (AI-READI) **passed its gate and should not have**. It put 45 resolver URLs into `uriorcurie` identifier slots — `https://doi.org/…`, `https://ror.org/…` — for prefixes the schema declares, against 12 in v4.

## I caused it in #573

| version | text |
|---|---|
| original | identifiers are "**never** as a URL, a resolver link or a bare IRI" |
| #573 | "a CURIE wherever the schema declares a prefix … **rather than** as a resolver URL" + two exemptions |

#573 was right to remove the unscoped ban — `download_url` and `access_urls` are `uri`-ranged, so it told the model to put a CURIE where the schema requires a URL. But the replacement softened the rule rather than scoping it.

Now **scoped, not softened**: in a slot whose declared range is an identifier, never a resolver URL where the schema declares a prefix — write the CURIE. URL-ranged slots and URLs in prose or citations are exempt entirely.

Pin rotated. Only the canary record names `generic_v5`, and it is superseded.

## The gate could not see it

A resolver URL for a declared prefix **grounds perfectly** — `doi.org/10.60775/fairhub.3` is in the bundle — so the grounding check is right to pass it. What is wrong is form, not evidence, and nothing measured that.

`resolver_urls_in_identifier_slots` is a fourth metric now. Replaying the canary through it:

```
pair errors                        5 vs baseline worst 10
report findings                    0 vs baseline worst  2
ungrounded identifiers             0 vs baseline worst  0
resolver URLs in identifier slots 45 vs baseline worst  0   REGRESSED
```

**The gate now refuses the run it waved through.** It counts only identifier-ranged slots on hosts whose prefix the schema declares — a URL in `download_url` is correct, and one on an undeclared host is exactly what the rule asks for.

## The parity probe was tracking wording

`test_every_shared_rule_reaches_both` probed for the literal `"as a CURIE"`, which this rewrite removes. A probe pinned to wording breaks on every rewrite; it now checks `"write the CURIE"` and, separately, that the surviving ban is scoped to identifier slots.

## What the canary confirms is working

On the same record, British spellings fell **144 to 8**, and grounding shows 13 minted fragments with **0 absent** — the American English rule and the fragment rule are both doing their job. Pair errors 5 against a v4 worst of 10, report findings 0 against 2.

Suite: 2118 passed, 4 skipped.

Generated with Claude Code

### DIFF
--- a/.claude/commands/d4d-uniform-rules.md
+++ b/.claude/commands/d4d-uniform-rules.md
@@ -30,12 +30,14 @@ to every project:
 - There is no target slot count, no expected density, and no expected
   relationship to any other arm or project. Apply your own judgment about what
   the evidence supports.
-- **Write an identifier as a CURIE wherever the schema declares a prefix for
-  it, rather than as a resolver URL.** `ROR:01an7q238`, not
+- **In an identifier slot, never write a resolver URL where the schema declares
+  a prefix — write the CURIE.** `ROR:01an7q238`, not
   `https://ror.org/01an7q238`; `ORCID:0000-0002-…`, not the orcid.org URL;
   `doi:10.13026/…`, not `https://doi.org/…`. Two records naming one thing in
   one form produce one identity; the same thing written as a prefix here and a
-  resolver URL there produces two.
+  resolver URL there produces two. A resolver URL in such a slot is a defect
+  even though it resolves: the v5 canary wrote 45 of them and that is what
+  #591 records.
 
   **This governs slots whose declared range is an identifier** —
   `uriorcurie`. A slot whose declared range is a URL takes a URL: `download_url`
--- a/src/data_sheets_schema/canary.py
+++ b/src/data_sheets_schema/canary.py
@@ -49,6 +49,14 @@ METRICS = (
     ("report findings", "report", lambda b: len(b.get("findings") or [])),
     ("ungrounded identifiers", "grounding",
      lambda b: int((b.get("distinct") or b.get("counts") or {}).get("absent") or 0)),
+    # #591. Invisible to the other three: a resolver URL for a declared prefix
+    # grounds perfectly — `doi.org/10.60775/…` is in the bundle — so what is
+    # wrong with it is form, not evidence. The v5 canary wrote 45 and passed a
+    # gate that measured pair consistency, report claims and grounding, none of
+    # which could see the rule v5 exists to enforce.
+    ("resolver URLs in identifier slots", "grounding",
+     lambda b: sum(1 for f in (b.get("findings") or [])
+                   if f.get("kind") == "resolver_url_in_identifier_slot")),
 )
 
 
--- a/src/data_sheets_schema/canary.py
+++ b/src/data_sheets_schema/canary.py
@@ -54,9 +54,13 @@ METRICS = (
     # wrong with it is form, not evidence. The v5 canary wrote 45 and passed a
     # gate that measured pair consistency, report claims and grounding, none of
     # which could see the rule v5 exists to enforce.
+    # Distinct identifiers, not occurrences (#593). Every identifier appears in
+    # both records of a pair, so an occurrence count is roughly double and reads
+    # as twice the problem — the error #556 records for the grounding counts,
+    # made again here: the canary's 45 is 22 distinct values.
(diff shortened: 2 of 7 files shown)


realmarcin:
Adversarial review done. Two findings, both filed and fixed here, and both mistakes with precedent in this repo.

**#593 — the host map was hardcoded.** Three entries, while the schema's `prefixes:` block declares **38 with an http base**. A resolver URL for any of the other 35 was invisible, and adding a prefix to the schema did not extend the check. Same shape as #340 (hardcoded condition chain missing v3/v4), #467 (four-project literal excluding VOICE_PEDIATRIC) and #563 (rules only `/d4d-full-core` could see). Now derived from `prefix_reference`, longest base first because several prefixes share the w3id.org host.

**#593 — and I reported occurrences as identifiers.** For the canary record: 45 occurrences, **22 distinct**. I wrote 45 in the issue, this PR and the commit message. That is #556 verbatim — every identifier appears in both records, so an occurrence count is roughly double. Findings now dedupe on (path, value); the metric counts distinct values. The gate fires on either reading since the v4 baseline is 0, so it was a reporting defect rather than a false pass.

**#594 filed, not fixed here.** While switching branches I found that `git add -A src tests data .claude` on the *other* branch had swept the v5 canary output — ten files including a 93KB generated record — into a code commit. Nothing lost, but corpus and code should not share a branch, and the symptom looked exactly like losing a billed run.

Re-verified the gate against the real canary after both fixes:

```
pair errors                        5 vs baseline worst 10
report findings                    0 vs baseline worst  2
ungrounded identifiers             0 vs baseline worst  0
resolver URLs in identifier slots 22 vs baseline worst  0   REGRESSED
```

Suite: 2121 passed, 5 skipped.

Issue #594, opened 2026-08-17
Canary artifacts were committed into a code branch by an over-broad git add


realmarcin:
Found while switching branches during the review of #592.

`git add -A src tests data .claude` on the `add-source-priority` branch swept in
the v5 canary output — ten files including the full and core records, the
provenance record, the reasoning log and five phase snapshots under
`intermediate/`.

Nothing is lost: they are in that branch's wip commit, and they reappear on
checkout. But they do not belong in a branch about source priority, and the
symptom was alarming in the moment — switching branches made the canary
artifacts vanish from the working tree, which looks exactly like losing a billed
run.

## Why it matters beyond tidiness

Generated records under `data/d4d_concatenated/` are corpus, not code. Mixing
them into a code branch means:

- a code review diff contains a 93KB generated YAML;
- reverting the code change would revert the corpus record with it;
- the branch cannot be rebased cleanly against another that also touched the
  corpus.

There is precedent for the corpus and the code moving separately — #508 removed
a whole label from git while leaving it on disk, which is why an untracked
orphan directory still exists.

## Fix

Commit corpus and code separately, and stage explicitly rather than with `-A`
over a directory that contains both. Worth considering whether a new run's
output should be committed at all until its arm completes: the v4 arm's records
were committed, the v5 canary's are one superseded run.

realmarcin:
Triage of the open queue on 2026-09-06 (main 42c4a48b, 186 issues read with comments): status: no mixed src+data commit in the last 60; CLAUDE.md carries the #795 data-PR rule but no explicit staging rule. One CLAUDE.md line would close this.

Issue #353, opened 2026-08-06
Request assembly is unversioned in provenance: records before/after #352 carry identical prompt evidence


realmarcin:
Found reviewing #352. Provenance records the prompt *file* hash and the *resolved prompt text* hash (`resolved_prompt_digest`), but nothing records how `build_phase()` assembles the request — part ordering, phase instruction text, carry headers. #352 materially changed the request (instruction moved after the carried artifacts; two instructions reworded) yet a record generated the day before and the day after carry byte-identical prompt evidence.

For sweep interpretation this matters exactly once per assembly change: v3 records generated 2026-08-05 (old assembly, `google/claude-opus-5-high`) vs post-#352 records differ in two unrecorded dimensions. Today the repo commit hash in provenance is the only witness, and it also witnesses a thousand unrelated things.

Possible fix: hash the fully assembled message parts (minus the bundle, already hashed) per phase into the provenance `prompts` block — e.g. `assembly: {phase: sha256}` — or at minimum stamp an assembly version constant bumped on any `build_phase`/`PHASE_INSTRUCTIONS` change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #595, opened 2026-08-17
Let the manifest say which source wins a disagreement (Harry, 2026-08-17)

realmarcin:
Harry's observation, implemented. Closes nothing yet — the tier ordering is a
domain judgement worth his eye before this merges.

## The gap it closes

The uniform rules have always said to represent a disagreement rather than
select a side. A v4 CHORUS record followed that and named what was missing:

> No instance count is asserted because the two sources give different figures
> (over 45,000 and 50,000) for the same released dataset and **the bundle offers
> no basis for preferring one**.

The rule was right. `source_priority` in `source_manifest.yaml` supplies the
basis — declared there rather than in a prompt, for the reason #422 records: a
constraint in the launch text is per-run adaptation nothing inherits and no test
can see.

| tier | source types | what the tier means |
|---|---|---|
| 1 | RO-Crate, structured metadata, data resource | the release describing itself |
| 2 | documentation, license, DUA, IRB | the project describing its release |
| 3 | publication, preprint, white paper | the literature |
| 4 | NIH project page, tutorial | about the project, not the release |
| 5 | historical data release / documentation | retained for provenance |

**A tier is proximity to the release, not trust.** Peer-reviewed work is
authoritative on method and routinely behind the current release on counts,
which is why it sits below the project's own documentation *for this purpose*.
All 14 source types in use are ranked; a source may override with `priority:`.

The CHORUS case resolves to `project_documentation` (tier 2) over
`cohort_2_webinar` (tier 4) — 50,000, with both figures preserved in the caveat.

**Equal tiers do not decide.** Where the strongest tier is shared, priority has
nothing to say and the record represents the disagreement as before.

```
d4d download priority                          # the table
d4d download priority --project CHORUS         # a project's ranking
d4d download priority --project CHORUS --decide a,b
d4d download priority --strict                 # fail on an unranked source_type
```

## Two structural points

**The rule is the fifth in the v5 block, not an edit to the inherited one.** The
disagreement rule is v2's and every version carries it, so changing it in v5
would silently change what v2, v3 and v4 mean. The suite caught the first
attempt to do exactly that.

**Rebuilt on top of #592, not rebased.** The first version of this branch was
cut before #592 and superseded pin `bc239c3a`, so merging it would have reverted
the restored CURIE force. Copying its playbook wholesale *did* revert it once
during the rebuild — caught by checking rather than assuming. The pin now
supersedes `45c6f86f`, which is #592's.

Canary artifacts are left untracked (#594).

14 tests. Suite: 2136 passed, 4 skipped.

Generated with Claude Code

### DIFF
--- a/.claude/commands/d4d-uniform-rules.md
+++ b/.claude/commands/d4d-uniform-rules.md
@@ -24,6 +24,22 @@ to every project:
 - Where the declared bundle contains sources that disagree, represent what the
   evidence states rather than silently selecting one. Do not merge distinct
   entities into a single claim.
+
+- **Where two sources disagree, prefer the one the manifest ranks higher.**
+  State its value, and record in the caveat that the sources disagreed, what
+  each said, and which was preferred. The ranking is `source_priority` in
+  `data/preprocessed/source_manifest.yaml`, lowest tier strongest;
+  `d4d download priority --project X` lists it and
+  `d4d download priority --project X --decide a,b` answers a specific pair.
+
+  **Where the disagreeing sources share the same rank the ranking cannot
+  decide**: represent what the evidence states, as the rule above says. This
+  refines that rule rather than replacing it.
+
+  It exists because a v4 CHORUS record wrote that no instance count was
+  asserted "because the two sources give different figures … and the bundle
+  offers no basis for preferring one". The rule was right; the basis was
+  missing.
 - `Dataset` admits one referent. Choose the one the declared bundle best
   supports, state that choice in the reconciliation report, and hold to it
   consistently across both records.
--- a/data/preprocessed/source_manifest.yaml
+++ b/data/preprocessed/source_manifest.yaml
@@ -192,6 +192,40 @@ scope:
           so the overlap is in the evidence, not in the referent
         express_as: related_datasets
 
+# Which source wins when two of them state different things (Harry, 2026-08-17).
+#
+# Before this the uniform rule said only to represent a disagreement, and a
+# record following it wrote: "No instance count is asserted because the two
+# sources give different figures (over 45,000 and 50,000) for the same released
+# dataset and **the bundle offers no basis for preferring one**." That last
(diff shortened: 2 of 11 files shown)


realmarcin:
Adversarial review done. **One blocking finding, filed as #596 and fixed here** — it would have made the rule inert on the arm that matters.

The rule says *"prefer the one the input manifest ranks higher"*. Inspecting a rendered CHORUS request:

```
source_priority   in the request: False
source_manifest   in the request: True    (only the path, in the bundle header)
ranks higher      in the request: True    (my rule text)
```

**The ranking was never sent.** The model was told to consult a table it does not receive, so it could only ignore the rule or invent a ranking — and an invented one is worse than the disagreement it was meant to settle. And since the agentic path *can* open the manifest, one condition name would have meant two behaviours: the defect #563 and #573 were about, reintroduced by the rule fixing a different one.

`source_ranking_block` now renders the project's ranking into the cached prefix beside the bundle, joining to the `Source ID` and `Source type` each file already carries in its SOURCE METADATA block:

```
# Declared source ranking
  tier 2  project_documentation  (documentation)
  tier 4  nih_reporter_project  (NIH project page)
  tier 4  cohort_2_webinar  (tutorial)
  tier 5  github_organization_overview  (historical documentation)
```

`ASSEMBLY_LAYOUT` and the assembly digest move (`9c2ad4a7` → `d59f8532`) — a block added to every request is a change of method, and #353's digest is what keeps a v5 record distinguishable from one made before it.

Two tests asserted a literal cached-block count of `2` and broke: a test about *what* is cached failing because of *how many*. Both assert content now.

Also considered and deliberately deferred: putting `Priority tier: N` in the bundle's SOURCE METADATA block, which is where the rank really belongs. It changes bundle bytes for every project, so every existing record drifts (#452) and the bundles need rebuilding — not something to fold into the change that introduces the rule. Recorded in #596.

Suite: 2142 passed, 4 skipped.

Issue #552, opened 2026-08-15
The pair check is inert for every record that already exists


realmarcin:
Review finding from #549, deliberately left out of it.

#544 makes `d4d api run` record a full/core pair verdict. Nothing gives that
verdict to the 122 records already on disk, so `runs check` reports:

```
ⓘ  122 record(s) predate the full/core pair check (#544); their pairs are
   unknown, not consistent.
```

That is honest, and it is also the entire corpus.

## Two distinct gaps

**1. The agentic arm ran the check and recorded it nowhere.** The playbook's
phase 4 runs `d4d_pair_consistency` twice — once with `--sync-core`, then again
without it as an independent check — and writes the outcome as prose into
`{PROJECT}_reconciliation.md`. None of it reaches the provenance record. So the
15 records of the 2026-08-11 arm, which are the ones that actually pass, report
`unrecorded` alongside the 12 v4 records that fail. The reporter currently
understates the arm that did the work.

**2. The 12 v4 records can be checked but are not.** Their artifacts are still
on disk and unmodified; `pair_consistency` reproduces the 11-of-12 result on
demand. Nothing writes it down.

## Why not in #549

Backfilling means writing 122 provenance records, each of which carries a
two-line header comment that a `yaml.safe_dump` round-trip strips. That is a
corpus-wide write and deserves its own canary and its own review, not a rider
on the PR that adds the check.

It also needs a decision first: a backfilled verdict is computed today about
today's files, not attested by the run. It must be labelled as such — the same
distinction `backfill-effort` draws between observed and asserted — or it will
later read as something the run established.

## Options

1. `d4d provenance backfill-pairs`, reporting by default and writing only under
   `--execute`, using the existing record writer so the header survives.
2. Have the playbook write the block at phase 4, fixing the agentic path going
   forward but not retroactively.
3. Both — (2) for new runs, (1) once for the existing corpus.

(3) is the only one that makes the reporter meaningful for both arms.

PR #597, opened 2026-08-18
Give the generation record a schema, and let it reference everything else

realmarcin:
Answers "don't we already have metadata output about d4d generation?" — we have
three overlapping artifacts, and the authoritative one had no schema.

| artifact | scope | schematised? | authoritative? |
|---|---|---|---|
| `{PROJECT}_provenance.yaml` | per run, 25 top-level keys | **no** | **yes** |
| `{PROJECT}_reasoning.jsonl` | per attempt | no | yes |
| `data/run_telemetry/{label}.yaml` | per label, 12 LinkML classes | **yes** | **no** (derived) |

In a repository whose purpose is schematising metadata, the only piece with a
LinkML schema was the derived report — and `AttemptTelemetry`'s 13 attributes
are a re-derivation of `api_usage` plus the reasoning sidecar.

## One record, referencing the rest

The record already named 9 of the 13 relevant artifacts: bundle, manifest,
prompts, playbooks, schemas, outputs, phase snapshots. It did **not** name the
reasoning log beside it, the prompt registry its hashes were judged against, the
digest inventory, or the telemetry report derived from it. `companions` adds
those four.

**By reference, not by copy.** A registry is shared across runs and appended to
over time: inlining it would multiply it into every record and still not say
which version was in force. A hash does. And a missing file is recorded with
`present: false` rather than omitted — an absent reasoning log is a fact about
the runtime (#400), not a gap in the record.

## The schema describes today's record

Surveyed across all 195 records before writing a line of it. The shape genuinely
varies — a `live` record has `model`/`system`/`api_usage`, a `derived` record has
`derivation`/`sources` and none of those, older records predate later blocks. So
only what **every** record has is required. Requiring a field an honest record
cannot have would make the schema a fiction.

Blocks still being added to are `AnyBlock` (`class_uri: linkml:Any`): the block
exists and its purpose is stated, its interior is not frozen. `grounding` gained
a finding kind this week — a schema needing an edit before an additive change
could be recorded would be a brake rather than a contract.

**All 195 records validate.** `d4d provenance validate-records --strict` is the
gate, and a record missing a required field is rejected — asserted, or the
schema would be decorative.

## Deliberately not imported by the main schema

It describes the pipeline, not a dataset. Importing it would move the `Dataset`
digest a generation arm is frozen against. Verified unchanged at `44d29023`.

## What this does not do

It does not yet collapse the three artifacts into one file — `reasoning.jsonl`
still sits beside the record, telemetry is still written separately. That is a
format migration, and it belongs after the arm, on records that already
validate against a schema.

9 tests. Suite: 2151 passed, 4 skipped.

Generated with Claude Code

### DIFF
--- a/src/data_sheets_schema/cli/provenance.py
+++ b/src/data_sheets_schema/cli/provenance.py
@@ -655,3 +655,52 @@ def backfill_context(execute, label):
                f"; {skipped} skipped as having no api_usage (agentic runs)")
     if not execute:
         click.echo("Nothing was changed. Re-run with --execute to write.")
+
+
+@provenance.command("validate-records")
+@click.option('--strict', is_flag=True, help='exit 1 if any record fails')
+@click.option('--label', default=None, help='restrict to one run label')
+def validate_records(strict, label):
+    """Validate generation records against their own LinkML schema.
+
+    This repository schematises metadata about datasets, and its own generation
+    metadata was a hand-built dictionary that nothing could check — 25
+    top-level keys and no definition. The one artifact that did have a schema,
+    `d4d_run_telemetry.yaml`, is the derived report rather than the
+    authoritative record.
+
+    The schema is deliberately not imported by `data_sheets_schema.yaml`: it
+    describes the pipeline, not a dataset, so importing it would move the
+    `Dataset` digest a generation arm is frozen against.
+    """
+    import subprocess
+    import sys
+
+    from data_sheets_schema.provenance import CONCAT_DIR
+
+    schema = Path("src/data_sheets_schema/schema/d4d_generation_record.yaml")
+    paths = sorted(CONCAT_DIR.glob("*_core/*/*_provenance.yaml"))
+    if label:
+        paths = [p for p in paths if p.parts[-2] == label]
+    if not paths:
+        click.echo("no records matched")
+        return
+
+    failed = []
+    for p in paths:
+        r = subprocess.run(
+            ["poetry", "run", "linkml-validate", "-s", str(schema),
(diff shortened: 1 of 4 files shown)


realmarcin:
Adversarial review done. No blocking findings; one caveat filed as #598.

**Verified the schema can actually reject a bad record** — a schema that accepts everything is decorative. Seven mutations of a real record, all rejected, control accepted:

```
REJECTED  a bogus record_mode
REJECTED  record_version as a string
REJECTED  record_generated_at as garbage
REJECTED  run missing its project
REJECTED  an unknown top-level key
REJECTED  schema block replaced by a string
REJECTED  replicate as a string
accepted  unchanged (control)
```

**Verified the write path**, not just the reader: `build_record` in both `live` and `reconstructed` mode produces `companions` and the result validates. The corpus-wide pass is 195/195.

**#598 — companion references must not be backfilled.** Every other block added this month was backfilled onto the existing 195 (#552, #568). This one must not be: `prompt_registry.md5` asserts *which version of the pin registry a run was judged against*, and hashing today's into a record from 2026-07-27 would assert it was checked against pins that did not then exist — the registry has rotated four times since. That is #470's error, not #568's arithmetic-over-existing-evidence. Filed so nobody adds `backfill-companions` later without noticing. Older records simply have no `companions` block, the schema makes it optional, and all 195 validate.

**Digest unmoved**: `44d29023`. The schema is not imported by `data_sheets_schema.yaml`, so the arm stays frozen against the same digest.

Suite: 2151 passed, 4 skipped.

Issue #606, opened 2026-08-18
[v5] Readiness tracker #2: what the 2026-08-17 Codex review found


realmarcin:
Second Codex readiness review, saved verbatim at
`notes/codex_v5_readiness_review_2026-08-17.md`.

**Verdict: not ready.** 13 defects — 2 blockers, 8 high, 3 medium — reviewing
the delta since the first review, whose 14 findings were all closed.

Each was verified against the working tree before filing. Where the review's
line references drifted, the finding was still real; nothing was filed on trust.

### P0 — blockers

| | issue | why |
|---|---|---|
| 1 | #599 | a mistyped `--canary-baseline` makes a run with 999 defects in every metric evaluate `ok`; the gate is also opt-in, and a resumed batch cannot re-gate |
| 2 | #600 | `superseded_by` is declared for five sources and the resolver ignores it, so a superseded release ties with the one replacing it |
| 3 | #601 | resume invalidation is one level deep, so core and report can survive a replaced full record |

### P1 — high

| | issue | |
|---|---|---|
| 4 | #602 | the canary measures two of the five preregistered outcome families — the #591 pattern, unfixed in general |
| 5 | #603 | four defects in the instruction texts: provenance-vs-minting contradiction, "two conditions exist", stale "four rules", ranking leaking into single-source arms |
| 6 | #604 | audit findings that name no record; the report still cannot see its diff and repair invalidates it afterwards; flat `--out-dir` records paths that do not exist |

### P2

| | issue | |
|---|---|---|
| 7 | #605 | the record schema is opt-in, mode-blind, and three descriptions contradict the writers |

## What the review confirmed as sound

- the v5 prompt is correctly registered and pinned; all 11 prompt files
  canonical and the pinned commit reproduces the current SHA-256
- no defects found outside the six areas examined

## A pattern worth naming

Three of these are the same defect class as ones already fixed: #599 is
"not established is not fine" missed inside the gate built to enforce it, #603's
"two conditions exist" is derive-don't-restate (#340, #467, #563, #593), and
#605 is a check that exists but is not on the generation path (#582). The
review is finding categories this codebase already knows about, in new places.

realmarcin:
Superseded by the 2026-08-19 readiness review (notes/codex_v5_readiness_review_2026-08-19.md), which dispositioned every item from the 08-17 review (12/14 fixed; the two remainders became #638/#639, both now fixed) — and by the arm itself completing.

Issue #601, opened 2026-08-18
[v5-P0] Resume invalidation is not transitive, so core and report can survive a replaced full record


realmarcin:
From the Codex readiness review of 2026-08-17. **Blocker.**

#575 made a rejected artifact drop its producers *and its direct consumers*.
That is one level, and the dependency graph is deeper:

```
full -> reconcile_full -> "Reconciled full record" -> reconcile_core -> report
```

When the full artifact is rejected, the phases requiring `"Completed full
record"` are dropped. `reconcile_core` requires `"Reconciled full record"` and
`report` requires that too — neither names `"Completed full record"`, so neither
is invalidated. After `reconcile_full` re-runs and produces different bytes,
`reconcile_core` and `report` can stay in `done` and be skipped.

**An interrupted run can therefore ship a core record and a reconciliation
report reconciled against a full record that no longer exists.**

Separately, progress records completed phase *names* and the audit text, but no
hash of the artifacts the audit consumed. An artifact that changes but stays
record-shaped passes `_looks_like_a_record`, so the old audit stays marked
complete with nothing tying it to the pair it audited.

## Fix

- Invalidate the transitive closure, not the direct consumers: walk
  `PHASE_NEEDS` to a fixed point.
- Persist a hash per phase input in the progress file, and invalidate when a
  hash no longer matches — the same reasoning that makes the pair and validation
  verdicts pin their artifacts (#426, #544).

This matters because arms take hours and interruption is normal — #513 is the
precedent, and the resume path exists for exactly that.

realmarcin:
Fixed in #607.

Issue #600, opened 2026-08-18
[v5-P0] Source priority ignores superseded_by, which the manifest already declares


realmarcin:
From the Codex readiness review of 2026-08-17. **Verified. Blocker.**

`decide()` returns no winner for the conflicts the manifest explicitly resolves:

```
AI_READI  dataset_documentation vs dataset_documentation_v3  -> winner=None
AI_READI  fairhub_dataset       vs fairhub_dataset_v3        -> winner=None
```

And the manifest **already says which is current**:

```
AI_READI  dataset_documentation          superseded_by=dataset_documentation_v3
AI_READI  fairhub_dataset                superseded_by=fairhub_dataset_v3
CM4AI     october_2025_dataverse_release superseded_by=june_2026_dataverse_release
VOICE     physionet_3_0_0                superseded_by=physionet_3_1_0
```

Five sources declare supersession. My resolver reads `source_type` and
`priority` and ignores `superseded_by` entirely, so a superseded source ties
with the release that replaced it — and by the equal-tier rule the record then
represents the disagreement instead of resolving it.

This is the exact case the mechanism was built for. Harry's CHORUS example
happened to differ by type (`documentation` vs `tutorial`), so it resolved and I
did not test the version-supersession case, which is the commoner one.

## It also splits the two runtimes

`source_ranking_block` sends tier, id and type — not `superseded_by`,
`captured_at`, or the curation note that says which to prefer. Meanwhile
`.claude/commands/d4d-agent.md` separately tells the agentic path to resolve
conflicts by authority *and recency*. So on these four conflicts the API arm
sees a tie and the agentic arm has a recency rule: same condition, two answers.

## Fix

- `superseded_by` outranks tier: a source explicitly replaced by another loses
  to it, whatever their types.
- Send the basis to both runtimes, not just the tier.
- Reconcile or remove `d4d-agent.md`'s independent recency tie-breaker, so one
  rule governs both paths.

realmarcin:
Fixed in #607.

Issue #599, opened 2026-08-18
[v5-P0] A mistyped canary baseline evaluates as ok, disabling the gate silently


realmarcin:
From the Codex readiness review of 2026-08-17. **Verified. Blocker.**

```
baseline_for('AI_READI', 'TYPO_DOES_NOT_EXIST')
  {'pair errors': None, 'report findings': None,
   'ungrounded identifiers': None, 'resolver URLs in identifier slots': None}

verdict(run_with_999_defects_in_every_metric, that_baseline)
  ok
```

A run with 999 pair errors, 999 report findings, 999 ungrounded identifiers and
999 resolver URLs passes the gate, because the *baseline* is missing rather than
the measurement.

## I wrote the test that encodes this

`test_a_missing_baseline_never_regresses` asserts exactly this behaviour, with
the reasoning: *"No bar is not a failed bar. A first-ever arm has nothing to
compare against and must not be blocked by that."*

That reasoning is right for a genuinely first arm and wrong for everything else,
and the code cannot tell the two apart. The distinction the whole corpus turns
on — "not established" is not "fine" — is the one I failed to draw here, in the
gate whose entire purpose is drawing it.

## Two further ways the gate is weaker than it reads

- `--canary-baseline` defaults to `None`, so **no gate runs unless the operator
  opts in**. The protection is opt-in on the command that spends the money.
- A resumed batch cannot re-gate: `execute()`'s completed-run early return omits
  `checks`, so every metric is `None` and the canary reports `unmeasurable`.

## Fix

- A baseline that was *requested* and matched no records is an error, not an
  absence. Fail with the prefix that matched nothing.
- A metric whose baseline is `None` while others resolved is `unmeasurable`, not
  `ok`.
- Keep "no baseline at all" permissible only when none was asked for, and say so
  in the output rather than passing silently.
- Decide whether `--canary-baseline` should default to the previous arm rather
  than to nothing.

realmarcin:
Fixed in #607.

PR #607, opened 2026-08-18
Fix the three v5 blockers from the 2026-08-17 review (#599, #600, #601)

realmarcin:
The three blockers from the 2026-08-17 Codex readiness review
(`notes/codex_v5_readiness_review_2026-08-17.md`, tracked in #606).

Every finding was verified against the working tree before filing; Codex's line
references drifted in places but each finding was real.

## #599 — a mistyped baseline passed the canary

```
baseline_for('AI_READI', 'TYPO_DOES_NOT_EXIST')  ->  {every metric: None}
verdict(run_with_999_defects_everywhere, that)   ->  ok
```

**I wrote the test that asserted this**, reasoning *"No bar is not a failed bar.
A first-ever arm has nothing to compare against and must not be blocked by
that."* True of an arm with no baseline, false of a typo, and the code could not
tell them apart — "not established is not fine", missed inside the gate built to
enforce it.

`verdict` now takes `baseline_requested`:

| case | verdict |
|---|---|
| baseline asked for, resolved | compared |
| baseline asked for, matched nothing | **unmeasurable** |
| baseline asked for, *partially* resolved | **unmeasurable** |
| no baseline asked for | ok, and says so |

The partial case matters: a gate that passes on the metrics it happens to have
is the #591 hole in another form.

Second half: `execute()`'s completed-run early return omitted `checks`, so every
metric was None on a resumed batch and a sweep interrupted after a **passing**
canary could not fan out under the gate it had already satisfied. The checks are
recomputed there from the bytes on disk — returning a verdict the record
asserted earlier would report a clean bill nobody checked.

## #600 — the manifest already said which source wins

`decide()` ignored `superseded_by`, declared for five sources all along, so a
superseded release *tied* with its replacement and the tie meant "cannot
decide":

| project | conflict | now resolves to |
|---|---|---|
| AI_READI | `dataset_documentation` | `dataset_documentation_v3` |
| AI_READI | `fairhub_dataset` | `fairhub_dataset_v3` |
| CM4AI | october 2025 release | june 2026 release |
| VOICE | `physionet_3_0_0` | `physionet_3_1_0` |

Supersession beats tier — it is a statement about *these two sources*, not about
their kinds — and follows a chain, not just a direct link. Harry's CHORUS case
differed by type, so it resolved, and the commoner version case went untested.

Both runtimes are told: the API ranking block marks a superseded source and says
the mark beats the tier, and `d4d-agent.md` loses its own "authority and
recency" tie-breaker, which had been giving the two paths different answers.

## #601 — invalidation was one level deep

```
full -> reconcile_full -> "Reconciled full record" -> reconcile_core -> report
```

`reconcile_core` and `report` need the *reconciled* record, so neither named a
discarded `Completed full record` and neither was invalidated. After
`reconcile_full` re-ran with different bytes both could stay `done` and be
skipped — **shipping a core record and a report reconciled against a full record
that no longer exists.**

`_dependents_of` walks `PHASE_NEEDS` to a fixed point, deriving what each phase
publishes from `PHASE_ARTIFACT` so a new phase is covered without a second list.
Discarding the full record now invalidates all six phases.

Progress also stores each artifact's md5: `_looks_like_a_record` cannot tell
*which* record it is looking at, so changed bytes now invalidate the work that
depended on them (#426, #544).

Suite: 2163 passed, 4 skipped.

Generated with Claude Code

### DIFF
--- a/.claude/commands/d4d-agent.md
+++ b/.claude/commands/d4d-agent.md
@@ -185,8 +185,13 @@ For each project (AI_READI, CM4AI, VOICE, CHORUS):
 When multiple sources describe the same dataset:
 1. Merge complementary information from all sources
 2. Prefer more detailed and specific information over generic descriptions
-3. Resolve conflicts using authority and recency metadata within the current
-   source bundle
+3. Resolve conflicts by the declared ranking, not by your own reading of
+   authority and recency: `source_priority` in
+   `data/preprocessed/source_manifest.yaml`, and `superseded_by` where a source
+   states which one replaced it. `d4d download priority --project X --decide
+   a,b` answers a specific pair. An independent recency rule here gave the two
+   runtimes different answers on the same conflict (#600); where the ranking
+   cannot decide, represent the disagreement rather than choosing
 4. Never consult an older generated D4D to fill a gap or break a tie
 
 ## Runtime Cases
--- a/notes/codex_v5_readiness_review_2026-08-17.md
+++ b/notes/codex_v5_readiness_review_2026-08-17.md
@@ -1,0 +1,255 @@
+# Codex readiness review #2 for a v5 generation arm — 2026-08-17
+
+Second independent adversarial review, run after the eight PRs that closed the
+findings of the first one (`notes/codex_v5_readiness_review_2026-08-15.md`).
+
+| | |
+|---|---|
+| date | 2026-08-17 |
+| reviewer | Codex (GPT-5.x), via the `codex:rescue` plugin |
+| session id | `01a012ef-79a2-7550-a722-546ef5f626ac` |
+| resume | `codex resume 01a012ef-79a2-7550-a722-546ef5f626ac` |
+| repo state | `main` at `e6b99727` |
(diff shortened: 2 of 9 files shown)


realmarcin:
Adversarial review done. **No new defects** — the first round in a while where the review found nothing to file.

Probed the riskiest piece, the transitive invalidation, for the two ways it could be wrong:

- **over-reach**: discarding core does *not* discard `full` (nothing upstream depends on it), but *does* discard `reconcile_full` — correct, since #566 made that phase consume the core record. A closure that over-reaches throws away good work on a multi-hour arm.
- **non-termination**: the fixed-point walk terminates on a synthetic self-dependency. The graph is derived from `PHASE_NEEDS` and nothing guarantees it acyclic.
- **backward compatibility**: progress files predating this carry no `artifact_md5`, and there are such files on disk from 2026-07-31. `recorded is not None` keeps the old behaviour for them; treating a missing hash as a mismatch would re-run every resumable phase of every older run.

All four properties are now tests rather than things I checked once.

Suite: 2167 passed, 4 skipped.

Issue #604, opened 2026-08-18
[v5-P1] Three defects in the phase dataflow: audit shape, report staleness, flat out-dir


realmarcin:
From the Codex readiness review of 2026-08-17.

## 1. The audit shape check accepts findings that name no record — high

The phase contract asks for a top-level `summary` and findings shaped
`{severity, record, slot, issue}`. `_audit_is_well_formed()` checks neither
`summary` nor `record`, so this passes:

```json
{"findings": [{"severity": "high", "slot": "id", "issue": "wrong"}]}
```

Both reconciliation phases are then told to apply "the findings that concern"
their record — from a finding that does not say which record it concerns. Given
#574 conditions absorption on the audit's verdict, a finding that cannot be
attributed is worse than absent.

## 2. The report still cannot see its own diff, and repair invalidates it — high

#580 gave `report` the reconciled records. It still does not get the
*originals*, so "what was changed in each record" remains unobservable — and
`reconcile_core` overwrites the `Completed core record` carry entry, so the
pre-state is gone by then.

Worse: validator-driven repair runs **after** the report is written and can
rewrite either record, so even the post-state the report saw may be obsolete by
the time the run finishes. `report_claims` then checks the stale report against
the repaired records.

Fix: repair before reporting, and supply both snapshots.

## 3. A flat `--out-dir` run records paths that do not exist — high

`RunSpec` sends artifacts to `out_dir`, but `execute()` calls `build_record()`
without those paths or an adjusted `concat_dir`, so `outputs` and the new
`companions` block are reconstructed under the standard
`data/d4d_concatenated/{method}/{label}` layout. A record written into
`out_dir` can name output and companion files that are elsewhere or absent.

This is the GitHub-assistant layout, and it is the same class of bug as the
declared-bundle defect: a path assumed rather than derived from the spec that
already knows it.

Issue #590, opened 2026-08-17
[v5] Pre-flight: three decisions before the arm starts, none of them defects


realmarcin:
Everything from the readiness review is closed (#581). This records what a
pre-flight sweep found that is a *decision* rather than a defect, so none of it
is silently assumed.

## 1. #457 must not precede v5

Adding a `pattern` to `uriorcurie` is the natural companion to v5's identifier
rules — it would make an invented prefix a validation failure rather than only
a rule violation. **It should still wait.**

Measured across the v4 arm:

| classification | count | share |
|---|---|---|
| `curie_declared` | 739 | 49.1% |
| `uri` | 397 | 26.4% |
| `curie_undeclared` | **320** | **21.2%** |
| `uri_unverified` | 50 | 3.3% |
| bare tokens | **0** | — |

All 320 are in `id` slots, and they are exactly what v5's rule 3 targets: do not
invent a prefix, hang the identifier off one the evidence supplies.

Adding the pattern first would hand 320 validation failures to a repair loop
that is shape-only and evidence-blind (`build_repair` excludes the bundle), so
it would have no way to know what to substitute. It would churn, and any
reduction v5's rules achieve would be invisible underneath it.

**Run v5 first.** If the rules work, the 320 collapse and the pattern becomes
cheap; if they do not, that is worth knowing before enforcing it. This is also
the issue's own conclusion under "Why not just add the pattern".

## 2. VOICE_PEDIATRIC has no v4 baseline

```
AI_READI         pair 10  report 2  ungrounded 0
CHORUS           pair  6  report 0  ungrounded 0
CM4AI            pair  9  report 4  ungrounded 10
VOICE            pair  9  report 2  ungrounded 19
VOICE_PEDIATRIC  none
```

The v4 batch was stopped before reaching it. Consequences, both real:

- the canary gate cannot protect it — a missing baseline never regresses, by
  design, so a VOICE_PEDIATRIC run is ungated;
- it will have no v4 counterpart to compare against afterwards.

Either include it and accept both, or leave it out of v5 and say so in the plan.
Including it is defensible for a production run — the goal is the best records
available, and it currently has none at all.

## 3. Freeze the schema now

`d4d api run` already refuses to start when the merged schema is stale (#582),
but nothing stops the schema being *changed* mid-arm. The v4 plan's rule is
"if the digest moves mid-arm, stop and restart the whole arm", and #576 records
what a digest move between arms costs.

[...]

PR #611, opened 2026-08-19
Make the record schema mode-aware, and check it where records are written (#605)

### LINKED ISSUE #605: [v5-P2] The generation-record schema is opt-in and mode-blind
From the Codex readiness review of 2026-08-17. Two findings about #597.

## 1. Nothing validates a record at write time — high

`ProvenanceRecord.write()` calls `yaml.safe_dump`. Schema validation exists only
as the manual `d4d provenance validate-records`. So a run can declare success
having written a record that does not conform.

That is the shape #582 fixed for the schema digest — a check that existed and
was not on the generation path — and I reproduced it for the record schema in
the same week.

## 2. The schema is blind to record_mode — high

`inputs`, `model`, `prompts`, `validation`, `pair_consistency`,
`report_claims` and `grounding` are optional for *every* mode, so a
`record_mode: live` record validates while omitting its bundle, its prompts, its
model, its validation verdict and all three post-generation checks.

I made them optional because a `derived` record legitimately has none of them —
which is true, and the answer is conditional requirements per mode, not blanket
optionality. As it stands the schema mostly asserts that a record is a mapping
with a few identity fields.

## 3. Descriptions contradict the writers — medium

- `record_type` is documented "Always `d4d_generation_provenance`"; derived
  records write `d4d_derived_provenance`.
- `outputs` is documented "by path and hash"; `_artifact()` deliberately records
  no hash and says why.
- `schema.digest_md5` is documented as the digest "the model was actually sent",
  but the agentic recorder computes it after generation and Claude Code reads
  schema files rather than receiving a digest. The API-sent digest and the
  agent-observed schema identity are different claims sharing a field.

## Fix

Mode-specific requirements (LinkML rules or subclasses), validation invoked from
both write paths, and descriptions corrected to what the writers do.

realmarcin:
Closes #605.

## Two weaknesses in the schema from #597

**It was mode-blind.** Only fields universal across all 195 records were required, so a `live` record validated while omitting its bundle, its model, the machine it ran on, and its validation verdict.

Conditional rules now say what each mode must carry. Every field was **measured across the corpus rather than chosen**:

| mode | n | always carries (beyond the universal set) |
|---|---|---|
| `live` | 158 | `inputs`, `model`, `system`, `validation` |
| `reconstructed` | 33 | the above + `unrecoverable` |
| `derived` | 4 | `derivation`, `sources`, `not_applicable` |

So this tightens the contract and fails nothing retroactively — all 195 records still conform.

**It was checked only by a command someone remembers to run.** Validation lived in `d4d provenance validate-records`, so a run could finish, report success, and leave a non-conforming record behind — the shape of #582. `write()` now validates what it wrote; `execute()` fails the run on a violation. The record is still written first: it is the run's only account of itself.

## The gate found a real gap on its first run

`execute()` writes `report_regenerated_after_repair` when repair rewrites a record after the report was rendered (#604). No record on disk carries it, so validating the corpus could never have reached it — and the schema forbade it. The first real repair-with-regeneration run would have failed at the new gate. Added, plus a test that reads the **writers** rather than the archive.

`RECORD_SCHEMA` now resolves against the package when the repo-relative path does not exist: `record_conformance` returns no findings when it cannot run, which is honest for a missing dependency and would have been a silent pass for a sweep launched from another directory.

## Three descriptions that contradicted their writers

- `record_type` — was "Always `d4d_generation_provenance`"; derived records write `d4d_derived_provenance`. Now enumerated, and tied to `record_mode` by a rule so the two discriminators cannot disagree.
- `outputs` — was "by path and hash"; a generation record deliberately carries no hash (`_artifact`, after 77 records hashed a state their files merely passed through). Integrity lives in `validation.artifacts`.
- `schema.digest_md5` — was "what the model was sent"; true of the API path, which fingerprints as it renders, but the agentic recorder runs after generation. Not the same claim.

## VOICE_PEDIATRIC

Recorded in the v5 plan, closing the last open item in #590. It has **no v4 record**, so no baseline for any of the eight gated metrics — `verdict()` returns `unmeasurable` and stops the sweep, which is the #599 fix working. Running it ungated reproduces the v4 failure; borrowing VOICE's baseline would assert two different datasets should show the same defect counts. The arm is **AI-READI, CHORUS, CM4AI, VOICE**.

## Verification

- 2197 tests pass (the 2 that failed were the new gate catching the missing field, now fixed; `tests/test_download/test_api_runner.py` 132 pass)
- All 195 records conform, via both the new write-path check and `d4d provenance validate-records --strict`
- **Mutation-tested**: dropping the rules block fails exactly 3 of the new tests, so they detect the change rather than restate it
- `d4d api prompts check --strict` clean; `d4d runs check --strict` unchanged

### DIFF
--- a/notes/generic_v5_analysis_plan.md
+++ b/notes/generic_v5_analysis_plan.md
@@ -203,6 +203,27 @@ person answers this from the corpus instead of re-deriving it. The *limit*
 stays null and says why: no route states it and the provider does not return
 it, and a guess would make headroom computable and wrong.
 
+## VOICE_PEDIATRIC is out of the v5 arm (2026-08-18)
+
+Decided rather than deferred, closing the last open item in #590. VOICE_PEDIATRIC
+has **no v4 record**, so it has no baseline for any of the eight gated metrics.
+Under `verdict(..., baseline_requested=True)` a project whose baseline resolves
+to nothing is `unmeasurable` and stops the sweep — the #599 fix, working as
+intended. The alternatives were to run it ungated, or to gate it against another
+project's numbers.
+
+Both are worse than skipping it. An ungated run is the v4 failure exactly: a
+record that enters the "succeeded" column on schema validation alone, with
+nothing able to say whether it regressed. Borrowing VOICE's baseline would
+assert that two datasets of different size and documentation should exhibit the
+same defect counts, which nothing supports.
+
+So the v5 arm is **AI-READI, CHORUS, CM4AI, VOICE** — the four projects with a
+v4 arm to be compared against, which is also the only comparison this plan
+licenses. VOICE_PEDIATRIC is not excluded on quality grounds and needs no
+inference drawn about it; it is a project this design cannot measure, and
+running it anyway would produce a number with no bar beside it.
+
 ## What this plan does not license
 
 Comparing v5 against anything other than v4. Every earlier arm sits at a
--- a/src/data_sheets_schema/api_runner.py
+++ b/src/data_sheets_schema/api_runner.py
@@ -41,7 +41,7 @@ from typing import Any
 
 import yaml
 
-from data_sheets_schema import reasoning, schema_digest
+from data_sheets_schema import provenance, reasoning, schema_digest
 from data_sheets_schema.provenance import (
     DETERMINISTIC_CONFIG,
     load_generation_config,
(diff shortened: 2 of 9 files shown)


realmarcin:
## Adversarial review done; six findings filed, five fixed here

Two of the findings were defects **this PR introduced**, and both are the repo's own recurring error classes:

- **#612 — the live rule required `validation`, which no writer writes.** I justified the required lists as "measured across 195 records rather than chosen". They *are* on all 158 live records — because `d4d runs validate` sweeps the corpus afterwards. `build_record` never emits it. So the rule rejected every freshly written agentic record, and the comment claiming to describe the writers described the archive after a later step touched it. The test now **derives** the lists from `build_record` instead of restating them.
- **#613 — "could not check" read as "conforms".** `record_conformance` returned `[]` for both, so a dependency conflict would have silently disabled the gate for an entire sweep with every run printing a tick. My own docstring claimed the caller "can tell apart by asking" and there was nothing to ask. `check_record` now returns `(violations, failure)` and `execute()` fails on either, with distinct messages.

Also fixed: **#615** (the writer-scan guard saw 14 of 34 fields — the regex could not see either builder's dict literal; now an AST scan, mutation-tested), **#617** (three tests passed vacuously with a dead validator; now behind a positive control that fails rather than skips), **#618** (error message named the wrong path, cwd test asserted existence rather than discrimination, `companions` unmentioned).

**Left open, deliberately:**
- **#614** — `value_presence: PRESENT` is not enforced through the JSON-Schema path, so `model: null` still satisfies the rule. Negative result recorded on the issue. Latent: no record on disk has one. Half-fixing it in code would make the schema and the gate disagree about what is required.
- **#616** — three `curated/` records fall outside the `*_core/` glob every checker uses and do not conform; three of their five violations predate this schema. That is a scope decision about what the checkers should look at, not a fix belonging here. The schema comment and test name now state their scope instead of claiming disk-wide conformance.

**Verification:** 2205 passed, 4 skipped. `d4d provenance validate-records --strict` 195/195. Prompts at pin. Mutation-tested twice — dropping the rules fails 3 tests, adding a field to a builder's dict literal fails the guard.

Issue #630, opened 2026-08-19
[generalize] Tracker: make the D4D pipeline work for datasets outside Bridge2AI


realmarcin:
Tracker for the 2026-08-19 Codex generalizability review, saved verbatim at `notes/codex_generalizability_review_2026-08-19.md`.

**Finding in one line:** the single-dataset API path generalizes; everything around it does not. `d4d api run --project NEW --bundle path --condition generic` works for an arbitrary name — download, preprocessing, concatenation, batch generation, most of evaluation, status, the RO-Crate arms and the Make targets all take their project universe from a hard-coded list.

## Priority and order

| | issue | why this rank |
|---|---|---|
| **P0** | #621 provenance falsely attests the B2AI manifest | a record asserting something untrue, and wrong for B2AI runs with an explicit `--bundle` too |
| **P0** | #622 underscore split misattributes `AI_READI` → project `AI` | corrupts evaluation output **today**, not hypothetically |
| **P1** | #623 make the manifest the project registry | the change that unblocks everything else |
| **P1** | #624 batch bundle mapping; tuned components | the batch path is where an arm actually runs |
| **P2** | #625 ingestion: formats, DOIs, hosts, thresholds | quality ceiling for any new domain |
| **P2** | #626 `VOICE_PEDIATRIC_source_dir` iterated as a project | concrete, small, and blocks the unfiltered download path |
| **P3** | #627 rubrics and presence evaluator are not domain-neutral, no N/A | biases any cross-domain score; read with #177 |
| **P4** | #628 separate the study arms from the pipeline | refactor; depends on #621 and #623 |
| **P5** | #629 legacy extraction paths | decide fix-or-delete; several are already broken |

The two P0s come first because they are **defects**, not missing capability. The rest is real but is the pipeline declining to do something; a false provenance claim and a corrupted evaluation key are the pipeline doing something wrong.

## Sequencing against the v5 arm

**None of this should be done on the v5 path before the arm runs.** v5 is four Bridge2AI projects, where the hard-coded lists are correct and the manifest attestation is true. Changing `RunSpec`, `build_record()` or the prompt registry now would move the assembly digest or the pins and invalidate the frozen condition — and the canary is already stale.

#622 is the exception: it touches evaluation only, never generation, and it is producing wrong output now.

So: #622 now, everything else after the arm, and #621 sequenced deliberately since it changes `build_record()`.

realmarcin:
## Scope decided (2026-08-19)

[...]

Issue #629, opened 2026-08-19
[generalize-P5] Legacy extraction paths are broken and GC-shaped — decide whether to fix or delete


realmarcin:
From the 2026-08-19 Codex generalizability review (§4.9–4.13, §9.5).

These paths are superseded by `d4d api run` and the agentic playbooks, and several are **already broken against the current pipeline**:

- `process_concatenated_d4d.py:275` discovers only `*_concatenated.txt`, but preprocessing now writes `*_preprocessed.txt` — it cannot see its own upstream output. Same mismatch in `project.Makefile:793,850`.
- `d4d_agent_wrapper.py:136` skips every filename containing `_row2`, `_row3` or `_row9` as a presumed duplicate. Any spreadsheet whose only valid source sits on those rows loses it silently.
- `interactive_d4d_wrapper.py` uses an obsolete OntoGPT schema URL and GC-specific keyword/clinical/host relevance tests, so an unknown project is flagged irrelevant. Still invoked by `make extract-d4d-individual-gpt5`.
- Text truncated at 10,000 / 50,000 characters; raw PDFs represented by filename only in two wrappers.
- `_rowN` stripped before grouping, so distinct sources with the same base name collapse to one output.

**The question is whether to fix these at all.** A hard-coded row-number skiplist and a dead schema URL in a path nothing current uses are better deleted than generalized. Deleting is also a claim — that no result depends on them — which should be checked against the corpus first.

**P5**: decide, then act. Do not quietly generalize code that should not exist.

Issue #621, opened 2026-08-19
[generalize-P0] Provenance falsely attests the Bridge2AI source manifest for an external bundle


realmarcin:
From the 2026-08-19 Codex generalizability review (`notes/codex_generalizability_review_2026-08-19.md`, §5.5).

**Verified.** `build_record()` takes no manifest argument; `SOURCE_MANIFEST` is a module global. Whenever `input_verified=True` it records the Bridge2AI manifest's path and md5 — regardless of where the bundle came from:

```
build_record('SOME_OTHER_DATASET', ..., input_bundle=/tmp/external_bundle.txt, input_verified=True)

inputs:
  bundle_path: /tmp/.../external_bundle.txt
  hash_basis: verified identical to the bytes consumed
  source_manifest:
    path: data/preprocessed/source_manifest.yaml     # <- never read for this bundle
    md5: c0028d716155a185d7000ec6ad60352f
```

The record attests that a file was an input to a run that never consulted it. That is the defect class this corpus keeps correcting — #470 (never write a guess, name the gap), #448 (observed vs asserted), #613 (not established is not fine) — and here it is asserted rather than merely unknown.

Also affects `api_runner.source_ranking_block()` and the resolved prompt header, which read the same fixed manifest, so an external project gets no ranking block even if its own manifest declares one. And the crate/Healthsheet arms whose headers say "manifest not used" still enter this branch.

**Fix:** thread `manifest: Path | None` through `RunSpec` and `build_record()`. Where a run has no manifest, record *not applicable with the reason* rather than the repository default — the distinction `not_applicable` already exists for.

**P0** because it makes a false claim in a provenance record, which is worse than a missing capability, and it is wrong for B2AI runs launched with an explicit external `--bundle` too.

PR #631, opened 2026-08-19
Save the 2026-08-19 generalizability review, and stop evaluation guessing identity (#622)

### LINKED ISSUE #622: [generalize-P0] Evaluation recovers project and method by splitting on the first underscore, misattributing AI_READI today
From the 2026-08-19 Codex generalizability review (§6.7, §6.14).

**Verified — this is wrong for the current corpus, not only for hypothetical projects.** `evaluate_d4d_llm.py` splits result keys on the first underscore in three places (`:262`, `:315`, `:328`):

```
AI_READI_claudecode_agent    -> project='AI'      method='READI_claudecode_agent'
VOICE_PEDIATRIC_curated      -> project='VOICE'   method='PEDIATRIC_curated'
MY_DATASET_x                 -> project='MY'      method='DATASET_x'
```

There is no special case in that file: `grep AI_READI` finds only a docstring and a `choices=` list. So every LLM-evaluation CSV/Markdown export attributes AI-READI's scores to a project called `AI`, and VOICE_PEDIATRIC's to `VOICE` — silently merging two datasets the manifest declares distinct (the exact confusion the scope work in #422/#441 exists to prevent).

`scripts/batch_evaluate_rubric10_hybrid.py:400` and `batch_evaluate_rubric20_hybrid.py:873` special-case `AI_READI` alone, which confirms the parse is known to be wrong and was patched for one name.

**Fix:** stop recovering identity from a delimited filename. Carry project and method as separate fields through the result structure. Any published number derived from these exports should be rechecked.

**P0** because it corrupts existing evaluation output.

realmarcin:
Saves the Codex generalizability review verbatim and fixes the one finding that is safe to make before the v5 arm.

Findings filed as #621–#629 under tracker **#630**.

## What the review found

The single-dataset API path generalizes: `d4d api run --project NEW --bundle path --condition generic` works for an arbitrary name. **Everything around it does not** — download, preprocessing, concatenation, batch generation, most of evaluation, status, the RO-Crate arms and the Make targets take their project universe from a hard-coded list, so a dataset cannot be onboarded by configuration alone.

Two findings are **defects** rather than missing capability, both of the "record asserts something untrue" class:

| | |
|---|---|
| #621 | provenance attests the Bridge2AI manifest as an input for bundles that never came from it |
| #622 | evaluation misattributes `AI_READI` to a project called `AI` — **today** |

## What this PR fixes: #622

Evaluation recovered project and method by splitting a composite key on its first underscore:

```
AI_READI_claudecode_agent    -> project='AI'      method='READI_claudecode_agent'
VOICE_PEDIATRIC_curated      -> project='VOICE'   method='PEDIATRIC_curated'
```

The second merges two datasets the manifest declares distinct — what the scope work in #422/#441 exists to prevent. The two hybrid batch scripts special-cased `AI_READI` alone, which is a standing admission the parse was wrong.

**There is no correct split.** A project may contain an underscore and so may a method (`claudecode_agent`), so the key is ambiguous in both directions. Both values are known where the key is built, so they are now carried — in the result dict and in the inventory entry — and read back. A legacy result carrying neither reports the raw key with an empty method: visibly incomplete rather than plausibly wrong.

Verified end to end: the inventory builder now yields `AI_READI_gpt5_concatenated -> project 'AI_READI', method 'gpt5'`.

## Why only this one

It touches evaluation and never generation, so it moves no schema digest and no prompt pin. Every other finding changes `RunSpec`, `build_record()`, the prompt registry or the manifest contract, and doing that now would invalidate the frozen v5 condition — the canary is already stale. Sequencing is recorded on #630.

## Verification

- 2213 passed, 4 skipped
- New `tests/test_evaluation_identity.py` (5 tests), including one that states the old bug as an assertion so the fix cannot be undone quietly
- Inventory builder exercised against the real corpus: 12 concatenated, 8 individual entries, all identities correct

### DIFF
--- a/data/evaluation_llm/rubric10/evaluation_plan.md
+++ b/data/evaluation_llm/rubric10/evaluation_plan.md
@@ -1,12 +1,12 @@
 # Rubric10 Evaluation Plan
 
-**Generated:** 2025-12-07 20:36:44
+**Generated:** 2026-08-19 11:50:31
 
 ## Scope
 
-**Total files to evaluate:** 127
-- Individual files: 111
-- Concatenated files: 16
+**Total files to evaluate:** 76
+- Individual files: 64
+- Concatenated files: 12
 
 ## Evaluation Settings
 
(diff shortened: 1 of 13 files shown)


realmarcin:
## Adversarial review found a regression this PR introduced

The first commit changed the inventory format from `{key: "path"}` to `{key: {...}}` **and did not carry the change to its consumers or to the committed file**. On the corpus as it stands that made the output worse than the bug it fixed:

```
before:  project='AI_READI'                    method='gpt5'
after:   project='AI_READI_gpt5_concatenated'  method=''
```

The old special-cased split got every project in the current corpus right. The new fallback got every one wrong, wrote orphan filenames, and fed the composite key into the `project` column the summarisers read — with nothing printed and exit status 0. The fallback docstring called it "visibly incomplete rather than plausibly wrong". It was neither visible nor incomplete.

Fixed (#632, #633, #634, #635):

- `entry_identity` **refuses** a legacy inventory and names the rebuild command, rather than falling back
- committed inventory regenerated — it was 8 months stale (111/16 vs the actual 64/12), which is also why a hardcoded `/16` denominator had looked right
- `total_individual_files` counted an entry's dict keys once entries became dicts: 32 reported, 64 actual. Fixed in the builder and both hybrids
- the evaluation plan was emitting `paths`, `project`, `method`, `granularity` **as filenames** — and that plan is the artifact handed to the rubric agent
- four more sites re-derived identity from a name (`compare_evaluation_methods.py` plus three filename parsers), all wrong today; each now reads the `project`/`method` **already present in the file it was parsing**
- the guard test was a grep over three hardcoded paths and passed throughout the regression above; replaced with tests that drive the real inventory and both hybrids

**Mutation-tested:** restoring the old inventory format fails exactly the three new tests that would have caught this.

Filed, not fixed: **#636** — three evaluation JSONs have the misparsed identity baked into their *data*. Not a regression (same value, different route), and rewriting evaluation artifacts inside a code change is the over-broad-commit pattern of #594.

**Verification:** 2218 passed, 4 skipped. Inventory resolves end to end: `AI_READI_gpt5_concatenated -> ('AI_READI', 'gpt5')`.

Issue #640, opened 2026-08-19
[v5 BLOCKER] Re-canary: the 2026-08-16 canary ran the pre-fix prompt and its stored verdict is an absence, not a pass


realmarcin:
From the 2026-08-19 readiness review. This is the blocker that gates the arm.

## The canary is stale, and specifically stale in the way that matters

| | 2026-08-16 canary | today |
|---|---|---|
| v5 prompt | `bc239c3a`, 12,167 bytes | `389f7bfe`, 13,237 bytes |
| assembly | `9c2ad4a7`, layout with no source-ranking block | `d59f8532`, layout with it |

## Its stored verdict reads as a pass and is not one

Read from its own recorded blocks:

```
status of the 2026-08-16 canary under today's gate: OK
```

Its recorded `grounding` block carries **no findings of any kind** — the resolver-URL finding did not exist when it was written. Recomputed from the artifacts on disk:

```
RECOMPUTED status: REGRESSED
  resolver URLs in identifier slots: 22 against a baseline worst of 0
```

An absence of measurement that reads like a pass — #447, #599, #613, again.

**The v4 bar is not the same absence.** Recomputed from v4 artifacts: `AI_READI resolver_urls=0`, `CM4AI resolver_urls=0`. So v4 genuinely had none, the canary genuinely had 22, and that is #591 — the regression fixed afterwards by restoring the CURIE rule and re-pinning.

**So: the canary exhibited a regression, the regression was fixed, and nothing has been canaried since.** Precisely the case the canary rule addresses — re-canary after a fix; never fix and fan out in the same step.

## Also changed on the generation path since that canary

Source priority and supersession; a fifth prompt rule and three re-pins; `companions` in every record; the `form` block and three new metrics; transitive hash-bound resume invalidation; stricter audit shape and report regeneration after repair; and **conformance gates that can now raise on both of `execute()`'s exits**.

## Verified good, so the re-canary should be clean

- A record built exactly as `execute()` writes it conforms: `([], None)` — including the repair branch with `report_regenerated_after_repair`. The new gate will not fire spuriously.
- All **seven** gated baselines resolve for all four projects from the v4 label.
- A failing first run stops the sweep after one execution even under `--continue-on-error`, in both the raise and validation-failure paths.
- All five predictions map to metrics that exist and whose v4 values recompute; `scripts/v5_baselines.py` reproduces 370 / 397 / 613.
[...]

Issue #639, opened 2026-08-19
[v5 BLOCKER] The report phase is asked what changed but is never shown the before-state


realmarcin:
From the 2026-08-19 readiness review. Raised by **both** prior reviews (2026-08-15 item 6, 2026-08-17 item 13) and still only partially fixed.

`api_runner.py:544` asks the report phase "what was changed in each record". Its declared inputs at `:578` are audit findings plus the **final** full and core records. Confirmed by constructing the actual assembled request:

```
AUDIT_MARKER          sent=True
ORIGINAL_FULL_MARKER  sent=False      <-- the before-state
FINAL_FULL_MARKER     sent=True
FINAL_CORE_MARKER     sent=True
```

`62c7b01c` fixed reports going stale after validator repair — a different problem. The phase still cannot independently determine what changed, so its change account is reconstruction, and `report_claims` checks only narrow claim forms.

**Two honest options, and the choice matters for the arm:**

1. Give the phase the original full/core snapshots — they exist, `intermediates` already records phase snapshots — and test the assembled request actually carries them.
2. Stop asking it for a before/after account and have it describe the final state only.

Option 2 is cheaper and is not a loss if nothing downstream depends on the change narrative. Option 1 changes `PHASE_NEEDS` and therefore the **assembly digest**, which re-baselines the condition and requires a pin rotation — so it must happen before the canary, not after.

Doing neither means shipping an arm whose reconciliation report asserts a diff it could not observe.

realmarcin:
Fixed in PR #643 (merged as 0e9ff4a3): report phase receives Original full/core records (marker-tested: all five inputs sent); resume path recovers originals from snapshots; instruction rewritten to compare rather than reconstruct. Exercised live by the v5 arm including report regeneration after repair.

Issue #638, opened 2026-08-19
[v5 BLOCKER] The analysis plan describes a different arm than the one that would run


realmarcin:
From the 2026-08-19 readiness review (`notes/codex_v5_readiness_review_2026-08-19.md`). The plan is the pre-registration; if it does not describe the arm, the arm is not pre-registered.

Four discrepancies, all verified:

1. **"Four rules"** at `notes/generic_v5_analysis_plan.md:18` — the executable block has **five**. This was reported by the 2026-08-17 review and marked fixed; it is still present in both the prompt rationale and the plan.
2. **Stale assembly digest** at `:39` — records `2c1442fc`; current is `d59f8532`.
3. **`comparable_conditions('generic_v4','generic_v5')`** stated at `:47` as returning `False`. It returns `True` — the function now assesses prompt adjacency only, with `runs.arm_confounds` handling procedure differences.
4. **Prediction 5 is unmeasurable as written** at `:91`: it requires British spelling to be unchanged *"on the agentic arm"*, and the planned arm is API-only with no agentic-v5 records.

**Item 4 needs a decision, not a fix.** Either narrow prediction 5 to the API arm, or add a defined agentic-v5 arm to the plan. Silently dropping the clause after seeing the results is the thing pre-registration exists to prevent, so it must be decided before generation.

Items 1–3 are corrections to text.

Also worth correcting wherever it appears: the gate has **seven gated metrics** plus one reported-only (`minted fragments`), not eight. `len(METRICS) == 7`, `len(REPORTED_ONLY) == 1`.

realmarcin:
Fixed in PR #643 (merged as 0e9ff4a3): prediction 5 narrowed to the API arm pre-generation with the cost stated; five-rules count corrected everywhere; assembly digest current with a re-check note; comparable_conditions claim corrected.

Issue #613, opened 2026-08-19
`record_conformance` reports 'could not check' as 'conforms', and its docstring claims otherwise


realmarcin:
Found in adversarial review of #611.

`record_conformance` (`src/data_sheets_schema/provenance.py`) wraps both the `linkml.validator` import and the `validate()` call in one `except Exception: return []`. `[]` is also what a conforming record returns. `execute()` reads `if rec.conformance:` — so a dependency conflict, a schema-compilation error, or a transient read failure **silently disables the gate for every run in a sweep**, each printing a tick.

The docstring I wrote says the two are "different things the caller can tell apart by asking." There is nothing to ask: `conformance: list[str]` is the only state `write()` sets. The docstring is false.

This is the repo's own recurring lesson, and it is solved correctly twice within one file:
- `api_runner.py:1300-1308` — "`findings` is None exactly when `failure` says why the validator itself could not run … a record that could not be checked is not a record that passed."
- `canary.py:169,181` — `blind` → `UNMEASURABLE`, "'not established' is not 'fine'."

**Fix:** return `(findings, failure)` or set `conformance_checked: bool`, and have `execute()` treat "could not check" as its own outcome rather than a pass.

`record_schema_path()` closes the missing-schema case only; the remaining exception paths still report clean.

realmarcin:
Fixed in 83000d04 on `close-605-record-schema-modes` (PR #611). Full suite: 2205 passed, 4 skipped; corpus 195/195 conforming.

PR #641, opened 2026-08-19
Save the 2026-08-19 v5 readiness review (NO-GO)

realmarcin:
Third readiness review, saved verbatim. **Verdict: NO-GO.** Blockers filed as #638, #639, #640.

Documentation only — no code change.

## The blocking reason

The 2026-08-16 canary ran a different prompt and a different assembly from what the arm would run today:

| | canary | today |
|---|---|---|
| prompt | `bc239c3a`, 12,167 B | `389f7bfe`, 13,237 B |
| assembly | `9c2ad4a7` (no source-ranking block) | `d59f8532` |

**Its stored verdict reads as a pass and is not one.** From its own recorded blocks it reports `ok`; its `grounding` block carries no findings of any kind, because the resolver-URL finding did not exist when it was written. Recomputed from the artifacts:

```
RECOMPUTED status: REGRESSED
  resolver URLs in identifier slots: 22 against a baseline worst of 0
```

An absence of measurement that reads like a pass — #447, #599, #613, this time inside the canary.

The v4 bar is *not* the same absence, and the asymmetry is sharper than first stated. **All 12** v4 records recompute cleanly and give resolver=0, and all 12 recorded blocks carry `recorded_by: backfill_checks` from 2026-08-18 — after the resolver and form metrics existed. So the gate compares a **stale run-time block against freshly backfilled baselines**. v4 genuinely had none and the canary genuinely had 22 — that is #591, fixed afterwards by restoring the CURIE rule and re-pinning. **The canary exhibited a regression, the regression was fixed, and nothing has been canaried since.**

## Verified — the re-canary should be clean

- A record built exactly as `execute()` writes it conforms, `([], None)`, including the repair branch — the new conformance gate will not fire spuriously on a correct run
- All **seven** gated baselines resolve for all four projects
- A failing first run stops the sweep after one execution even under `--continue-on-error`, in both failure paths
- All five predictions map to existing metrics; `scripts/v5_baselines.py` reproduces 370 / 397 / 613
- Bundles present and matching every v4 replicate; 174 GiB free

**Not verified, and worth naming:** endpoint reachability. `CBORG_API_KEY` is present and `_client()` constructs, but the endpoint was deliberately not called — that section's verdict is UNVERIFIED. The canary rule names missing credentials in a child process as a recurring failure, so the re-canary is what tests it. Lock-file creation, end-to-end resume, and `d4d schema check-digest --strict` are also UNVERIFIED (read-only sandbox).

## Prior reviews

**11 of 14** fixed from 2026-08-15 — item 6 partially, item 11 reframed-not-cleaned, item 12 **not fixed** and now #642. **12 of 14** from 2026-08-17. An earlier version of this PR said 14 of 14, which was wrong and erased #642.

Also corrects a claim of mine: seven gated metrics plus one reported-only, not eight.

### DIFF
--- a/notes/codex_v5_readiness_review_2026-08-19.md
+++ b/notes/codex_v5_readiness_review_2026-08-19.md
@@ -1,0 +1,325 @@
+<!-- Saved verbatim. The review body is evidence of what was found on this
+     date; corrections belong in the issues filed from it. -->
+
+# Codex review: is the repository ready to run the v5 arm?
+
+- **Run:** 2026-08-19, Codex session `01a01ba9-79ca-76a1-af0b-1b282c4a8094`
+- **Verdict: NO-GO.**
+- Third readiness review, after `codex_v5_readiness_review_2026-08-15.md` and
+  `..._2026-08-17.md`. This one was also asked to check whether what those two
+  found was *fixed* or merely *filed*: 14 of 14 and 12 of 14 respectively are
+  fixed, with the report before/after finding partially fixed in both.
+
+## The blocking reason, and a correction to how the review states it
+
+The 2026-08-16 canary cannot authorise this sweep. It ran a different prompt
+(`bc239c3a`, 12,167 bytes) and a different assembly (`9c2ad4a7`, a layout with
+no source-ranking block) from what the arm would run today (`389f7bfe`, 13,237
+bytes; assembly `d59f8532`).
+
+The review reports that canary as `regressed` today. **Read from its own
+recorded blocks it reports `ok`** — verified:
+
+```
+status of the 2026-08-16 canary under today's gate: OK
+```
+
+Both are true and the difference is the point. Its recorded `grounding` block
+carries **no findings of any kind**, because the resolver-URL finding did not
+exist when it was written. Recomputing the same checks from the artifacts on
+disk gives:
+
+```
+RECOMPUTED status: REGRESSED
+  resolver URLs in identifier slots: 22 against a baseline worst of 0
+```
+
+So the canary's stored verdict is not a pass; it is an absence of measurement
+that reads like one — the failure this corpus keeps naming (#447, #599, #613).
+
+The v4 bar it is measured against **is** a real measurement, not the same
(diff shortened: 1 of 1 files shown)


realmarcin:
## Adversarial review: one false claim, in the part I wrote

The Codex body was confirmed verbatim — the only removals are the two trailing `Codex session ID` / `Resume in Codex` lines, and the session ID is preserved in the header, so no provenance is lost. Every empirical claim in the header reproduced: the six prompt/assembly values, the recorded-`ok`-versus-recomputed-`REGRESSED` discrepancy and its stated cause, and the seven-gated-metrics count.

**The defect was my summary of the prior reviews.** I wrote "14 of 14 fixed" for 2026-08-15. Counting the review's own words 285 lines below in the same file: 11 fixed, 1 partially fixed, 1 reframed-not-cleaned, 1 **not fixed**. The sentence was also self-contradicting on its face — "14 of 14 fixed, with the report finding partially fixed in both" cannot hold.

Worse than the arithmetic: it erased an open defect. 2026-08-15 item 12 is genuinely unfixed — `cli/provenance.py:49` records any bare `--phase` value as a completed phase. That matters specifically for the agentic arm, whose runtime cannot attest its own phases (#400), so `phase_log` is the only claim there is. Filed as **#642**.

Two further corrections, both verified independently rather than inherited:

- **The v4 claim was understated.** I had recomputed 2 of 12 records. All 12 recompute cleanly, all 12 give resolver=0, and all 12 recorded blocks carry `recorded_by: backfill_checks` from 2026-08-18 — after the resolver and form metrics existed. So the gate compares a *stale run-time block against freshly backfilled baselines*, which is a sharper statement of the asymmetry than I made.
- **The PR body overclaimed "credentials reachable"** — a row Codex marked UNVERIFIED. `CBORG_API_KEY` is present and `_client()` constructs; the endpoint was deliberately not called. Given the canary rule names missing credentials in a child process as a recurring failure, that was the wrong thing to round up. Corrected, with the other UNVERIFIED rows named.

Issue #644, opened 2026-08-20
[v5 BLOCKER] The v5 CURIE rule inverts the behaviour it was written to enforce — canary 2026-08-19 regressed


realmarcin:
The 2026-08-19 re-canary (AI-READI, from `main` at `0e9ff4a3`) **regressed and the gate stopped the sweep**. No money was spent on the remaining 11 runs.

```
pair errors                       13 vs baseline worst 10   ❌
resolver URLs in identifier slots 24 vs baseline worst 0    ❌
report findings                    1 vs 2
ungrounded identifiers             0 vs 0
organisational fragments           0 vs 0
undeclared prefixes                0 vs 15
British spellings                 14 vs 146
```

## The rule produced the opposite of its intent

Structural comparison of every `id` in the two records:

| arm | ids | resolver URLs | CURIEs |
|---|---:|---:|---:|
| v4 (2026-08-13) | 68 | 6 | **62** |
| v5 (2026-08-19) | 29 | **29** | **0** |

v4: `doi:10.60775/fairhub.3`, `ROR:01yc7t268`.
v5: `https://doi.org/10.60775/fairhub.3`, `https://ror.org/01yc7t268`.

**v4's prompt mentions "resolver URL" or "CURIE" exactly zero times** and produced 62/68 CURIEs. v5 adds five mentions of the rule and produces 0/29. Adding the rule inverted the behaviour it was written to enforce.

## Probable mechanism, and it is checkable

The schema declares these slots as range **`uriorcurie`** — verified: `Dataset.id`, `DatasetProperty.id`, `Organization.id` all `uriorcurie`, and the digest the model is sent renders `## `id` — *uriorcurie* [req]`.

Rule 1 ends with an exemption:

> Two things this does not govern, and they are exempt entirely: **a slot whose declared range is a URL takes a URL**, and a URL inside prose or a citation is text.

A range literally named `uriorcurie` satisfies "declared range is a URL" on a plain reading. So the exemption plausibly swallows every identifier slot the rule was written for, and the model — which defaulted to CURIEs with no rule at all — now reads permission to use URLs everywhere.

Testable before spending again: reword the exemption to name the ranges it means (`uri`, not `uriorcurie`), or state explicitly that `uriorcurie` is governed by the rule rather than exempt from it.

## This is #591 recurring, with a cause

[...]

PR #645, opened 2026-08-20
Name the range the CURIE rule governs: uriorcurie, not "an identifier" (#644)

realmarcin:
Fixes the defect behind the 2026-08-19 canary regression. #644 stays open until a canary passes under this prompt.

## What the canary found

The gate stopped the sweep after one run (~786k in / 250k out instead of twelve runs): resolver URLs **24 vs baseline 0**, pair errors 13 vs 10. Structural comparison showed the rule had **inverted** its own intent:

| arm | ids | resolver URLs | CURIEs |
|---|---:|---:|---:|
| v4 — no CURIE rule at all | 68 | 6 | **62** |
| v5 — with the rule | 29 | **29** | **0** |

## The mechanism, confirmed by the run itself

Rule 1's exemption read *"a slot whose declared range is a URL takes a URL"*. The schema's identifier slots are ranged **`uriorcurie`** — a name that satisfies that clause on a plain reading. The run's own reconciliation report says it read it exactly that way: *"the resolver URL for the `uriorcurie`-ranged `id`"*.

This is a **parity defect**: the agentic playbook's wording never had the hole (*"This governs slots whose declared range is an identifier — `uriorcurie`"*). The API prompt lost that disambiguation when the rule was ported, and #591's fix re-pinned without touching the exemption — so the defect survived a fix aimed at it. The prompt now matches the playbook: the governed range is named literally, and the exemption names `uri` — not `uriorcurie`.

## The test that enforced the bug

`test_playbook_reach`'s AGREEMENTS list **required** the phrase "declared range is a URL" verbatim — a check enforcing the ambiguity. It now accepts the disambiguated wording and adds a standing requirement that both texts name `uriorcurie` explicitly. Mutation-tested: reverting the prompt to the ambiguous exemption fails the guard.

## A second leak the suite caught

My first version of the examples carried `doi:10.60775/…` — **AI-READI's DOI prefix** — into the generic prompt, and `test_no_expected_quantities` failed it. That is the per-GC leakage this arm exists to measure the absence of, and the same class the generalizability review catalogued. Examples are now form-only. Two pin rotations in one branch is honest churn; the superseded chain records the intermediate bytes.

## What the canary settled without a fix

- The 68→29 entity drop is **mostly intended**: the 16 dropped ORCID creators were clinical-trial investigators v4 promoted from a lower-tier source; the release's own citation names the AI-READI Consortium as sole creator — source priority working. The train/val/test split survives in prose.
- The extra pair errors are `notes` divergences from repair rewriting full only — downstream of the identifier mess; re-measure after this fix.

Canary history is now recorded in the analysis plan.

## Verification

2221 passed, 4 skipped. Prompts at pin (`c7fcc228`); prior pins superseded so both canaries still verify. Mutation test on the parity guard.

**After merge: re-canary under a fresh label.** Fix and fan-out stay separate steps.

### DIFF
--- a/.claude/commands/d4d-uniform-rules.md
+++ b/.claude/commands/d4d-uniform-rules.md
@@ -51,8 +51,8 @@ to every project:
   `ORCID:` CURIE, not the orcid.org URL; a `doi:` CURIE, not the doi.org
   resolver form. (Form only, deliberately: an example carrying a real registry
   identifier or a real DOI prefix leaks one project's identity into text every
-  project reads — a PhysioNet DOI prefix, i.e. VOICE's publisher, sat here for
-  a month.) Two records naming one thing in one form produce one identity;
+  project reads — one project's real publisher prefix sat here for a month,
+  #647.) Two records naming one thing in one form produce one identity;
   the same thing written as a prefix here and a resolver URL there produces
   two. A resolver URL in such a slot is a defect even though it resolves: the
   v5 canary wrote 45 of them and that is what #591 records.
--- a/.claude/commands/d4d-uniform-rules.md
+++ b/.claude/commands/d4d-uniform-rules.md
@@ -47,19 +47,26 @@ to every project:
   relationship to any other arm or project. Apply your own judgment about what
   the evidence supports.
 - **In an identifier slot, never write a resolver URL where the schema declares
-  a prefix — write the CURIE.** `ROR:01an7q238`, not
-  `https://ror.org/01an7q238`; `ORCID:0000-0002-…`, not the orcid.org URL;
-  `doi:10.13026/…`, not `https://doi.org/…`. Two records naming one thing in
-  one form produce one identity; the same thing written as a prefix here and a
-  resolver URL there produces two. A resolver URL in such a slot is a defect
-  even though it resolves: the v5 canary wrote 45 of them and that is what
-  #591 records.
+  a prefix — write the CURIE.** A `ROR:` CURIE, not the ror.org URL; an
+  `ORCID:` CURIE, not the orcid.org URL; a `doi:` CURIE, not the doi.org
+  resolver form. (Form only, deliberately: an example carrying a real registry
+  identifier or a real DOI prefix leaks one project's identity into text every
+  project reads — a PhysioNet DOI prefix, i.e. VOICE's publisher, sat here for
+  a month.) Two records naming one thing in one form produce one identity;
+  the same thing written as a prefix here and a resolver URL there produces
+  two. A resolver URL in such a slot is a defect even though it resolves: the
+  v5 canary wrote 45 of them and that is what #591 records.
 
   **This governs slots whose declared range is an identifier** —
-  `uriorcurie`. A slot whose declared range is a URL takes a URL: `download_url`
-  and `access_urls` are declared `uri` and a CURIE there is wrong. A URL inside
-  prose or a citation is text, not an identifier, and must be left exactly as
-  written.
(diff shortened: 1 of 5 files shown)


realmarcin:
## Round-2 review: five findings, four fixed here, two issues filed

**The most important one is a correction to my own claim.** I wrote "Mutation-tested: reverting the prompt to the ambiguous exemption fails the guard." That was false on its stated terms — I had tested a *full* rule-1 revert. An **exemption-only** revert (the exact #644 sentence back in, governed clause intact) passed all 15 tests, because the AGREEMENTS entry still accepted "declared range is a URL" — kept as an alternative *because the playbook itself still used that sentence at line 59*. So "the playbook's wording never had the hole" also overstated: it carried the diagnosed sentence verbatim, mitigated by its neighbours.

Fixed at the root: both texts are now disambiguated, the guard **forbids** the ambiguous sentence in both, and the mutation test now exercises the actual mutation — the exemption-only revert fails.

Also fixed:
- **String-ranged identifier slots were in neither branch of the rule.** `doi`/`orcid` are range `string`; `doi`'s unanchored pattern accepts bare DOI, CURIE and resolver URL. Both texts now say string-ranged slots follow their own description. The durable fix — anchoring the pattern — is a schema change and waits for the arm: **#646**.
- **`ROR:01an7q238` is UC Berkeley's real ROR id**, in text sent verbatim to the model, evading the quantity regex because its digits are flanked by letters. All examples in both texts are now form-only; guard gap filed as **#647**. No leakage into any generated record found.
- **The plan misstated the pair-error mechanism** (repair touched both records, not full-only; 7 of 13 errors in notes paths). Corrected to what was measured.

Verified clean by the review: the rule's schema claims (`download_url`/`access_urls` are `uri`; all seven `uriorcurie` slots suit the rule), the pin chain including the honest intermediate, every number in the canary-history section, and the em-dash mechanics of the parity phrases.

2221 passed, 4 skipped. Pin now `d22c9084`; four rotations on this branch, chain intact.

realmarcin:
## Round 3: one defect in the changed text, fixed; two issues filed; merge-ready

The round-2 fixes all verified empirically — the mutation claim is now true in both texts, the pin chain holds end-to-end (every superseded hash a real git blob), the string-range sentence matches the `doi` slot's own description, and the plan's corrected numbers are exact.

**Fixed here:** the exemption paragraph opened "Two things this does not govern" while listing three — a stale literal count in the exact paragraph #644 proved is read literally, introduced by the round-2 fix itself. Also: the playbook's parenthetical had replaced a leaked DOI prefix with a leaked publisher↔project association ("PhysioNet, i.e. VOICE's publisher") — the same leakage one abstraction up, now dropped; and the plan's "same class" rounded 12+1 into one bucket.

**Filed:** #648 — the parity guard blocks the historical sentence, not the concept: paraphrases ("is a URL, takes a URL", "it takes a URL") evade the blocklist, so an ambiguous sentence *added alongside* the intact rule passes every test. The required-phrase side still fails closed on any rewrite of the canonical exemption, and every body change forces a pin rotation with a human-read reason — the residual is documented rather than closed. #649 — the plan cites the 2026-08-19 canary whose artifacts are untracked; they need a data commit of their own, not a ride on this branch (#594).

**One overstatement of mine corrected for the record:** "all examples in both texts are now form-only" — a real ROR (`ror.org/032db5x82`, University of South Florida) remains at prompt line 28, in the rationale above `## Prompt body`, which is stripped before sending and reaches no model. Model-visible text is clean.

Five pin rotations on this branch; chain intact; both prior canaries still verify. 2221 passed, 4 skipped.

Issue #650, opened 2026-08-20
[v5 BLOCKER] Full/core divergence is systematically higher under v5: both canaries exceed the entire v4 spread


realmarcin:
The 2026-08-20 canary (post-#644-fix, from `main` at `1551df42`) confirmed the CURIE fix completely — **resolver URLs 24 → 0**, and six of seven gated metrics pass:

```
pair errors                       12 vs baseline worst 10   ❌
report findings                    1 vs 2
ungrounded identifiers             0 vs 0
resolver URLs in identifier slots  0 vs 0    (was 24)
organisational fragments           0 vs 0
undeclared prefixes                9 vs 15
British spellings                 11 vs 146  (prediction 5 signal)
```

The gate stopped the sweep on pair errors alone. One run spent (~638k in / 267k out, cache_read 199k).

## Why this looks systematic, not unlucky

AI-READI pair errors per run: **v4 = {10, 8, 6}; v5 = {13, 12}.** Both independent v5 canaries sit above v4's entire spread. Yesterday's hypothesis — that the pair errors were downstream of the identifier mess — is **disproven**: identifiers are now clean and the count barely moved.

## What the 12 divergences are

Shared-slot content differences where **core carries more than full**: `acquisition_methods` full=2 vs core=3, `collection_mechanisms` full=11 vs core=13, plus `description`, `notes`, `source_caveats` divergences and one new `resource-projection-presence` finding on `$.resources`. Core finding facts full lacks is exactly what #566 gave `reconcile_full` the core record to absorb — so either the absorption instruction is losing against the v5 rules (five new constraints competing for the same edit), or the v5 rules legitimately forbid some absorptions and the checker counts them as divergence.

Not explained by repair: v4 ran 4–6 repair rounds per run, v5 runs 2.

## What this needs before the next canary

Diagnosis of the reconcile phases' behaviour, not a bar adjustment — moving the baseline after seeing results is the thing pre-registration forbids. Concretely: for each of the 12, determine whether full is missing a grounded fact (a real defect the arm exists to reduce) or the records legitimately differ and the checker over-counts. The `full=2 vs core=3` cases are checkable against the bundle directly.

Also verified end-to-end on this run, for the record: `report_regenerated_after_repair` fired (#604/#639 path), validation passed, lock acquired/released, all artifacts on disk.

realmarcin:
## Diagnosis complete: every examined divergence is benign, and my "systematic" claim needs a statistical correction

[...]

Issue #649, opened 2026-08-20
Commit the 2026-08-19 canary artifacts: the plan now cites data not in the repo


realmarcin:
Round-3 review of #645 (F7). `notes/generic_v5_analysis_plan.md`'s canary-history section cites the 2026-08-19 run — its regression numbers, its reconciliation report's own words — and the artifacts under `data/d4d_concatenated/claudecode_agent{,_core}/2026-08-19_claude-opus-5-api-generic-v5_rep1/` are **untracked**. The 2026-08-16 canary is tracked; the one the plan quotes is not.

Deliberately not swept into #645, which is a code/prompt branch — #594 records what happens when canary artifacts ride a code commit. But "deliberately excluded from this commit" and "committed" are different states, and only one of them lets a reader of the plan check its claims.

A separate data commit should add the run directory (full, core, report, provenance, reasoning log, intermediates). Worth doing before or with the next canary so the evidence trail for the #644 diagnosis is complete in-repo.

realmarcin:
The 2026-08-19 and 2026-08-20 evidence is committed in c7363c30 (PR #651). Stays half-open in spirit: the 2026-08-20b directory lands in the post-arm data commit, as the plan now states explicitly.

PR #651, opened 2026-08-20
Commit the canary evidence the plan cites, and record the fan-out decision (#649, #650)

realmarcin:
Data + docs PR; no code.

**The artifacts (#649):** the 2026-08-19 and 2026-08-20 canary runs — 26 files: full, core, reconciliation report, provenance, reasoning log, intermediates. The plan quotes the 2026-08-19 reconciliation report verbatim as the #644 diagnosis ("the resolver URL for the `uriorcurie`-ranged `id`"); until now those bytes were untracked. A data commit of its own per #594 — **the 2026-08-20b label is deliberately excluded** because the production arm is writing into it now; it lands post-arm.

**The plan (#650):** canary history extended through the third canary and the fan-out decision:

- v5 pair errors {13, 12, 12} vs v4 {10, 8, 6} — systematic (p ≈ 1/20), content-verified benign on all three runs, same slots recurring
- everything else passing, several far better than baseline (resolver URLs 0, undeclared prefixes 0, British 28 vs 146)
- decision: fan out under `--no-canary-gate`, baseline **unadjusted**, gate's verdict standing on the record
- consequence stated for any future analysis: v5-vs-v4 pair-error counts are not comparable at face value

The arm is running under `2026-08-20b_claude-opus-5-api-generic-v5` as this PR is opened.


realmarcin:
## Review: one selective statistic of mine, corrected; one recorder defect filed

**D1 — I dropped the fourth canary's pair count.** The 2026-08-16 canary's pair errors are **5**, below v4's *minimum*, measured live on the same finding-code family. I had disqualified that run's grounding verdict (its block predates the resolver check) and let the disqualification quietly swallow its pair count, which was never disqualified. Corrected: v5 = {5, 13, 12, 12}; "all v5 above v4's max" is false; p ≈ 4/35 ≈ 0.11, with exchangeability strained anyway (three prompt texts, different v4 schema digest). The honest statement — *suggestive, not established* — weakens the evidence of regression and so **strengthens** the override it was cited against. The selective version overstated the case for the conclusion I had already reached, which is the direction bias always points.

**D2** — the plan's 20b entry now states its evidence commit is pending (the arm is writing under that prefix), so the plan no longer silently cites untracked bytes. The post-arm data commit must include it. **D3** — header count fixed (four canaries).

**D4 filed as #652** — `companions.reasoning_log.md5` hashes a prefix: the recorder computes it before repair appends its lines (8/10 lines for 08-19, 6/8 for 08-20). Same class `_artifact()` dropped hashes to avoid, reproduced in the block added later. Authoritative pins (`validation.artifacts`, intermediates) all match.

Verified clean by the review: completeness of the 26-file evidence set, no credentials or absolute paths, the #644 quote at reconciliation.md:118, all hash pins matching, every plan number, and zero interference with the running sweep.

realmarcin:
Round 2 reviewed the round-1 correction itself and found one defect the correction introduced: "three post-#644 runs" — 08-19 ran the **pre**-fix prompt (its record hashes `97ccf288`); it is the run that produced the #644 diagnosis, as the same section says. Fixed, with the claim and the p-computation now scoped to the same grouping (the post-fix prompt alone has contributed {12, 12}).

Round 2 verified everything else: the 4/35 arithmetic and its conditioning (defensible; alternatives bracket it at 1/35–0.2 and "suggestive, not established" survives all three), the checker-version fact that makes 08-16's pair count comparable (the pair checker last changed *before* that run executed), the three-distinct-prompt count, both #652 examples independently, and the single-file footprint — nothing the running sweep reads.

